PR #1, opened 2019-03-20
Your changes + my changes for cloud

iphands:


src/entry-dev.js
```diff
@@ -6,17 +6,16 @@ import { init } from './store';
...
+}
 
-/**
- * Hooks up redux to app.
- *  https://redux.js.org/advanced/usage-with-react-router
- */
 ReactDOM.render(
     <Provider store={ init(logger).getStore() }>
-        <Router basename={ `/${RELEASE}/platform/advisor` }>
+        <Router basename={ `${pathName[0]}/${pathName[1]}` }>
```
karelhala: This is not correct because basename should start with leading `/`. The best way would be to update shif of beta
```JSX
const pathName = window.location.pathname.split('/');
pathName.shift();
let release = '/';
if (pathName[0] === 'beta') {
    release = `/${pathName.shift()}/`;
}

ReactDOM.render(
  <Provider store={ init(logger).getStore() }>
    <Router basename={ `${release}${pathName[0]}/${pathName[1]}` }>
      <App/>
    </Router>
  </Provider>

File: src/entry.js [hunk 2]
```diff
@@ -5,19 +5,19 @@ import { Provider } from 'react-redux';
...
+}
 
-/**
- * Hooks up redux to app.
- *  https://redux.js.org/advanced/usage-with-react-router
- */
 ReactDOM.render(
     <Provider store={ init().getStore() }>
-        <Router basename={ `/${RELEASE}/platform/advisor` }>
+        <Router basename={ `${pathName[0]}/${pathName[1]}` }>
```
karelhala: Same as above.

PR #17, opened 2019-09-10
Fix appname here too.

wenottingham:


ryelo:
@wenottingham should be https://github.com/RedHatInsights/tower-analytics-frontend/blob/6ef8040be29e61f5c452ac342585fc965891ec2c/package.json#L92

PR #26, opened 2019-09-20
Update docs url.

kialam:
Updating docs URL per https://github.com/ansible/product-docs/pull/856#issuecomment-532976864,


elyezer:
The following is shown where there is no data imported:

Screenshot from 2019-09-20 15-56-43


The link is now directing to the right URL, the one this PR is updating.

PR #28, opened 2019-09-20
Revert CSS changes until afer fest.

kialam:
Turns out `100vh` is not the height we want. It can cause background bugs like this if your browser height isn't maximized:
Screen Shot 2019-09-20 at 12 15 35 PM
This PR reverts the CSS changes that caused this new visual regression.

### DIFF
--- a/src/SmartComponents/SamplePage/SamplePage.js
+++ b/src/SmartComponents/SamplePage/SamplePage.js
@@ -288,9 +288,8 @@ class SamplePage extends Component {
                   id="refTab1Section"
                   ref={ this.contentRef1 }
                   aria-label="Tab item 1"
-                  style={ { height: '100vh' } }
               >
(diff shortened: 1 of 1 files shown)


elyezer:
I can't see the black background anymore, checked on firefox:

Screenshot from 2019-09-20 13-52-07

And Chrome:

Screenshot from 2019-09-20 13-52-27

elyezer:
Actually I can still see if I zoom out:

Screenshot from 2019-09-20 13-59-42

Will check again after https://github.com/RedHatInsights/tower-analytics-frontend/pull/29 hits production. It is supposed to completely fix the issue.

elyezer:
With #29 in place, now we don't get the black bar anymore:

Chrome:

Screenshot from 2019-09-20 14-42-12

Firefox:

Screenshot from 2019-09-20 14-43-09

Both with 50% zoom

PR #107, opened 2020-04-08
Roi calculator

kialam:
Related # https://github.com/RedHatInsights/tower-analytics-backend/issues/58
This PR adds the Automation Calculator to the UI.

scolling_chart


benthomasson:
In the description: 
Screen Shot 2020-04-07 at 10 22 07 AM

AparnaKarve:
@kialam This looks really great!

I ran it locally and only have a few css-y kind of suggestions that I have listed below, nothing too compelling.

- Would it be possible to make the left side label bold?
Since for multiple orgs or multiple clusters the text continues to the next line, and it would be nice to get a clear demarcation of the left from the right

Screen Shot 2020-04-09 at 12 03 33 PM

- Can we extend the length of the Top Templates card so that it aligns with the lower edge of the Automation Formula card on the left
Screen Shot 2020-04-09 at 12 07 35 PM

- In the Automation Formula card, is the short-form JT OK? or should we specify Job Template?

Issue #111, opened 2020-04-14
[RFE] Investigate options for Automation Calculator graph


appuk:
# Description
We had lots of suggestions on how we can improve the graph in Automation Calculator tab, this card tracks that

**Describe the solution you'd like**
A couple of ideas:
- [ ] Make the graph to display money savings instead
- [x] Just one bar to show the savings instead of 2 bars for manual and automation
- [ ] maybe specify title as “Automation time vs. Manual time”

Will update the ticket as more ideas pour in, Feel free to add to the above list

kialam:
I think @wenottingham mentioned during planning that this would remain a grouped bar chart with two bars, one bar representing/showing manual cost (e.g. $1200) and the other bar showing automated cost ($.07). Would that be preferable than the above RFE of showing just one bar that represents the delta of the two bars?

kialam:
For savings we need a reasonable default for cost of manual automation.

kialam:
We've decided to go with option 2. Latest changes are deployed to CI.

appuk:
looks great! 👍 
<img width="865" alt="Screen Shot 2020-04-21 at 4 47 53 PM" src="https://user-images.githubusercontent.com/11617351/79912339-e031bd80-83ef-11ea-93a3-2756442fb78f.png">
Verified the following scenarios to test the graph:
- graph when no templates ( no data page)
- less than 15 templates ( no scroll)
- more than 15 templates (scroll functionality enabled)
- more than 25 templates (only 25 templates display)
- tooltip contains manual cost, automation costs and savings
- the eye icon enables/disables the template from graph
- auto rendering when cost/ time changes

PR #114, opened 2020-04-16
Convert to single bar chart with updated tooltips showing savings.

kialam:
Related #111 
This PR changes the graph representing template elapsed run time (manual vs automated) to a single bar chart representing savings per template.

single_bar_savings


bronaghs:
those changes look great on the UI! So clear and concise.

PR #142, opened 2020-05-14
[WIP] Create fetch hook

### LINKED ISSUE #105: Create Reusable Fetch Hook
We would like to create a reusable fetch hook to use across components to reduce the amount of repeat code we have in our code base. Currently, when a container component needs to make a request to the API we use a custom fetchEndpoints method. 

The new reusable fetch hook should:
- Accept any number of APIs as an arguments
- Return the appropriate number of responses 
- Call getUser() method to authorize user token before request/response made
- STRETCH GOAL: hook will return certain booleans that the FE uses to dynamically render UI, e.i. isLoading, emptyState, etc.

brumik:
Closes #105 

Introduces a new hook `useFetch` which accepts two optional params: array to pass to Promise.all and the second optional param is the default data value.

Returns the data, if it is loading, if it had an error and preflight error. 

This PR also will update the previous components to use the fetch hook namely: Notifications, Clusters, ?OrganizationStatiustics?.

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -128,3 +128,17 @@ export const readROI = ({ params = {}}) => {
     );
     return fetch(url).then(handleResponse);
 };
+
+const API = {
+    getUser: window.insights.chrome.auth.getUser,
+    preflightRequest,
+    templateJobs: readTemplateJobs,
+    clusters: readClusters,
+    chart30: readChart30,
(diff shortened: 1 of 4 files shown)


brumik:
Moved the branch: https://github.com/RedHatInsights/tower-analytics-frontend/pull/150

Issue #81, opened 2020-02-20
Orgs should share colors in different charts if they are in multiple charts


kdelee:
noticing that the colors used across charts for same org is not consistent
For example, MasterReputation goes from a UNC Blue, to blue-green, to mustard yellow

![Screenshot from 2020-02-20 13-33-51](https://user-images.githubusercontent.com/29239279/74969014-db4d8f00-53e9-11ea-971f-10cff6e0e3ba.png)
![Screenshot from 2020-02-20 13-33-40](https://user-images.githubusercontent.com/29239279/74969024-dee11600-53e9-11ea-9d30-6d9a39f0501d.png)

some responses when discussing this:
```
1:41 PM
Corey Welton that's weird
1:41 PM
kialam:speech_balloon: @notting noticed this to. it has to do with the order in which the orgs get returned from the API
1:42 PM
so with 3 separate endpoints, each endpoint can return [orga, org b, orgc], [orgc, orga, orgb], etc.
1:43 PM
the color palette is an array of hex values and we just loop through the orgs and assign it by index value
1:43 PM
this might be worth making an orgs endpoint however, where the UI can query for all the orgs, assign it a color, and then make charts; i dunno
```

jctanner:
migrated to https://issues.redhat.com/projects/AA/issues/AA-146

PR #143, opened 2020-05-18
Converting charts to functional components

brumik:
Closes: #94 closes #102 closes  #103 and closes #104 

* Converts the BarChart, GruppedBarChart, LineChart, PieChart to a functional component.
* Coverts `BaseChart` to a functional component and fixes a `console` error where the `propTypes` were declared inside the component instead of after.

No UI changes as far as I tested it well.

PR #145, opened 2020-05-19
[Depends] Orgs in charts share color

brumik:
Closes: #81 

Builds on: https://github.com/RedHatInsights/tower-analytics-frontend/pull/143

Generating the colours to use in the start for each "name" which are ordered first to keep the same arrays same colours. It may not work if there are different arrays with the same names, but it should solve the problems in the Organisation Statistics page.

**Before**
Screenshot from 2020-05-19 17-25-47

**After**
Screenshot from 2020-05-19 17-24-40

### DIFF
--- a/src/Charts/BarChart.js
+++ b/src/Charts/BarChart.js
@@ -1,37 +1,18 @@
-import React, { Component } from 'react';
+import React, { useState, useEffect } from 'react';
 import initializeChart from './BaseChart';
 import * as d3 from 'd3';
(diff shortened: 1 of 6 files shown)


benthomasson:
Please make a branch on the upstream repo with the name `brumik-org-color-scheme-fix` instead of your own repo.  That will fix the Travis build failures

brumik:
Move to https://github.com/RedHatInsights/tower-analytics-frontend/
#149

PR #148, opened 2020-05-20
Converting charts to functional components

brumik:
Closes: #94 closes #102 closes #103 and closes #104

Converts the `BarChart`, `GruppedBarChart`, `LineChart`, `PieChart` to a functional component.
Coverts `BaseChart` to a functional component and fixes a console error where the `propTypes` were declared inside the component instead of after.

No UI changes as far as I tested it well.

PR #149, opened 2020-05-20
Orgs in charts share color

brumik:
Closes: #81 

Generating the colours to use in the start for each "name" which are ordered first to keep the same arrays same colours. It may not work if there are different arrays with the same names, but it should solve the problems in the Organisation Statistics page.

**Before**
Screenshot from 2020-05-19 17-25-47

**After**
Screenshot from 2020-05-19 17-24-40

### DIFF
--- a/src/Charts/GroupedBarChart.js
+++ b/src/Charts/GroupedBarChart.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import initializeChart from './BaseChart';
 import * as d3 from 'd3';
 import Legend from '../Utilities/Legend';
-import { pfmulti } from '../Utilities/colors';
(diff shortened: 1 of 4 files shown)


kialam:
Thanks for looking into this. It looks like the toggles are not working as expected with these new changes.

Expected:
toggle-expected

Actual:
toggle-actual

brumik:
This behaviour is caused by the erroneous refactor in #148. It should be fixed there, and after it gets merged I will rebase this to master (only the last commit).

brumik:
Now that the refactor was merged I think this is good to go. Retested after rebase and it is working afaik.

brumik:
@kialam

brumik:
Closing in favour of #188. Does the same thing, better.

PR #158, opened 2020-06-08
Adds basic unit tests

benthomasson:
These unit tests test whether a component can be loaded and shallowly rendered.   This might be useful in the upgrade to patternfly 2020.07 process.  It also raises the coverage above 0% for many files.


brumik:
You added unit tests to my `Utilities/colour` faster than I even finalized the PR :smile: #149

Also now we have duplicated tests for charts: 3a03051af288a6ac4ac363c4d388cc0dcdb4fbee

benthomasson:
@brumik feel free to replace any of my tests with better ones.

Issue #75, opened 2020-02-19
Annotate org automation with success rates


benthomasson:
When we show automation for an organization, we should also show the success/failure rate of that automation.

E.g. "Weyland Yutani Organization: 33% (75% successful)"

benthomasson:
This should appear on the Organizational Status and the Jobs by Organization tooltips.

jctanner:
@benthomasson  Are you talking about this card's piechart tips hover over thingy? ...

file:///home/jtanner/Desktop/Screenshot%20from%202020-06-03%2012-41-55.png

benthomasson:
@jctanner yes that is correct.

jctanner:
The elements of the tooltip are current aligned vertically to fit within the bounds of the graph, so a new element would likely have to go below the total jobs versus horizontally aligned ...

![success](https://user-images.githubusercontent.com/1869705/83677495-43e00680-a5aa-11ea-8733-fb30ccc4d6ea.png)

Does it make sense to also qualify the total percentage as well? If the legend depicts an integer for job count, and there's going to be a new element to depict the percentage of failures ... the "total" percent might be construed as the total number of failures rather than percentage of all jobs in the cluster.

jctanner:
resolved by https://github.com/RedHatInsights/tower-analytics-frontend/pull/157

jctanner:
@appuk sure, I'll try to look at it today.

jctanner:
reopening for testing

appuk:
I apologize for not doing this earlier 😄 
Verified the tooltip for success rates, looks good 👍 
![Screen Shot 2020-06-23 at 11 43 26 AM](https://user-images.githubusercontent.com/11617351/85425230-ee629e00-b546-11ea-890f-b3337c22fac3.png)
![Screen Shot 2020-06-23 at 11 43 32 AM](https://user-images.githubusercontent.com/11617351/85425237-f0c4f800-b546-11ea-820f-c782b9a2e5c2.png)

PR #163, opened 2020-06-11
refactor piechart tooltip to understand the data shape

jctanner:
addresses https://github.com/RedHatInsights/tower-analytics-frontend/issues/75


benthomasson:
Does this delay the drawing until it is moused over?

jctanner:
@benthomasson correct

PR #172, opened 2020-06-29
Job Explorer Feature

kialam:
Related #127

PR #182, opened 2020-07-08
ROI calculator date filter

### LINKED ISSUE #173: Add date filtering to the ROI calculator
# Description

Currently, the ROI calculator shows the calculation across the entirety of data we have stored for the account.
Customers want to be able to show the ROI for a particular date range (most often, year to date, or for a particular month)

**Describe the solution you'd like**
The ROI calculator should use our standard date picking widgetry, and use that to query the API for the date range specified.
The API should already support these queries.

The date range should potentially have a default. One suggestion would be "year to date".

**Describe alternatives you've considered**
Alternatives would be hardcoded UI choices (year to date, last 30 days, last week). But if we have a standard date picker interaction model, and the API supports it, I don't see a reason not to just do arbitrary dates.

**Additional context**

**Mock up**

**Related PRs**

# Verification

**Screenshot**

Add screenshot of implementation here when implemented

**Steps**

Add verification steps here when ready for QE

jctanner:
Fixes #173


jctanner:
image

@wenottingham @trahman73 does this look suitable ^ ?

wenottingham:
If we don't have full date filters in other places, this would work. At some point we should be able to do arbitrary date filtering in general.

jctanner:
@wenottingham yep, all of our date selectors are static buckets.

kialam:
@jctanner please sync up with @AparnaKarve who wrote the API backend for ROI. AFAIK it does not currently support custom date ranges but I could be mistaken.

Ladas:
@kialam @jctanner custom date ranges are supported, so it should work the same as other endpoints in v0_1 API

jctanner:
The component does a GET on a url like `https://prod.foo.redhat.com:1337/api/tower-analytics/roi_templates/?startDate=2020-03-08&endDate=2020-07-10`. The parameters are sent to the handler by https://github.com/RedHatInsights/tower-analytics-backend/blob/master/restapi/routers/v0_1.py#L175-L176 when then appear to be used in the sql query here https://github.com/RedHatInsights/tower-analytics-backend/blob/master/restapi/handlers.py#L1141

kialam:
@Ladas @jctanner we don't have a UI/UX workflow for any of our components for custom date ranges for `v0_1` api endpoints. So for now, let's stick with the pre-determined date ranges we currently have.

kialam:
Once we convert our existing `v0_1` endpoints to the new `v1` endpoints, we can reuse the custom date range picker from the Job Explorer list.

jctanner:
@kialam works for me!

AparnaKarve:
@jctanner How about we also add a Year To Date (YTD) filter?

jctanner:
@AparnaKarve that is the default ("Past Year") setting.

AparnaKarve:
oh I see - for some reason I interpreted "Past Year"  as the actual past year, as in, 2019.

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -128,6 +150,21 @@ const title = (
...
+    startDate: moment
+    .utc()
+    .subtract(1, 'year')
+    .format('YYYY-MM-DD'),
+    endDate: moment.utc().format('YYYY-MM-DD')
+};
+
+const timeFrameOptions = [
+    { value: 'please choose', label: 'Select date range', disabled: true },
+    { value: 365, label: 'Past year', disabled: false },
```
AparnaKarve, line 163: 'Past year' seems to be for the past 365 days. So for Today, this would mean 5 months of 2019 and 7 months of 2020.

Year To Date, OTOH, is a period, starting from the beginning of the current year and continuing up to the present day ([source](https://en.wikipedia.org/wiki/Year-to-date))

I think it would be a good addition. But you make the call :)

jctanner: Hrm, yeah I see your point. Let's check with @wenottingham

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -128,6 +150,21 @@ const title = (
...
+    { value: 31 * 3, label: 'Past quarter', disabled: false },
```
kialam, line 164: I think we can leverage moment here to calculate past quarter instead of doing a manual 31 * 4 days: `moment().startOf('quarter');` or you could do the last 3 months:
```
moment
    .utc()
    .subtract(3, 'months')
    .format('YYYY-MM-DD')
```

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -236,12 +273,15 @@ export const useAutomationFormula = () => {
     const [ roiData, setRoiData ] = useState([]);
     const [ selectedIds, setSelectedIds ] = useState([]);
 
+    /* this drives the start date parameter sent to the API */
+    let [ selectedStartValue, setSelectedStartValue ] = useState(365);
```
kialam, line 277: Line `153-158` + Line `386-387` are the ones that set the initial start and end date parameters when the page first loads.

jctanner:
@kialam i've implemented what I think you've suggested, however I have a few concerns:

1) useQueryParam's setEndDate function ignores inputs and always forces the value to $TODAY.
2) useQueryParam's setStartDate only works with integer values that are implied to be X days prior to $TODAY
3) readROI takes in timestamps not integers.

That all requires a confusing mix of using ints and timestamps in AutomationCalculator. If I have to create all parameters for readROI with useQueryParams, I'm always limited to the last N days. We will have to change it completely if we ever move to real date pickers.

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -128,6 +151,30 @@ const title = (
...
+const pastYTD = moment().startOf('year');
+const pastQuarter = moment().startOf('quarter');
+/* const pastQuarter = moment.utc().subtract(3, 'months'); */
+const pastMonth = moment().startOf('month');
+const today = moment.utc();
+
+/* we want a simple integer value to use in the form as the selected value */
+/* useQueryParam's setStartDate only handles integers for setStartDate */
+/* useQueryParam's setEndDate always forces the date to today and ignores any input */
+const timeFrameOptions = [
```
kialam, line 164: This seems overly complex for what it needs to be. Suggest something like:
```javascript
const timeFrameOptions = [
    { value: 'please choose', label: 'Select date range', disabled: true },
    { value: 365, label: 'Past year', disabled: false },
    { value: 1, label: 'Past year to date', disabled: false },
    { value: 90, label: 'Past quarter', disabled: false },
    { value: 31, label: 'Past quarter', disabled: false },
    { value: 31, label: 'Past month', disabled: false }
];
```

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -238,10 +285,19 @@ export const useAutomationFormula = () => {
 
     const { formatData } = automationCalculatorMethods();
 
+    const { queryParams, setEndDate, setStartDate } = useQueryParams(
```
kialam, line 288: I would move this into `Line 395`

kialam: Likewise I would also move the `useEffect()` hook into `AutomationCalculator`.

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -329,11 +385,23 @@ export const useAutomationFormula = () => {
...
 
 const AutomationCalculator = () => {
+
+    const {
+        updateData,
+        handleManualTimeChange,
+        handleToggle
+    } = automationCalculatorMethods();
+
+    const [ roiTimeFrame, setRoiTimeFrame ] = useState(1);
```
kialam, line 403: If we want the default date selection to be "Past year" and our `timeFrameOptions` object is the following as I suggested, then set this value to `365`.

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -346,14 +414,27 @@ const AutomationCalculator = () => {
         templatesList,
         selectedIds,
         setSelectedIds,
-        preflightError
+        preflightError,
```
kialam, line 417: There is no need to import these values if they get moved into the main block itself.

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -346,14 +414,27 @@ const AutomationCalculator = () => {
...
+        setStartDate,
+        setEndDate
     } = useAutomationFormula();
 
-    const {
-        updateData,
-        handleManualTimeChange,
-        handleToggle
-    } = automationCalculatorMethods();
+    const onChange = (index) => {
```
kialam, line 423: The PF `FormSelect` component returns the form `value` as a string as its first argument, and the synthetic event as its second argument (should you need it). You can read the documentation here: http://patternfly-react.surge.sh/documentation/react/components/formselect#props

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -371,6 +452,34 @@ const AutomationCalculator = () => {
...
+                              value={ roiTimeFrame }
+                              onChange={ onChange }
+                              aria-label="Select Date Range"
+                              style={ { margin: '2px 10px' } }
+                          >
+                              { timeFrameOptions.map((option, index) => (
+                                  <FormSelectOption
+                                      isDisabled={ option.disabled }
+                                      key={ index }
+                                      value={ index }
```
kialam, line 474: Should be `option.value`.

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -260,7 +316,7 @@ export const useAutomationFormula = () => {
 
         initializeWithPreflight();
         return () => (ignore = true);
-    }, []);
+    }, [queryParams]);
```
kialam, line 319: 👍

kialam:
> @kialam i've implemented what I think you've suggested, however I have a few concerns:
> 
> 1. useQueryParam's setEndDate function ignores inputs and always forces the value to $TODAY.
> 2. useQueryParam's setStartDate only works with integer values that are implied to be X days prior to $TODAY
> 3. readROI takes in timestamps not integers.
> 
> That all requires a confusing mix of using ints and timestamps in AutomationCalculator. If I have to create all parameters for readROI with useQueryParams, I'm always limited to the last N days. We will have to change it completely if we ever move to real date pickers.

@jctanner thanks for raising these concerns.
RE first concern: Yes, when we implement custom date pickers we will have to set actual end dates. Admittedly, `setEndDate` is a somewhat mis-name of the actual method, as it sets the current days date and should be called `setCurrentDate` or something. #172  correctly refactors `setEndDate` to accept user input. See [here](https://github.com/RedHatInsights/tower-analytics-frontend/blob/jo_job-explorer-list-view/src/Utilities/useQueryParams.js#L172).

RE second concern: Agreed, likewise #172 correctly refactors `setStartDate` to accept user input. See [here](https://github.com/RedHatInsights/tower-analytics-frontend/blob/jo_job-explorer-list-view/src/Utilities/useQueryParams.js#L171).

I plan to deprecate the old `setStartDate` and `setEndDate` methods once we completely switch over to the new v1 API.

As a side note, since we are adding new date ranges in your PR (Past year, year to date, and past quarter), we should update the [setStartDate](https://github.com/RedHatInsights/tower-analytics-frontend/blob/master/src/Utilities/useQueryParams.js#L60-L89) to accept these new ranges. E.g.:
```javascript
...
if (days === 365) {
                startDate = moment
                .utc()
                .subtract(1, 'year')
                .format('YYYY-MM-DD');
if (days === 90) {
                startDate = moment
                .utc()
                .startOf('quarter');
                .format('YYYY-MM-DD');
```

RE 3: Yes, the API expects a date formatted string, `YYYY-MM-DD`.

💡  Alternatively, if we think passing integers to the `useQueryParams` hook is too confusing, we can leverage `moment` to set the values in our `timeFrameOptions` object to the date formatted string the API expects and borrowing bits of what you've already written:

```javascript
const timeFrameOptions = [
    { value: 'please choose', label: 'Select date range', disabled: true },
    { value: pastYear.format('YYYY-MM-DD'), label: 'Past year', disabled: false },
    { value: pastYTD.format('YYYY-MM-DD'), label: 'Past year to date', disabled: false}
    ...
];

const handleOnChange = value => {
        setStartDate(value);
        setRoiTimeFrame(value);
};
```

`setStartDate` could then be refactored as simply:

`setStartDate: formattedDate => dispatch({ type: 'SET_STARTDATE', formattedDate })`

If we go this route, the rest of the app will have to be refactored to reflect these changes (which should be fairly easy).

src/Containers/AutomationCalculator/AutomationCalculator.js
kialam: If we go with the moment formatted date string value, then this would be `useState(pastYear.format('YYYY-MM-DD'))`

jctanner:
Notes from the call with @kialam ...

```
diff --git a/src/Containers/AutomationCalculator/AutomationCalculator.js b/src/Containers/AutomationCalculator/AutomationCalculator.js
index f64ae89..a5af171 100644
--- a/src/Containers/AutomationCalculator/AutomationCalculator.js
+++ b/src/Containers/AutomationCalculator/AutomationCalculator.js
@@ -155,18 +155,26 @@ const pastYear = moment.utc().subtract(1, 'year');
 const pastYTD = moment().startOf('year');
 const pastQuarter = moment().startOf('quarter');
 /* const pastQuarter = moment.utc().subtract(3, 'months'); */
-const pastMonth = moment().startOf('month');
+// const pastMonth = moment().startOf('month');
+const pastMonth = moment().utc().subtract(1, 'month');
 const today = moment.utc();
 /* we want a simple integer value to use in the form as the selected value */
 /* useQueryParam's setStartDate only handles integers for setStartDate */
 /* useQueryParam's setEndDate always forces the date to today and ignores any input */
+// const timeFrameOptions = [
+//     { id: 0, value: 'please choose', label: 'Select date range', disabled: true },
+//     { id: 1, value: 365, range: [today.diff(pastYear, 'days'), today.format('YYYY-MM-DD')], label: 'Past year', disabled: false },
+//     { id: 2, value: 2, range: [today.diff(pastYTD, 'days'), today.format('YYYY-MM-DD')], label: 'Past year to date', disabled: false },
+//     { id: 3, value: 3, range: [today.diff(pastQuarter, 'days'), today.format('YYYY-MM-DD')], label: 'Past quarter', disabled: false },
+//     { id: 4, value: 4, range: [today.diff(pastMonth, 'days'), today.format('YYYY-MM-DD')], label: 'Past month', disabled: false }
+// ];
 const timeFrameOptions = [
-    { id: 0, value: 'please choose', label: 'Select date range', disabled: true },
-    { id: 1, value: 1, range: [today.diff(pastYear, 'days'), today.format('YYYY-MM-DD')], label: 'Past year', disabled: false },
-    { id: 2, value: 2, range: [today.diff(pastYTD, 'days'), today.format('YYYY-MM-DD')], label: 'Past year to date', disabled: false },
-    { id: 3, value: 3, range: [today.diff(pastQuarter, 'days'), today.format('YYYY-MM-DD')], label: 'Past quarter', disabled: false },
-    { id: 4, value: 4, range: [today.diff(pastMonth, 'days'), today.format('YYYY-MM-DD')], label: 'Past month', disabled: false }
+    { value: 'please choose', label: 'Select date range', disabled: true },
+    { value: pastYear.format('YYYY-MM-DD'), label: 'Past year', disabled: false },
+    { value: pastYTD.format('YYYY-MM-DD'), label: 'Past year to date', disabled: false },
+    { value: pastQuarter.format('YYYY-MM-DD'), label: 'Past quarter', disabled: false },
+    { value: pastMonth.format('YYYY-MM-DD'), label: 'Past month', disabled: false }
 ];
 /* readROI must take datestrings for startDate and endDate */
@@ -285,7 +293,7 @@ export const useAutomationFormula = () => {
     const { formatData } = automationCalculatorMethods();
-    const { queryParams, setEndDate, setStartDate } = useQueryParams(
+    const { queryParams, setEndDate, setStartDateAsString } = useQueryParams(
         initialQueryParams
     );
@@ -420,20 +428,22 @@ const AutomationCalculator = () => {
         setEndDate
     } = useAutomationFormula();
-    const onChange = (index) => {
-        console.log('AC.onchange ...');
-        console.log(index);
-        console.log(timeFrameOptions[parseInt(index)]);
-        console.log('AC.onchange setstartdate to ...');
-        setStartDate(timeFrameOptions[parseInt(index)]['range'][0]);
-        console.log(timeFrameOptions[parseInt(index)]['range'][0]);
-        console.log(queryParams);
-        console.log('AC.onchange setenddate to ...');
-        setEndDate(timeFrameOptions[parseInt(index)]['range'][1]);
-        console.log(timeFrameOptions[parseInt(index)]['range'][1]);
-        console.log(queryParams);
-
-        setRoiTimeFrame(index);
+    const onChange = (val) => {
+        console.log('val', typeof val);
+        setStartDateAsString(val);
+        // console.log('AC.onchange ...');
+        // console.log(index);
+        // console.log(timeFrameOptions[parseInt(index)]);
+        // console.log('AC.onchange setstartdate to ...');
+        // setStartDate(timeFrameOptions[parseInt(index)]['range'][0]);
+        // console.log(timeFrameOptions[parseInt(index)]['range'][0]);
+        // console.log(queryParams);
+        // console.log('AC.onchange setenddate to ...');
+        // setEndDate(timeFrameOptions[parseInt(index)]['range'][1]);
+        // console.log(timeFrameOptions[parseInt(index)]['range'][1]);
+        // console.log(queryParams);
+
+        setRoiTimeFrame(val);
     };
     return (
@@ -471,7 +481,7 @@ const AutomationCalculator = () => {
                                   <FormSelectOption
                                       isDisabled={ option.disabled }
                                       key={ index }
-                                      value={ index }
+                                      value={ option.value }
                                       label={ option.label }
                                   />
                               )) }
```

```
diff --git a/src/Utilities/useQueryParams.js b/src/Utilities/useQueryParams.js
index 3a60489..5174df2 100644
--- a/src/Utilities/useQueryParams.js
+++ b/src/Utilities/useQueryParams.js
@@ -42,6 +42,9 @@ export const useQueryParams = initial => {
                 }
                 return { ...state, severity: action.severity };
+
+            case 'SET_STARTDATEASSTRING':
+                return { ...state, startDate: action.dateString };
             default:
                 throw new Error();
         }
@@ -79,10 +82,7 @@ export const useQueryParams = initial => {
                 .subtract(1, 'month')
                 .format('YYYY-MM-DD');
             } else {
-                startDate = moment
-                .utc()
-                .subtract(days, 'days')
-                .format('YYYY-MM-DD');
+                startDate = days;
             }
             dispatch({ type: 'SET_STARTDATE', startDate });
@@ -90,6 +90,7 @@ export const useQueryParams = initial => {
         setSortBy: sort => dispatch({ type: 'SET_SORT_BY', sort }),
         setLimit: limit => dispatch({ type: 'SET_LIMIT', limit }),
         setOffset: offset => dispatch({ type: 'SET_OFFSET', offset }),
-        setSeverity: severity => dispatch({ type: 'SET_SEVERITY', severity })
+        setSeverity: severity => dispatch({ type: 'SET_SEVERITY', severity }),
+        setStartDateAsString: startDate => dispatch({ type: 'SET_STARTDATE', startDate })
     };
 };
```

jctanner:
TODO:
jest tests to assert hooks are called when selection changes

package.json
```diff
@@ -71,6 +71,7 @@
     "npm-run-all": "^4.1.3",
     "postcss": "^7.0.16",
     "prop-types": "^15.6.2",
+    "redux-mock-store": "^1.5.4",
```
kialam, line 74: not sure why this library is needed?

jctanner:
@kialam current tests complete successfully, but @keithjgrant has figured out that we need to update react and react-dom to get act+async working together ...


```
[jtanner@jtx1 tower-analytics-frontend]$ git diff package.json
diff --git a/package.json b/package.json
index e750546..afc90f7 100644
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
     "babel-plugin-transform-imports": "^2.0.0",
     "classnames": "^2.2.5",
     "moment": "^2.24.0",
-    "react": "16.8.6",
-    "react-dom": "16.8.6",
+    "react": "^16.9.0",
+    "react-dom": "^16.9.0",
     "react-redux": "^5.0.7",
     "react-router-dom": "^4.2.2",
     "redux": "^3.7.2",
```

I'm not sure how this would affect the rest of the app. Any thoughts on what to do?

package.json
keithjgrant: This mocks the redux store to give to the Provider so you don't have to spin up the full redux environment in tests

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -1,7 +1,11 @@
+/*eslint-disable */
```
brumik, line 1: I think you forgot to delete it :)

src/Containers/AutomationCalculator/AutomationCalculator.test.js
```diff
@@ -1,7 +1,30 @@
 /*eslint camelcase: ["error", {properties: "never", ignoreDestructuring: true}]*/
+/*eslint-disable padding-line-between-statements */
```
kialam, line 2: remove

src/Containers/AutomationCalculator/AutomationCalculator.test.js
```diff
@@ -1,7 +1,30 @@
...
+
 import { act } from 'react-dom/test-utils';
 import { automationCalculatorMethods, useAutomationFormula } from './AutomationCalculator';
 
+//import Api, { preflightRequest, readROI } from '../../Api';
```
kialam, line 7: remove

src/Containers/AutomationCalculator/AutomationCalculator.test.js
```diff
@@ -1,7 +1,30 @@
...
+//import * as math from "./math";
```
kialam, line 8: same as above

src/Containers/AutomationCalculator/AutomationCalculator.test.js
```diff
@@ -1,7 +1,30 @@
...
+
+ApiFuncs.preflightRequest = jest.fn();
+ApiFuncs.readROI = jest.fn();
+const mockResponse = {
+    ok: true,
+    json: () => Promise.resolve({})
+};
+ApiFuncs.preflightRequest = jest.fn();
+ApiFuncs.preflightRequest.mockResolvedValue(mockResponse);
+ApiFuncs.readROI = jest.fn();
```
kialam, line 25: Is this the same as line `18`?

src/Containers/AutomationCalculator/AutomationCalculator.test.js
```diff
@@ -291,3 +314,70 @@ describe('automationCalculatorMethods()', () => {
...
+
+        //await window.insights.chrome.auth.getUser();
+        global.insights = mockInsights;
+
+        // full mount ...
+        wrapper = mount(<Provider store={ store } ><AutomationCalculator /></Provider>);
+
+    });
+
+    afterEach(() => {
```
kialam, line 348: It does not appear that cleanup is occurring after each test assertion so this can be removed.

src/Containers/AutomationCalculator/AutomationCalculator.test.js
```diff
@@ -291,3 +314,70 @@ describe('automationCalculatorMethods()', () => {
         expect(automationCalculatorMethods().handleToggle('1', [ 1, '1' ])).toEqual([ 1 ]);
     });
 });
+
+describe('AutomationCalculator()', () => {
+
+    let wrapper;
+    //let mockHandleOnChange = jest.fn();
```
kialam, line 321: remove

src/Containers/AutomationCalculator/AutomationCalculator.test.js
```diff
@@ -291,3 +314,70 @@ describe('automationCalculatorMethods()', () => {
...
+    //const spy = jest.spyOn(Api, 'readROI');
```
kialam, line 322: same as above

src/Containers/AutomationCalculator/AutomationCalculator.test.js
```diff
@@ -291,3 +314,70 @@ describe('automationCalculatorMethods()', () => {
...
+                getUser: () => {
+                    return 'bob';
+                }
+            }
+        }
+    };
+
+    beforeEach(() => {
+
+        //newSpy = jest.spyOn(AutomationCalculator.prototype, 'handleOnChange');
```
kialam, line 336: remove

src/Containers/AutomationCalculator/AutomationCalculator.test.js
```diff
@@ -291,3 +314,70 @@ describe('automationCalculatorMethods()', () => {
...
+        const mockStore = configureStore();
+        const store = mockStore({});
+
+        //await window.insights.chrome.auth.getUser();
```
kialam, line 340: same as above?

src/Containers/AutomationCalculator/AutomationCalculator.test.js
```diff
@@ -291,3 +314,70 @@ describe('automationCalculatorMethods()', () => {
...
+        global.insights = mockInsights;
+
+        // full mount ...
+        wrapper = mount(<Provider store={ store } ><AutomationCalculator /></Provider>);
```
kialam, line 344: 👍 that's a tricky one

src/Utilities/useQueryParams.js
```diff
@@ -87,6 +87,7 @@ export const useQueryParams = initial => {
 
             dispatch({ type: 'SET_STARTDATE', startDate });
         },
+        setStartDateAsString: startDate => dispatch({ type: 'SET_STARTDATE', startDate }),
```
kialam, line 90: 🥇

PR #185, opened 2020-07-09
Jo job explorer ui cleanup

jeanetteobr:
This PR handles UI cleanup for #172


TODO
```diff
@@ -0,0 +1,9 @@
+
```
kialam, line 1: Let's leave out this file

src/Components/JobExplorerList.js
```diff
@@ -1,14 +1,15 @@
-import React, { useState } from 'react';
+/* eslint-disable no-console */
```
kialam, line 1: remove debugging

kialam:
After speaking with @trahman73 the toggle should be grouped with the dropdowns. Also noticed a double left indent for the filterable toolbar.

benthomasson:
@kialam @brumik is this PR still needed?

kialam:
I think we can close this for now and I will work with @trahman73 on any outstanding UI issues for job explorer v1 since we've done a fair bit of updating to the UI since this PR was first opened and some changes no longer apply.

PR #188, opened 2020-07-10
Feature: Hosts tab for org stats with groupped bar chart

brumik:
## Host chart tab

Issue: https://github.com/RedHatInsights/tower-analytics-backend/issues/59

It adds 2 tabs: Jobs and Hosts. The `Jobs` tab contains the original chart, the `Hosts` tab contains a similar chart with the host data - according to the mockup: https://tower-mockups.testing.ansible.com/examples/analytics/orgs/analytics-hosts/

The chart contained hardcoded strings and the data has instead of `value` -> `count` and `org_name` -> `name` therefore is the duplication (the hostsBarChart is near identical to the grouppedBarChart). When using tabs the charts are forced to render when changing tabs. Without it, their size will be always the default. 

Additionally simplified the code for fetching data with the help of better hooks.

**The tabs:**
Screenshot from 2020-07-23 14-54-13
Screenshot from 2020-07-23 14-54-24

## Orgs colour fix: same orgs same colour across all charts
Issue: #81

**Before:**
Screenshot from 2020-07-10 19-09-23
Screenshot from 2020-07-10 19-09-38

**After:** See the screenshots under the first section.

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -3,6 +3,7 @@
 const barChartEndpoint = '/api/tower-analytics/chart30/';
 const clustersEndpoint = '/api/tower-analytics/clusters/';
 const groupedBarChartEndpoint = '/api/tower-analytics/jobs_by_date_and_org_30/';
+const hostsEndpoint = '/api/tower-analytics/hosts_by_date_and_org/';
 const modulesEndpoint = '/api/tower-analytics/modules/';
 const notificationsEndPoint = '/api/tower-analytics/notifications/';
 const pieChart1Endpoint = '/api/tower-analytics/job_runs_by_org_30/';
@@ -69,6 +70,15 @@ export const readChart30 = ({ params = {}}) => {
 export const readJobsByDateAndOrg = ({ params = {}}) => {
     const formattedUrl = getAbsoluteUrl();
     let url = new URL(groupedBarChartEndpoint, formattedUrl);
+    Object.keys(params).forEach(key =>
+        url.searchParams.append(key, params[key])
+    );
+    return fetch(url).then(handleResponse);
+};
+
+export const readHostsByDateAndOrg = ({ params = {}}) => {
+    const formattedUrl = getAbsoluteUrl();
+    let url = new URL(hostsEndpoint, formattedUrl);
     Object.keys(params).forEach(key =>
         url.searchParams.append(key, params[key])
     );
--- a/src/App.scss
+++ b/src/App.scss
@@ -59,7 +59,7 @@ h2 {
   color: #393f44;
 }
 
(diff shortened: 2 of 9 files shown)


kialam:
Ah yes, I see what you mean by the colors changing when you toggle the tabs. Is there any way to lift this [block of code you've already written](https://github.com/RedHatInsights/tower-analytics-frontend/pull/149/files#diff-ef6c089ce9dbe620f7e3ec601ae50c45R1-R35) and add it to this PR? Seems like it should be easily integrated, sans the conversion from class-based components to functional components.

@appuk I know during our `release 1.1.0` meeting we discussed that the hosts by org feature was not going to be in 1.1.0, however since @brumik is now back he has cycles to do feature work. Do we have QE resources to test this feature for `release 1.1.0`?

appuk:
> @appuk I know during our `release 1.1.0` meeting we discussed that the hosts by org feature was not going to be in 1.1.0, however since @brumik is now back he has cycles to do feature work. Do we have QE resources to test this feature for `release 1.1.0`?

@kialam  AFAIK, we are still not aware of the timeline for when 1.1 will be released, right? so if I can't really say anything unless I get an idea about that first.
I also have a backlog of Test Coverage I would like to cover soon

src/Api.js
```diff
@@ -3,6 +3,7 @@
 const barChartEndpoint = '/api/tower-analytics/chart30/';
 const clustersEndpoint = '/api/tower-analytics/clusters/';
 const groupedBarChartEndpoint = '/api/tower-analytics/jobs_by_date_and_org_30/';
+const hotsEndpoint = '/api/tower-analytics/hosts_by_date_and_org/';
```
Ladas, line 6: typo: hots -> hosts ?

Ladas:
@wenottingham @trahman73 we can probably play a bit more with the naming?

We are showing `Unique hosts usage per organization` and it might be goo to note somewhere that this serves for checking if customer utilizes the host limit?

Then this is `unique host count per day`, we can also show unique host count for the whole period of time (e.g. as pie chart or in the legend?) And we can show unique host count across all orgs, so something like `1110 unique hosts were configured across 25 organizations for past month`

And we could correlate the actual host runs to that? Maybe showing it in a tooltip? (it's not in the API now) So e.g.
org1 running on host: A, B, B, B,B

That is 2 unique hosts, but 5 host runs.

brumik:
@Ladas Any update on this?

brumik:
Talked with @trahman73 and the wording should be a separate issue and it is non-blocking. @kialam Can you look at this and merge if you see fit?

brumik:
Closing in favour of https://github.com/RedHatInsights/tower-analytics-frontend/pull/291

Issue #183, opened 2020-07-08
Link top templates to job explorer view


kialam:
# Description

Now that we have the new `v1/job_explorer` endpoint, we can now link the top templates to the job explorer list, filtered by `template_id`.

**Is your feature request related to a problem? Please describe.**
As an AA user, I'd like to click on a template name on the Top Templates section on the Clusters page and see a list of jobs filtered by the specific template.

**Describe the solution you'd like**
When a user clicks on a template in the Top Templates list (see Exhibit A) the user should should also see a "View all jobs" link that redirects them the job explorer page filtered by template (see Exhibit B).

[Exhibit A]
![view_all_jobs](https://user-images.githubusercontent.com/2293210/87099845-cd719c80-c218-11ea-9f51-230e8b9242ef.png)

[Exhibit B]
<img width="1792" alt="filter_by_template" src="https://user-images.githubusercontent.com/2293210/86970125-de4ce000-c13c-11ea-9bc7-f4853c7018c6.png">

Additionally, the top templates detail modal list can be refactored to use the new `v1/job_explorer` endpoint, ordered by most recent from the past month, returning the first 5 results, filtered by the specific template_id and job type of job.

The top templates detail modal aggregate stats box can be refactored to use the new `v1/job_explorer` endpoint by passing a `group_by: template` param which will then return (most) of the aggregate statistics.

This work should be branched off the job explorer feature branch - `jo_job-explorer-list-view`.

**Additional context**
@benthomasson @Ladas there appears to be a discrepancy in run count for the templates across views in our app. On the Clusters page the API is returning `total_run_count: 16` while on the Job Explorer page the total number of jobs is `48`. I _believe_ this is because the `/v0/template_jobs` is ignoring nested workflow jobs and jobs, and only returning `status: successful` jobs, thus `16` instead of `48`. Not sure how we want to handle this - should we pre-filter with those same parameters on the job explorer page or adjust the `v0/template_jobs` endpoint to return 48?

**Mock up**

Add mock up here when ready

**Related PRs**

Add PRs here when ready

# Verification

**Screenshot**

Add screenshot of implementation here when implemented

**Steps**

Add verification steps here when ready for QE

Ladas:
@kialam so the v0_1 list of jobs uses filter:

Filters only `job_type: job` then specific cluster if set and last 30 days by default. Then the total count should be the same. If it's not, there might be a bug.


--------

[...]

PR #202, opened 2020-07-23
Brumik link top templates to job expoler

brumik:
#183 

Adds a link to the top templates modal for the job-explorer page. Build on top of: `jo_job-explorer-list-view` branch.

Additionally fixes a console error: when `clusters` page was loaded it automatically loaded the modal window as well because of a wrong if statement and throw a `Modal needs a title or ...` error in the console. 

**Before**:
Screenshot from 2020-07-22 16-57-12

**After**:
Screenshot from 2020-07-24 15-14-39

**After clicking on `View all jobs`:**
Screenshot from 2020-07-23 13-52-36

### DIFF
--- a/src/Components/TemplatesList.js
+++ b/src/Components/TemplatesList.js
@@ -1,9 +1,12 @@
+/*eslint camelcase: ["error", {allow: ["template_id", "job_type", "cluster_id", "start_date", "end_date"]}]*/
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { formatDateTime, formatSeconds, formatPercentage } from '../Utilities/helpers';
 import styled from 'styled-components';
 import LoadingState from '../Components/LoadingState';
 import NoData from '../Components/NoData';
+import { Paths } from '../paths';
+import { formatQueryStrings } from '../Utilities/formatQueryStrings';
 
 import {
     Badge,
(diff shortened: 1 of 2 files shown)


src/Components/TemplatesList.js
```diff
@@ -88,7 +91,14 @@ const formatTopFailedTask = data => {
     return `Unavailable`;
 };
 
-const TemplatesList = ({ templates, isLoading, queryParams }) => {
+const borderedButtonStyle = {
```
kialam, line 94: Should use the PF `label` component for this: https://www.patternfly.org/v4/documentation/react/components/label#outline

kialam:
New elements should be vertically aligned as per the mockups
Screen Shot 2020-07-23 at 12 49 13 PM

src/Components/TemplatesList.js
brumik: ah, I was searching for this, but could not find anything... never thought of label, thanks

Ladas:
@brumik so to get the same counts:

Pass the main filter values of clusters page:
- cluster (cluster_id)
- date_range (pass exactly the same start_date and end_date)

Then other filters are:
- status (none selected, we filter all jobs)
- job_type: [Playbook run]

(count should fit, the list of jobs with `make data` will be probably different, since many jobs have the same `created` timestamp, the order gets kind of random. In prod, the probability that 2 jobs will have a same created down to microsecond is ~0)

brumik:
The numbers are the same now on both pages after clicking the link. @kialam @AlexSCorey

Issue #192, opened 2020-07-15
Job explorer table states are wrong


Ladas:
**Describe the bug**

We need to be showing the states tower provides which are

```
New
Pending
Waiting
Running
Successful
Failed
Error
Canceled
```

**To Reproduce**
1. Go to job explorer
2. Filter by all states except successful and failed
3. See that e.g. pending and error states are represented as failed

**Expected behavior**
I want to see the tower state name

**Screenshots**

![Screenshot from 2020-07-15 16-14-39](https://user-images.githubusercontent.com/1737058/87561279-ba4d4980-c6bc-11ea-9e03-280ef80930c9.png)

kialam:
@trahman73 do you think it makes sense to refactor the existing status component to reflect these job states, or should we just build out the new status pills in the mockups to support them:
![Screen Shot 2020-07-15 at 11 02 56 AM](https://user-images.githubusercontent.com/2293210/87561517-d0daad00-c68a-11ea-8dbb-e9d9b2aa5cd5.png)

trahman73:
@kialam lets map and build out the new status pills

trahman73:
![image](https://user-images.githubusercontent.com/13351591/87590587-03999b00-c6b5-11ea-877f-5922258b2ad3.png)

benthomasson:
We should only show terminal states in Automation Analytics because of the delay caused by periodic uploads.   We should only show these states for jobs:

- Successful
- Failed
- Error
- Canceled

Other states should be hidden until we get a real-time feed from Tower.

Ladas:
@benthomasson right, we were talking about that in another issue with @wenottingham , that we are not sure what is the last non-terminal state https://github.com/RedHatInsights/tower-analytics-backend/issues/356#issuecomment-617834471

But we should still show that the job didn't finish, since we can detect that. (you can set tower to send payload every 15 minutes now?) 15 mins is kind of realtime :-)


So the issue is more of that if the job is in state `running` for 5 days, our DB can have `pending` or `waiting` in some cases, but we can quite accurately say the job hasn't finished for 5 days.

Ladas:
@wenottingham ah, we haven't backported https://github.com/ansible/tower/pull/4180 https://github.com/ansible/tower/pull/4179 , so for tower < 3.7, we'll not receive the job updated status if it finishes at all :-D

We should get some eyes on these again (it is quite a big bug indeed)

benthomasson:
[...]

PR #206, opened 2020-07-24
Better job status component

brumik:
#192 

Added better component for displaying job status supporting: `successful, failed, error, running, new, waiting, pending, cancelled` statuses.

**After**
Screenshot from 2020-07-24 16-45-54


src/Components/JobStatus.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable */
```
kialam, line 1: Remove

kialam, line 1:
```suggestion
```

PR #207, opened 2020-07-27
Feature: Add link from ROI calculator to job expoler for templates

brumik:
https://github.com/RedHatInsights/tower-analytics-frontend/issues/183

**Before**
Screenshot from 2020-07-27 16-18-06

**After**
Screenshot from 2020-08-06 14-50-39

**After clicking on one:**
Screenshot from 2020-08-06 14-50-47

### DIFF
--- a/src/Containers/AutomationCalculator/AutomationCalculator.js
+++ b/src/Containers/AutomationCalculator/AutomationCalculator.js
@@ -9,6 +9,8 @@ import LoadingState from '../../Components/LoadingState';
 import NoData from '../../Components/NoData';
 import EmptyState from '../../Components/EmptyState';
 import { preflightRequest, readROI } from '../../Api';
+import { Paths } from '../../paths';
+import { formatQueryStrings } from '../../Utilities/formatQueryStrings';
 
 import {
     Main,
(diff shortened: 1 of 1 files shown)


Ladas:
@brumik so to get the same counts:

Pass the main filter values of clusters page:
- cluster (cluster_id)
- date_range (pass exactly the same start_date and end_date)

Then other filters are:
- status (none selected, we filter all jobs)
- job_type: [Playbook run]

(count should fit, the list of jobs with `make data` will be probably different, since many jobs have the same `created` timestamp, the order get kind of random. In prod, the probability that 2 jobs will have a same created down to microsecond is ~0)

brumik:
https://github.com/RedHatInsights/tower-analytics-frontend/pull/182 - waiting for dates

kialam:
@brumik is this PR now ready for review?

brumik:
Not yet, need to debase to Roi calculator and finish it

brumik:
There is only one date range that we can display in the job explorer @Ladas. So I did in a way that it displays always the last month and added a tooltip to the link saying that.

Issue #189, opened 2020-07-14
ROI calculator doesn't scale vertically to fill window


wenottingham:
**Describe the bug**

![Screenshot from 2020-07-14 13-17-58](https://user-images.githubusercontent.com/5208768/87464006-65d49c00-c5e0-11ea-834e-759d7dd0a446.png)

Window size: 2560x1440. Scale: 80%

**Expected behavior**

Full vertical usage (esp of the scrollable top templates)

**Desktop (please complete the following information):**
 - OS: Fedora 32
 - Browser: Firefox
 - Version: 78.0.1

wenottingham:
This is on beta with the pf4 upgrade, if that matters.

appuk:
![image](https://user-images.githubusercontent.com/11617351/90423160-fdb72100-e089-11ea-93f3-ebb637ed28a6.png)
 the top templates is now scaling vertically to fill the window, hence closing the issue

PR #208, opened 2020-07-28
force calculator height to 100%

jctanner:
Addresses https://github.com/RedHatInsights/tower-analytics-frontend/issues/189


jctanner:
The total height available to the Wrapper is actually coming from a higher component such as the patternfly drawer or drawercontent.

Issue #204, opened 2020-07-24
Handle 422 validation error nicely


Ladas:
**Describe the bug**
If we send invalid params into the API, it returns 422 code and we should print the error message to user

Right now the only case should be if we filter bigger date interval than max (max is 62 days)

**To Reproduce**
1. Go to job explorer
2. Click on date filter and select custom
3. Select from date to be 3 months ago (the distance between dates must be more than 62 days)
4. See table loading indefinitely

**Expected behavior**
A message from the API should be displayed, saying max 62 days can be filtered

**Screenshots**
![Screenshot from 2020-07-24 10-25-42](https://user-images.githubusercontent.com/1737058/88373758-b5f9ee00-cd98-11ea-9feb-39fd98fe3837.png)

brumik:
@Ladas Is there any more specification for "nicely" displaying the info?
Also, in this case, I think we should limit the calendar's selection range to 62 days (if it is possible).

Ladas:
@brumik riight, I am not really UX expert on what is nicely presented. @trahman73 is that ^ ok? Or should it be popup?

Yeah, so ideally we'd limit what you can select or we'd show you a popup with `You can select max 62 days with this precission, do you want to switch to 'monthly'  precision to display this range? (yes) (back)` . Or something in this sense. Not sure what is the right UX, but we should guide the user.

brumik:
I wasn't sure who else ask, but thanks for the mention ^. I was thinking about just showing 62 days and give an extra line there, that the displayed items are for 62 days as it is the maximum available.

trahman73:
@brumik @Ladas I agree that we should limit the actual selection but here is the PF alert example we would use.
![image](https://user-images.githubusercontent.com/13351591/88808899-70b33300-d181-11ea-884b-3c69e1958b54.png)

appuk:
I see the validation error  as expected
![image](https://user-images.githubusercontent.com/11617351/91064109-d669d200-e5fc-11ea-8756-c36bf82ea625.png)

@brumik any specific reason we did not go ahead with the PF alert modal?

brumik:
Yes, we had a discussion about it with @trahman73 . In short the problem with modal: what should be displayed after closing it.

appuk:
Ahh makes sense, closing as verified

PR #211, opened 2020-07-29
Catch api call errors and display the message nicely

brumik:
#204 

**Before**:
image

**After**:
Screenshot from 2020-08-03 17-03-02

### DIFF
--- a/src/Components/ApiErrorState.test.js
+++ b/src/Components/ApiErrorState.test.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import ApiErrorState from './ApiErrorState';
+
+describe('Components/JobStatus', () => {
+    let wrapper;
+    const msg = 'API Error Description...';
+
+    beforeEach(() => {
+        wrapper = null;
+    });
+
+    it('should render successful', () => {
+        wrapper = shallow(<ApiErrorState message={ msg } />);
+        expect(wrapper.length).toBe(1);
+    });
+
+    it('should contain the correct string', () => {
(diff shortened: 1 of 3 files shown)


src/Components/ApiErrorState.test.js
```diff
@@ -0,0 +1,16 @@
...
+describe('Components/JobStatus', () => {
+    let wrapper;
+
+    beforeEach(() => {
+        wrapper = null;
+    });
+
+    it('should render successful', () => {
+        wrapper = shallow(<ApiErrorState message="API Error Description..." />);
+        expect(wrapper.html()).toMatchSnapshot();
```
AlexSCorey, line 14: This is not wrong at all, but in my experience, while building a product that is not really in a stable form yet, the snapshots are not really useful because if 1 user legitimately changes the html the snapshot test fails as it should and so people just get in the habit of updating snapshots all the time and don't really look at them.  

for these components I do at least 2 tests.  1 is a smoke test that just determine that the `ApiErrorState.length === 1` and then I have an assertion that check that the text is right. 

This is not a blocker for merging this PR for me, just offering some of my own learning experiences.

brumik: Okey, I was use to checking the diffs and updating it but I can do your suggestion as it seems reasonable. I will also need to do the component as alert modal, so I will need to update the tests too anyways. Thanks for the feedback.

trahman73:
@brumik Based on your screenshot, the alert seems to be inset to the table instead of behaving like a modal, which is the desired state.

brumik:
@trahman73 If you like the screenshot I think this is good to go.

Issue #209, opened 2020-07-29
Job expoler's date range picker not working after custom date


brumik:
**Describe the bug**
When selecting a custom date in the filter then trying to select a predefined range (eg. Last 30 days, Last week, ...) the items are not changed - the filter ignores the change.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to `Job Explorer`
2. Click on `Filter -> by date -> custom`,
3. Select a custom date (for example 50 days)
4. Click on `Filter -> by date -> Last week`
5. The shown items are not changed

**Expected behaviour**
The filter should change the displayed items accordingly

**Desktop (please complete the following information):**
 - OS: Fedora 32
 - Browser: Chrome
 - Version: 84

appuk:
Verified this issue with following steps, working as expected:
1. Filter by Custom date
2. Filter by Last week
3. Filter for last week is applied successfully
Hence closing

PR #213, opened 2020-07-29
Fix: switching back from custom date in job expoler works properly

brumik:
#209 

**Before**: See issue ^

**After**:
Screencast from 07-29-2020 01_19_15 PM

### DIFF
--- a/src/Containers/JobExplorer/JobExplorer.js
+++ b/src/Containers/JobExplorer/JobExplorer.js
@@ -205,9 +205,13 @@ const JobExplorer = props => {
 
         setQuickDateRange(filters.date);
 
-        setStart_Date(filters.startDate);
-
-        setEnd_Date(filters.endDate);
+        if (filters.date !== 'custom') {
+            setStart_Date(null);
+            setEnd_Date(null);
+        } else {
+            setStart_Date(filters.startDate);
+            setEnd_Date(filters.endDate);
+        }
     }, [ filters ]);
 
     useEffect(() => {


src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -205,9 +205,13 @@ const JobExplorer = props => {
 
         setQuickDateRange(filters.date);
 
-        setStart_Date(filters.startDate);
-
-        setEnd_Date(filters.endDate);
+        if (filters.date !== 'custom') {
+            setStart_Date(undefined);
+            setEnd_Date(undefined);
```
AlexSCorey, line 210: What the purpose to using `undefined` as a argument for these function call.

brumik: We need to unset the filter, so the `undefined` keyword is given in js. We usually check variables by `if (variable)`. However, in this case, I see this is maybe not the case as this particular variable is checked by `=== null`. It still works because later setting the filter is checked by `variable ? true : false`. 

I think it should be undefined as we unsettling a filter and we should check for undefined and null instead of `=== null` in `useQueryParams.js` line 99 & 106

benthomasson: Please don't pass `undefined` as an argument.    If `null` will work use that.  Also passing nothing `setStart_Date()` is equivalent to passing undefined so use `setStart_Date()`  if `null` will not work.

brumik: Oki

Is there a reason not passing undefined that I am not aware of?

AlexSCorey: IMO it goes against convention and I think it raises more questions and could make the code a bit more difficult to maintain in the future without any added benefit.

PR #214, opened 2020-07-31
Sort by filter - array to single select

### LINKED ISSUE #195: V1 API doesn't support appending [] to filter params
**Describe the bug**
So e.g. if we sent filter `sort_by[]=id:asc` or `attributes[]='name'`, API will ignore it and fallback to default.

Because FastAPI is not able to parse this right and it takes `sort_by[]` as the name of the param. So all params must be send without `[]` suffix

**To Reproduce**
1. Go to job explorer
2. Sort by something
3. See it doesn't work and always sorts by `created` date, because we send `sort_by[]` and sort_by values are the default, because we send `attributes[]` to options filter

**Expected behavior**
Sending filters as `sort_by` and `attributes` the sorting will start to work properly.

**Screenshots**

Screenshot from 2020-07-15 18-58-14

brumik:
#195 

**Before**:
Screencast-from-07-31-2020-12_13_01-PM

**After**:
Screencast-from-07-31-2020-12_11_36-PM-NEW

### DIFF
--- a/src/Components/Toolbar.js
+++ b/src/Components/Toolbar.js
@@ -194,6 +194,14 @@ const FilterableToolbar = ({
             });
         };
 
+        const onSortSelect = (_event, selection) => {
+            handleFilters({
+                ...passedFilters,
+                sortby: selection
+            });
+            setSortByIsExpanded(!sortByIsExpanded);
+        };
+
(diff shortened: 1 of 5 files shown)


brumik:
cc @Ladas

PR #218, opened 2020-08-04
Add cluster_id to properly filter jobs from Line Chart.

kialam:
Related #219


brumik:
Screencast from 08-04-2020 02_28_22 PM

kialam:
@brumik seems like this is a timezone issue and not related to the PR changes. I'll card up the issue separately and we can go ahead and merge this if you think it's good to go.

Issue #196, opened 2020-07-17
How to represent default selected filters as tags in job explorer


Ladas:
# Description

Some of the filters do have default values (that can change based on selection). So we should figure out how to present these to users.

E.g. by default we order by `created:desc` and filter `last_30_days` for job explorer_job list

Once we are able to group by (e.g. group by template) and use granularity, that will switch the default values. So e.g. grouping by `template` and switching to `monthly` granularity, we'll switch to default: 
order by `total_count:desc` (total number of jobs)
and `last_3_months`

**Is your feature request related to a problem? Please describe.**
It's not clear what is a default selection of some filters, like time filters and sort_by 

So e.g. in this screenshot, I should see it's sorted by default by `created:desc` showing me last 30 days of data

![Screenshot from 2020-07-17 18-09-38](https://user-images.githubusercontent.com/1737058/87807672-d554c080-c858-11ea-89f9-3614e9d21790.png)

**Describe the solution you'd like**
We should probably show default values as greyed out tag? (having title description it's a default value) (We can't really remove it, since it's default, but we can select something else for e.g. sort_by, but by removing the tag it will jump back to default)

**Mock up**

Add mock up here when ready

**Related PRs**

Add PRs here when ready

# Verification

**Screenshot**

Add screenshot of implementation here when implemented

**Steps**

Add verification steps here when ready for QE

Ladas:
discuss @kialam @trahman73 @wenottingham

wenottingham:
For ordering, wouldn't the expected UX be that it's ordered by one of the columns, and that's how it's shown to the user? i.e., ordering is determined by the UI, not the API return.

If it's date ranged by default, that should definitely show in the filter.

trahman73:
I would recommend default view to be ID/Name descending. I would also not show the search tags for a default view until the user customized the filter themselves.

Ladas:
@trahman73 @wenottingham so I'd expect the default ordering to show latest jobs? So that should be either created or finished timestamp.

ID is tricky for multi-cluster, for one cluster, ordering by ID is basically the same as created timestamp (so that works nicely for tower). But for multi-cluster the biggest cluster will overshadow the smaller. Since e.g. biggest will have IDs around 100k and small around 10k, you'll see just biggest cluster's jobs and no new small cluster's jobs.

Ladas:
[...]

PR #224, opened 2020-08-04
Add default filters for job explorer.

kialam:
Related #196


kialam:
@brumik good catch, I knew I had left something out. I've now added the default `sort_by` filter.

@Ladas @trahman73 the UI now shows the correct set of default filters/sorting when they visit the job explorer list:
default_filters_ui

Ladas:
@trahman73 @kialam btw. I am able to remove the default filter, which is a bit confusing (since date range and sort_by is always there), but we can probably fix that in 1.2

Issue #220, opened 2020-08-04
Chart builder phase 1 `clusters dashboard chart`


Ladas:
# Description

Lets experiment with showing job explorer table as chart

We'd reuse existing chart we have in `clusters` page, we'd just make it configurable, so it can show us more values, small tasks
- [ ] **1. Switch job explorer table view to chart**
With a checkbox `show as chart` (@trahman73 )  ? Which would send attribute `group_by_time=True` to the API

- [ ] **2. Granularity selectbox**
By default the group_by_time returns records per day, we can control this with `granularity: daily` , so we can show it in chart as daily, monthly or yearly records 

- [ ] **3. Configure which attribute to show**
We have a list of numeric attrs available in `_options` endpoint, let user pick one. (API might need to expose what attrs are numeric and what is the Unit?)

- [ ] **4. Display mode**
If line chart and stacked bar charts are done, lets use them. (we use them for clusters dashoboard) Otherwise I'd start with line chart (we show it now, when we filter specific cluster). Then we can continue with stacked/grouped bar charts in other tasks.

- [ ] **5. Pack this as configurable component** 
So we can use this on other dashboards (cluster dashboard should use this), display more of these on 1 page, etc.

**Is your feature request related to a problem? Please describe.**
We should be able to explore the tabular data as charts in time.

**Describe the solution you'd like**
Lets do a switch to chart view, using our simplest chart in clusters dashboard

**Additional context**

We should do some experimentation in this first phase (will create separate tasks from these):

- [ ] **E1 How can we display multiple values in a chart**

This will be needed for anomaly detection correlation. We can experiment with multiselect box here. But we'll need to figure out how to display values with different units and scale.

We can allow user to pick several groups (max 3-4?) which would result in multiple y axis legends in 1 chart (see e.g. bottom of https://plotly.com/python/multiple-axes/ for inspiration)

And we should definitely allow user to render multiple charts on 1 page, so we can see multiple values correlated, with aligned x (time) axis.

**Mock up**

@trahman73 we'll need to see where to put the `show as chart` (or better naming) in the job explorer main filter. And that will also display the granularity selectbox.

Otherwise we'd use chart design as from https://tower-mockups.testing.ansible.com/analytics/dashboard/analytics/ clusters dashboard. This dashboard shows stacked bar chart and line chart when you filter specific cluster. I'd default to line chart.

**Related PRs**

Add PRs here when ready

# Verification

**Screenshot**

[...]

PR #230, opened 2020-08-05
Feature: Generalised Charts - Refactor

brumik:
Connected to #220 
Solves #81 
Added changes to match: #283 #284 #273 

## Visual and behavioral (maybe unwanted) changes
* Grouped bar chart has a black right border instead of open.
* Same orgs in different charts on the same page are the same color now

@kialam @Ladas 

This is how I started out with it. You can call the chart builder right now like this:
```
<ChartWrapper
  id="d3-grouped-bar-chart-root"
  data={ mainChartData }
  lineNames={ [ 'failed', 'successful' ] }
  colors={ [ '#6EC664', '#A30000', '#06C' ] }
  xAxis={ {
      type: 'date',
      text: 'Date'
  } }
  yAxis={ {
      type: 'number',
      text: currentChart().yLabel
  } }
  value={ clusterTimeFrame }
  onClick={ redirectToJobExplorer }
  chart={ currentChart().chart }
  tooltip={ tooltip({
      lineNames: [ 'failed', 'successful' ],
      colors: [ '#6EC664', '#A30000' ]
  }) }
/>
```

Right now if you check this out you get the same results as originally (with less code, more configuration at call) - Clusters page charts. The previous example is from there too.

**What it can do**:
* you can specify which object property to display from the passed object array
* unified x and y-axis data descriptors
* can set the label text on the chart
* working on the same data set
* you can pass custom tooltips to the same charts - WIP: How to make the tooltip easier to build?

**Colors**: Fixes (maybe only temporarily) the mismatching color for the same orgs in different chars by passing down the `scaleOrdinal` function from the parent instead of creating it every time in the chart.


src/Containers/Clusters/Clusters.js
```diff
@@ -215,24 +250,28 @@ const Clusters = ({ history }) => {
...
-                  lineChartData.length > 0 &&
-                  !isLoading && (
-                          <LineChart
-                              margin={ { top: 20, right: 20, bottom: 50, left: 70 } }
-                              id="d3-line-chart-root"
-                              data={ lineChartData }
+                      { !isLoading && (
+                          // change id to more general when find out how to modify css
+                          <ChartWrapper
+                              id="d3-grouped-bar-chart-root"
```
Ladas, line 256: btw. it this is the `clusters` chart, it's called `stacked bar chart`, since we stack bars on top of each other

Will the `id` need to be unique? So we can display more of these on page?

brumik: yeah, cannot make css recompile, if you check the code I wanted to make a `d3-chart-root` but I could not, so it will be replaced by a more proper name

brumik: You need unique id for each ChartWrapper though, but we can make it dynamically if we can pass a class for styling with our css

src/Containers/Clusters/Clusters.js
```diff
@@ -137,14 +153,33 @@ const Clusters = ({ history }) => {
...
                 setTemplatesData(templatesData);
                 setIsLoading(false);
             })
         );
     }, [ queryParams ]);
 
+    const currentChart = () => {
+        if (queryParams.id) {
+            return {
+                chart: lineChart,
```
Ladas, line 172: btw. line chart and stacked bar chart can be under 1 component. With a `chart_type` switch, then it'll be easier to make selecbox with the type and let the user switch.

brumik: It would be the same as passing a different function, and as I plan to move the other charts to this form too, I think it would just be confusing if we had `chart` prop and `chartType` prop too for some selected charts.

Ladas: ok, that makes sense

brumik:
@benthomasson It should be fine I think. If the extra right border is really unwanted I can patch it, but it would introduce quite a bit of copy-paste code, so even that this is a refactor I would like to leave it like this.

src/Charts/ChartWrapper.js
```diff
@@ -0,0 +1,211 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import * as d3 from 'd3';
+import Legend from '../Utilities/Legend';
+
+export const ChartWrapper = props => {
```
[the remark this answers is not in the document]
benthomasson: Use ...args to grab the rest of the props so we can pass them to the chart function.

benthomasson (changes requested):
I like the general idea.  I had a few changes.

brumik:
@benthomasson another round? 

I didn't figured out yet how to properly test the generated chart svg.

benthomasson (approved):
This looks good now.  I like breaking out the d3 code from the rest of the react code.

benthomasson:
We should probably wait on merging this until 1.1.0 testing is done.

benthomasson:
I'm closing this pull request since we are moving to victory charts in the future and this is tied to D3.

Issue #186, opened 2020-07-10
Switch component in Filterable Toolbar defaults to True sometimes


kialam:
**Describe the bug**
![Screen Shot 2020-07-10 at 9 47 56 AM](https://user-images.githubusercontent.com/2293210/87161394-76a8a900-c292-11ea-80ce-e1ff7902965d.png)

On occasion after a hard refresh, the switch component on the filterable toolbar that toggles the "ignore nested wf and jobs" option defaults to `true` instead of `false`. The accompanying console error says:
```
react_devtools_backend.js:6 Warning: Received the string `true` for the boolean attribute `checked`. Although this works, it will not work as expected if you pass the string "false". Did you mean checked={true}?
    in input (created by Switch)
    in label (created by Switch)
    in Switch (created by Context.Consumer)
    in StyledComponent (created by Styled(Switch))
    in Styled(Switch) (created by FilterableToolbar)
    in div (created by FilterableToolbar)
    in div (created by Context.Consumer)
    in div (created by ToolbarContent)
    in ToolbarContent (created by Context.Consumer)
    in StyledComponent (created by Styled(ToolbarContent))
    in Styled(ToolbarContent) (created by FilterableToolbar)
    in div (created by Toolbar)
    in Toolbar (created by FilterableToolbar)
    in FilterableToolbar (created by JobExplorer)
    in div (created by CardBody)
    in CardBody (created by JobExplorer)
    in article (created by Card)
    in Card (created by JobExplorer)
    in section (created by Context.Consumer)
    in Main (created by Connect(Main))
    in Connect(Main) (created by JobExplorer)
    in JobExplorer (created by AsyncComponent)
    in AsyncComponent (created by Route)
    in Route (created by InsightsRoute)
    in InsightsRoute (created by Routes)
    in Switch (created by Routes)
    in Routes (created by App)
    in div (created by App)
    in App (created by Connect(App))
    in Connect(App) (created by Route)
    in Route (created by withRouter(Connect(App)))
    in withRouter(Connect(App))
    in Router (created by BrowserRouter)
    in BrowserRouter
    in Provider
```

**To Reproduce**
1. Hard refresh the Job Explorer page.
2. Observe that the toggle for "ignore nested wf and jobs" at the top is set to enabled instead of the default disabled.

**Expected behavior**
Refreshing the Job Explorer page should reset all toggles back to their default states, e.g. the toggle should be disabled.

**Desktop (please complete the following information):**
[...]

PR #254, opened 2020-09-01
Quickfix 2.0 - switch default false on job explorer page 

brumik:
#186 

Could not find the exact cause of the problem, but checking with type for true solved it. The values passed to the component are seem like always correct.

### DIFF
--- a/src/Components/Toolbar.js
+++ b/src/Components/Toolbar.js
@@ -425,7 +425,7 @@ const FilterableToolbar = ({
                         id="showRootWorkflowJobs"
                         label="Ignore nested workflows and jobs"
                         labelOff="Ignore nested workflows and jobs"
-                        isChecked={ !!passedFilters.showRootWorkflows }
+                        isChecked={ passedFilters.showRootWorkflows }
                         onChange={ val => {
                             handleFilters({
                                 ...passedFilters,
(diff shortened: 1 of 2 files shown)


kialam (changes requested):
This code forces the param to be false always, so that if you were to pass it from a different component (e.g. from say the bar chart), it would be false regardless.

brumik:
It took me a while but you were right. The problem is that in the URL you cannot send boolean, so `true` or `false` after parsing the URL becomes `"true"` resp. `"false"`. I think we could maybe make the conversion in the `queryParams` component but after messing with it I could not get it to work. Or we can just convert all true and false to strings when used and in the toolbar set the `isChecked={ onlyRootWorkflows === 'true' }`

@kialam 

Also, it is good to go if you think so too.

benthomasson:
@kialam @brumik what is the state of this fix?

brumik:
It fixes the issue. What I wrote above is more of a consideration to make it more consistent in behavior or more bullet proof. 

So for me it could be merged and maybe do a follow-up pr if kia thinks so

benthomasson:
OK thanks!

benthomasson (approved):

kialam (approved):
LGTM

PR #256, opened 2020-09-07
[WIP] Api wrapper component - waiting for aprooval for continuation

brumik:
We talked a while ago about making unification of API requests. I know this is off right now, but I had only paper today and this was the only thing I was able to work on.

It was inspired by React's experimental suspense (https://reactjs.org/docs/concurrent-mode-suspense.html)

What do you think @kialam and @jeanetteobr? I personally think this component could help most of our use cases where we use the fetched data in one component from one API endpoint. However, it is a bit too much configuration and the component won't be able to cover more advanced use cases as it is used in the `render` function. 

I think @benthomasson you may be also interested.


brumik:
Closing to lack of interest in this component
@kialam

Issue #238, opened 2020-08-12
Job Explorer UI Cleanup Tasks


kialam:
# Description
Here are the individual tasks remaining for cleaning up the Job Explorer UI:
1. Refactor filter toolbar with the following changes:
- Break out Sort By into it's own dropdown
- Break out Date Range selector into it's own drop down
- Move "Ignore nested workflows..." toggle into filters dropdown
- Remove any default filters/chips when a user first visits the Job Explorer page.
Edit @brumik ^^ should be done as part of your filter toolbar component refactor for ease of reuse.

Current implementation:
<img width="1300" alt="Screen Shot 2020-08-11 at 8 46 34 PM" src="https://user-images.githubusercontent.com/2293210/89962788-c96bdc80-dc13-11ea-9506-b97e2f86af50.png">

Mockup:
<img width="858" alt="Screen Shot 2020-08-11 at 8 42 16 PM" src="https://user-images.githubusercontent.com/2293210/89962595-32068980-dc13-11ea-8419-d36db10555fe.png">

2. Update Pagination [PR OPEN]:
- Top of Job Explorer page should have the minimized pagination component.
- Bottom of Job Explorer page should have the default pagination component.
- Remove "Total Count" section at the top header.

Mockup:
<img width="1531" alt="new_pag" src="https://user-images.githubusercontent.com/2293210/89962897-0c2db480-dc14-11ea-8f84-90fbb199f072.png">

3. Update page header title from "Automation Analytics | All jobs" to "Jobs explorer" [PR OPEN].

4. Use compact version of datalist component [PR OPEN].

Current implementation:
<img width="859" alt="Screen Shot 2020-08-11 at 8 51 06 PM" src="https://user-images.githubusercontent.com/2293210/89963011-6d558800-dc14-11ea-91d3-33892b146e13.png">

Mockups:
<img width="682" alt="Screen Shot 2020-08-11 at 8 51 14 PM" src="https://user-images.githubusercontent.com/2293210/89963019-721a3c00-dc14-11ea-8cbf-edc4863f0e95.png">

5. Handle long drop downs -> set max height for the dropdown selector element with scrollbar [PR OPEN].

# Verification
Mockups here:
https://tower-mockups.testing.ansible.com/analytics/jobs/analytics-jobs/

appuk:
@jeanetteobr I see that the data in columns start a bit earlier than the column header

[...]

PR #268, opened 2020-09-16
Job Explorer UI Cleanup

jeanetteobr:
This PR handles 4 TODOs from issue #238:

2. Update Pagination:
- Top of Job Explorer page should have the minimized pagination component.
- Bottom of Job Explorer page should have the default pagination component.
- Remove "Total Count" badge from the top header.

3. Update page header title to "Automation Analytics | Jobs explorer".

4. Use the compact version of the datalist component

- remove the header labels when the window size is mobile. 
- Add DataCell item labels when window size is mobile.

Screen Shot 2020-08-17 at 3 50 31 PM

5. Handle long dropdown menus by setting a max-height and auto scroll.
before:
Screen Shot 2020-08-17 at 4 05 37 PM
after:
Screen Shot 2020-08-17 at 4 04 31 PM

### DIFF
--- a/src/Components/JobExplorerList.js
+++ b/src/Components/JobExplorerList.js
@@ -160,8 +160,8 @@ const AllJobsTemplate = ({ jobs }) => {
     const onResize = () => setWindowWidth(window.innerWidth);
 
     useEffect(() => {
-        window.addEventListener('resize', onResize());
-        return () => window.removeEventListener('resize', onResize());
+        window.addEventListener('resize', onResize);
+        return () => window.removeEventListener('resize', onResize);
     }, []);
 
     return buildListRow(jobs, 'All jobs view', 'all-jobs', windowWidth);
@@ -173,8 +173,8 @@ const JobExplorerList = ({ jobs }) => {
     const onResize = () => setWindowWidth(window.innerWidth);
 
     useEffect(() => {
-        window.addEventListener('resize', onResize());
-        return () => window.removeEventListener('resize', onResize());
+        window.addEventListener('resize', onResize);
+        return () => window.removeEventListener('resize', onResize);
     }, []);
 
     return (
--- a/src/Components/JobExplorerList.js
+++ b/src/Components/JobExplorerList.js
@@ -1,4 +1,5 @@
-import React, { useState } from 'react';
+/* eslint-disable no-console */
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
@@ -7,12 +8,12 @@ import {
     DataListContent,
     DataList,
     DataListItem,
-    DataListItemRow,
-    DataListItemCells,
+    DataListItemRow as PFDataListItemRow,
+    DataListItemCells as PFDataListItemCells,
     DataListToggle
 } from '@patternfly/react-core';
 
-import { ArrowIcon as PFArrowIcon } from '@patternfly/react-icons';
+import { ExternalLinkAltIcon as PFExternalLinkIcon } from '@patternfly/react-icons';
 
 import LoadingState from '../Components/LoadingState';
 import { formatDateTime, formatJobType } from '../Utilities/helpers';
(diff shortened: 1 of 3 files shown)


src/Components/JobExplorerList.js
```diff
@@ -1,4 +1,5 @@
-import React, { useState } from 'react';
+/* eslint-disable no-console */
```
brumik, line 1: I think this line is not needed

src/Components/Toolbar.js
```diff
@@ -343,7 +351,7 @@ const FilterableToolbar = ({
                         } }
                         selections={ passedFilters.template }
                         isExpanded={ templateIsExpanded }
-                        placeholderText="Filter by template"
+                        placeholderText="Filter by template"z
```
brumik, line 354: Trailing 'z' - accidental?

src/Components/JobExplorerList.js
```diff
@@ -110,25 +156,41 @@ const buildListRow = (items, ariaLabel, ariaLabelledBy) => {
...
-      <AllJobsTemplate jobs={ jobs } />
+const JobExplorerList = ({ jobs }) => {
+    const [ windowWidth, setWindowWidth ] = useState(window.innerWidth);
+
+    useEffect(() => {
+        return window.addEventListener('resize', () => setWindowWidth(window.innerWidth));
+    }, [ windowWidth ]);
+
+    return (
+        <>
```
brumik, line 176: I think it has a wrong indent

src/Components/JobExplorerList.js
```diff
@@ -110,25 +156,41 @@ const buildListRow = (items, ariaLabel, ariaLabelledBy) => {
...
-const JobExplorerList = ({ jobs }) => (
-  <>
-    { jobs.length <= 0 && <LoadingState /> }
-    <>
-      { buildHeader(headerLabels) }
-      <AllJobsTemplate jobs={ jobs } />
+const JobExplorerList = ({ jobs }) => {
+    const [ windowWidth, setWindowWidth ] = useState(window.innerWidth);
+
+    useEffect(() => {
```
brumik, line 171: I am not sure but I think you wanted this:
```js
    // Make a constant reference to the function
    const onResize = () => setWindowWidth(window.innerWidth);

    // On mount add event listener to update
    useEffect(() => {
        window.addEventListener('resize', onResize);

        // After unmount remove the event listener from the window.
        return () => window.removeEventListener('resize', onResize);
    }, []);
```
Maybe I am wrong but your code seems to create an event listener every time the window is resized *before* rerender (and would do it on a loop if react would actually run the code). Also, it generates console warning about not using cleanup.

src/Components/JobExplorerList.js
```diff
@@ -110,25 +156,41 @@ const buildListRow = (items, ariaLabel, ariaLabelledBy) => {
 };
 
 const AllJobsTemplate = ({ jobs }) => {
-    return buildListRow(jobs, 'All jobs view', 'all-jobs');
+    const [ windowWidth, setWindowWidth ] = useState(window.innerWidth);
+
+    useEffect(() => {
```
brumik, line 161: See the same bellow at `JobExpolerList`

brumik (changes requested):

src/Components/JobExplorerList.js
jeanetteobr: @brumik Hey, that's exactly what I was wanting. Thanks for the suggestion!

src/Components/JobExplorerList.js
```diff
@@ -110,25 +155,47 @@ const buildListRow = (items, ariaLabel, ariaLabelledBy) => {
...
 
 const AllJobsTemplate = ({ jobs }) => {
-    return buildListRow(jobs, 'All jobs view', 'all-jobs');
+    const [ windowWidth, setWindowWidth ] = useState(window.innerWidth);
+
+    const onResize = () => setWindowWidth(window.innerWidth);
+
+    useEffect(() => {
+        window.addEventListener('resize', onResize());
+        return () => window.removeEventListener('resize', onResize());
```
brumik, line 164: You just need to pass the reference to the event listener
```js
        window.addEventListener('resize', onResize);
        return () => window.removeEventListener('resize', onResize);
```

This way it is called when added and when removed but not when the window is resized

brumik (changes requested):
Except for that one thing it is working

brumik (approved):
LGTM

Issue #240, opened 2020-08-13
Switch to V1 API - orgs dashboard


Ladas:
# Description

Switching to v1 API, so we can use the job_explorer filter on the orgs dashboard

jobs_by_date_and_org_30
-----------------------------
`/api/tower-analytics/v1/job_explorer/` with POST params `group_by_time': True` and `group_by: 'org'`
<details>
<summary>Result</summary>

```
 'dates': [{'date': "2020-10-17",
                       'items': [{'canceled_count': 40,
                                  'elapsed': 3265.4,
                                  'error_count': 50,
                                  'failed_count': 118,
                                  'failed_host_count': 110,
                                  'host_count': 656,
                                  'host_task_changed_count': 116,
                                  'host_task_count': 1668,
                                  'host_task_failed_count': 110,
                                  'host_task_ok_count': 220,
                                  'host_task_retry_count': 336,
                                  'host_task_skipped_count': 336,
                                  'host_task_unreachable_count': 110,
                                  'id': 1,
                                  'name': 'organization_0',
                                  'new_count': 10,
                                  'pending_count': 70,
                                  'running_count': 60,
                                  'successful_count': 188,
                                  'total_cluster_count': 3,
                                  'total_count': 556,
                                  'total_org_count': 1,
                                  'total_template_count': 4,
                                  'unreachable_host_count': 110,
                                  'waiting_count': 20},
                                 {'canceled_count': 8,
                                  'elapsed': 734.1,
                                  'error_count': 10,
                                  'failed_count': 45,
                                  'failed_host_count': 174,
                                  'host_count': 964,
                                  'host_task_changed_count': 130,
                                  'host_task_count': 2208,
                                  'host_task_failed_count': 174,
                                  'host_task_ok_count': 312,
                                  'host_task_retry_count': 442,
                                  'host_task_skipped_count': 424,
                                  'host_task_unreachable_count': 138,
                                  'id': -2,
                                  'name': '',
                                  'new_count': 2,
                                  'pending_count': 14,
                                  'running_count': 12,
                                  'successful_count': 30,
                                  'total_cluster_count': 3,
                                  'total_count': 125,
                                  'total_org_count': 1,
                                  'total_template_count': 13,
                                  'unreachable_host_count': 138,
                                  'waiting_count': 4},
                                 {'canceled_count': 0,
                                  'elapsed': 2314.0,
                                  'error_count': 0,
                                  'failed_count': 132,
[...]

Issue #82, opened 2020-02-20
Org selection should be preserved across time filter changes


kdelee:
![org_selection_does_not_persist_after_time_filter_change](https://user-images.githubusercontent.com/29239279/74969994-99bde380-53eb-11ea-9abd-ac36f8847b0b.gif)

downselect orgs then change time filter, (this is with all orgs), expected that my selection would stay valid -- if orgs selected do not have data for that time period, show all orgs disabled, allow user to change org selection

jctanner:
migrated to https://issues.redhat.com/projects/AA/issues/AA-147

PR #291, opened 2020-09-24
Org dashboards update

brumik:
#240 
Fix for negative values: https://issues.redhat.com/browse/AA-131

* [x] Convert existing charts API to the v1`
* [x] Wait for https://github.com/RedHatInsights/tower-analytics-backend/pull/485 and pass proper data to piecharts
* [x] Same org same color in charts (this also solves #81 )
* [x] Display None/Null orgs (id = -2) in charts and under `No Organisation`
* [x] Use the new Toolbar (this also solves #82 )
* [x] Wait for the meeting about the filters and update the Toolbar
* [x] API Error display
* [x] Parse chart dates as a string, so there is no problem with the timezone

@Ladas

### DIFF
--- a/src/Containers/OrganizationStatistics/OrganizationStatistics.js
+++ b/src/Containers/OrganizationStatistics/OrganizationStatistics.js
@@ -13,7 +13,7 @@ import {
     readJobsByDateAndOrg,
     readJobRunsByOrg,
     readJobEventsByOrg,
-    readJobExplorerOptions
+    readOrgOptions
 } from '../../Api';
 
 import { Main, PageHeader, PageHeaderTitle } from '@redhat-cloud-services/frontend-components';
@@ -107,7 +107,7 @@ const OrganizationStatistics = ({ history }) => {
         setIsLoading(true);
         window.insights.chrome.auth.getUser().then(() => Promise.all(
             [
-                readJobExplorerOptions({ params: queryParams }),
+                readOrgOptions({ params: queryParams }),
                 readJobsByDateAndOrg({ params: queryParams }),
                 readJobRunsByOrg({ params: queryParams }),
                 readJobEventsByOrg({ params: queryParams })
@@ -120,9 +120,9 @@ const OrganizationStatistics = ({ history }) => {
         ]) => {
             if (didCancel) { return; }
 
-            const { sort_by, group_by, attributes, ...rest } = options;
+            const { meta, inventory_id, ...rest } = options;
 
-            setOptions({ ...rest, sort_by: constants.sort_by });
+            setOptions({ ...rest, sort_by: meta.sort_by });
             setorgsChartData(orgsChartMapper(orgsChartData));
             setPieChart1Data(pieChartMapper(pieChart1Data));
             setPieChart2Data(pieChartMapper(pieChart2Data));
--- a/src/Containers/OrganizationStatistics/OrganizationStatistics.js
+++ b/src/Containers/OrganizationStatistics/OrganizationStatistics.js
@@ -1,7 +1,9 @@
-import React, { useState, useEffect } from 'react';
+import React, { useEffect } from 'react';
+import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
 import { useQueryParams } from '../../Utilities/useQueryParams';
+import useApi from '../../Utilities/useApi';
 
 import LoadingState from '../../Components/LoadingState';
 import NoData from '../../Components/NoData';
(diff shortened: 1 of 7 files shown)


brumik:
Waiting for merging, otherwise ready: https://github.com/RedHatInsights/tower-analytics-backend/issues/59

kialam (changes requested):
We will want to reuse the filter toolbar on the org dashboards page as well. See #293 for a working first pass at that implementation. We will need to standardize how the filter toolbar gets reused on other pages.

kialam (changes requested):
The chart seems to be rendering the dates off by one day when custom dates are selected. The response body from the API is correct so this is something on the client side.

Screen Shot 2021-01-24 at 6 59 16 PM

src/Containers/OrganizationStatistics/OrganizationStatistics.js
```diff
@@ -64,117 +68,75 @@ const TopCard = styled(Card)`
...
-                    { usages: pieChart2Data = []}
-                ]) => {
-                    setGroupedBarChartData(groupedBarChartData);
-                    setPieChart1Data(pieChart1Data);
-                    setPieChart2Data(pieChart2Data);
-                    setIsLoading(false);
-                }
-            );
-        };
+    const orgsChartMapper = data => data.map(el => ({
```
kialam, line 95: Prefer destructuring here:

```
const orgsChartMapper = data => data.map(({ date, items }) => ({
    date: new Date(date),
    items: items.map(({ id, total_count, name }) => ({
        id,
        date: new Date(date),
        value: total_count,
        name: name ? name : 'No organization'
    }))
}));
```

src/Containers/OrganizationStatistics/OrganizationStatistics.js
```diff
@@ -64,117 +68,75 @@ const TopCard = styled(Card)`
...
+            { items: pieChart1Data = []},
+            { items: pieChart2Data = []}
+        ]) => {
+            if (didCancel) { return; }
+
+            const { meta, inventory_id, ...rest } = options;
+
+            setOptions({ ...rest, sort_by: meta.sort_by });
+            setOrgsChartData(orgsChartMapper(orgsChartData));
+            setPieChart1Data(pieChartMapper(pieChart1Data, 'host_count'));
```
kialam, line 133: Wondering if you could change the attributes in the request body sent to `readJobRunsByOrg` and `readJobEventsByOrg` instead of calling `pieChartMapper` to map the relevant attributes that are returned?

kialam (changes requested):
Toggling the "Ascending" and "Descending" does not seem to change the chart - I would assume that Descending would sort the data from greatest to least (aka Top 5), and Ascending would be the opposite (aka Bottom 5). Looking at the request body we are sending `id:asc` and `id:desc` and not `total_count`.

kialam:
I would also remove the extraneous call to `dashboard_organizations` whenever `queryParams` gets updated.

kialam:
Overall, good job on this! There are a few small bugs but this is very close to being merge-able.

src/Containers/OrganizationStatistics/OrganizationStatistics.js
brumik: I optimized the request and have the absolute minimum in the attributes (check out in `Api.js`), however here I need to map to the pie chart data format which is: `id, count, name` where the `count` in the first pie chart is `host_count` and in the second one `host_task_count`.

Maybe I did not understand what you want, but I think I cannot do that mapping in the request.

brumik:
> The chart seems to be rendering the dates off by one day when custom dates are selected. The response body from the API is correct so this is something on the client side.
> 
> Screen Shot 2021-01-24 at 6 59 16 PM

I cannot reproduce this

brumik:
> I would also remove the extraneous call to `dashboard_organizations` whenever `queryParams` gets updated.

I was also commenting on your PR that we should include. There Ladas said we can put it later in another pr, so I think it's not a bad thing to leave it in for now.

kialam:
It's an extra call for a functionality that has yet to be implemented. I would remove it for now.

kialam:
re: chart dates - this could be a timezone issue.

kialam:
@jeanetteobr @benthomasson can any of you reproduce the org chart off by one day bug above?

brumik:
@kialam I parse the date as a string so it would work now, can you check out if the date problem still persists?

jeanetteobr:
@kialam @brumik Levente's fix is working. I am not longer seeing the off by one day bug. :D

jeanetteobr (approved):

kialam (approved):

PR #298, opened 2020-09-28
Fix 227 timezone discrepancy

### LINKED ISSUE #227: Timezone discrepancy when clicking through from charts
**Describe the bug**
The formatted time appears to be off by a day when clicking through from the charts to the job explorer.

Related #218 

**To Reproduce**
1. Go to 'Clusters'
2. Click on a single bar to view Job Explorer
3. Observe that the specified day from the chart is off by one day.

**Expected behavior**
The selected dates for the charts and the job explorer should match.

**Screenshots**
timezone

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

kialam:
Fixes #227 

@Ladas @appuk 
Wondering if this change fixes the timezone off by one day issue - can you test this locally before I make the rest of the changes?

Basically we are currently converting UTC formatted date from API to local, and _then_ back to UTC, so causing an off by one day bug in the UI I think. This PR fixes the format ordering.

### DIFF
--- a/src/Utilities/helpers.js
+++ b/src/Utilities/helpers.js
@@ -9,7 +9,7 @@ export function formatDateTime(dateTime) {
 }
 
 export function formatDate(date) {
-    return moment(date).utc().format('YYYY-MM-DD');
+    return moment.utc(date).format('YYYY-MM-DD');
 }
 
 export function formatSeconds(seconds) {
--- a/src/Utilities/helpers.js
+++ b/src/Utilities/helpers.js
@@ -5,7 +5,7 @@ export function trimStr(str) {
 }
 
 export function formatDateTime(dateTime) {
-    return moment(new Date(dateTime).toISOString()).utc().format('M/D/YYYY h:mm:ssa');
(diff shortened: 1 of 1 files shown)


src/Utilities/helpers.js
```diff
@@ -9,7 +9,7 @@ export function formatDateTime(dateTime) {
 }
 
 export function formatDate(date) {
-    return moment(date).utc().format('YYYY-MM-DD');
+    return moment.utc(date).format('YYYY-MM-DD');
```
Ladas, line 12: looks like this function should work across the board for us, from the docs:

`moment.utc(...) is utc mode. Ambiguous input is assumed to be UTC. Unambiguous input is adjusted to UTC.`

kialam: Great, thanks for checking @Ladas ! I'll go ahead and apply this formula to the rest of the areas in the UI that handle dates.

benthomasson (approved):

PR #306, opened 2020-10-01
Add height and hidden overflow on templates dropdown.

jeanetteobr:
This PR handles a TODO item from issue #238:

5. Handle long dropdown menus by setting a max-height and auto-scroll.
before:
Screen Shot 2020-08-17 at 4 05 37 PM
after:


brumik (commented):
I think `<Select>` should be a styled component because we are adding inline styling to all of them

jeanetteobr:
@brumik  Totally agree. Thanks for the input.

brumik (approved):
LGTM

PR #307, opened 2020-10-02
Toolbar changes for ease of use

brumik:
#238 

@kialam 

TODO:
* [x] Datum custom selector --> sometimes null initialization warning
* [x] Remove chips for date and sort since they are visible always
* [x] ? Filter icon outside the first sort (since if it is not represented then there is no filter icon)
* [x] Get the visiblity prop correction to this PR too

Screenshot from 2020-10-09 16-47-27


src/Utilities/useQueryParams.js
```diff
@@ -10,121 +10,136 @@ export const useQueryParams = initial => {
     const paramsReducer = (state, action) => {
```
kialam, line 10: You could destructure out `value` here.

src/Utilities/useQueryParams.js
```diff
@@ -10,121 +10,136 @@ export const useQueryParams = initial => {
...
         switch (action.type) {
             case 'SET_STARTDATE':
```
kialam, line 12: `SET_STARTDATE` and `SET_ENDDATE` are used in the `v0` version of our API. With this new `urlMappedQueryParams()` it will always return `end_date` and `start_date` when passing these params in the URL which will break the v0 API since it does not accept snake case. I _think_ this should be fine, since we are going to be moving everything to `v1` so we could probably remove these dispatch methods since we already have `SET_START_DATE` and `SET_END_DATE` below.

kialam: I think Clusters, Org Stats, and ROI use `SET_STARTDATE` and `SET_ENDDATE` so we once we convert these views to `v1` of the API we can get rid of it.

brumik: Yes, I was thinking the same, we definitely not using BOTH of them at once.

src/Utilities/useQueryParams.js
brumik: I was trying to make something out of this value and most probably I will change it one more time, but if I leave it like this I definitely will follow your tip, thanks

src/Components/Toolbar/CategoryDropdown.js
```diff
@@ -0,0 +1,53 @@
...
+                onToggle={ () => setIsExpanded(!isExpanded) }
+                onSelect={ (event, selection) => {
+                    setSelected(selection);
+                    setIsExpanded(false);
+                } }
+                selections={ selected }
+                placeholderText={ 'Filter by' }
+            >
+                {
+                    categories.map(el => (
```
kialam, line 36: You can destructure `{ key, name }` here.

src/Components/Toolbar/CategoryDropdown.js
```diff
@@ -0,0 +1,53 @@
...
+import {
+    ToolbarItem,
+    Select,
+    SelectOption,
+    Button
+} from '@patternfly/react-core';
+import { FilterIcon } from '@patternfly/react-icons';
+
+const CategoryDropdown = ({
+    selected,
```
kialam, line 12: You have a no-op below and an empty array for `categories`; does `selected` need to have a default value as well?

brumik: `null` or `undefined` is fine for the selected default value, but if `categories` are undefined (which are for a brief moment when initializing values from the API) then it throws an error and dies.

src/Components/Toolbar/CategoryDropdown.js
```diff
@@ -0,0 +1,53 @@
...
+        <ToolbarItem>
+            <Button onClick={ () => setIsExpanded(true) } variant="control">
+                <FilterIcon />
+            </Button>
+            <Select
+                isOpen={ isExpanded }
+                variant={ 'single' }
+                aria-label={ 'Categories' }
+                onToggle={ () => setIsExpanded(!isExpanded) }
+                onSelect={ (event, selection) => {
```
kialam, line 28: Thoughts on using either `{ (, selection) => ...}` or `{ (_event, selection) => ... }` to denote useless variable here?

src/Components/Toolbar/CategoryDropdown.js
```diff
@@ -0,0 +1,53 @@
...
+                        </SelectOption>
+                    ))
+                }
+            </Select>
+        </ToolbarItem>
+    );
+};
+
+CategoryDropdown.propTypes = {
+    categories: PropTypes.array,
```
kialam, line 48: Should this be a required prop?

src/Components/Toolbar/CategoryDropdown.js
```diff
@@ -0,0 +1,53 @@
...
+}) => {
+    const [ isExpanded, setIsExpanded ] = useState(false);
+
+    return (
+        <ToolbarItem>
+            <Button onClick={ () => setIsExpanded(true) } variant="control">
+                <FilterIcon />
+            </Button>
+            <Select
+                isOpen={ isExpanded }
```
kialam, line 24: Wondering if this needs to be a constant that we import to set the details of this dropdown.

```
import { categoryDropdown.settings } as categorySettings from `.../constants`;

....
<Select
is Open={ isExpanded }
variant={ categorySettings.variant }
aria-label={ categorySettings.ariaLabel }
placeholderText={ categorySettings.placeholderText }
...
```

src/Components/Toolbar/CustomDateSelector.js
```diff
@@ -0,0 +1,55 @@
...
+    InputGroup,
+    InputGroupText,
+    TextInput
+} from '@patternfly/react-core';
+import {
+    CalendarAltIcon
+} from '@patternfly/react-icons';
+
+const CustomDateSelector = ({
+    onInputChange,
```
kialam, line 13: Do these need default values?

src/Components/Toolbar/CategoryDropdown.js
brumik: Since it has a default value... I think not, however, the component is pretty useless without it. I am not sure on this one, should I make it so? I don't think makes a difference

src/Components/Toolbar/CustomDateSelector.js
```diff
@@ -0,0 +1,55 @@
...
+              />
+          </InputGroup>
+        </>
+    );
+};
+
+CustomDateSelector.propTypes = {
+    startDate: PropTypes.string,
+    endDate: PropTypes.string,
+    onInputChange: PropTypes.func
```
kialam, line 52: Again, are these props required?

src/Components/Toolbar/CategoryDropdown.js
brumik: Right now I don't see where would we take advantage of this. This change, however, is small and local enough to add it later if the need arises.

src/Components/Toolbar/CustomDateSelector.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {
+    InputGroup,
+    InputGroupText,
+    TextInput
+} from '@patternfly/react-core';
```
kialam, line 7: We may need to start using this beta DatePicker from PF: https://www.patternfly.org/v4/components/date-picker/react/#ws-page-title

src/Components/Toolbar/CustomDateSelector.js
brumik: From the view of the functionality of the component: no. However, as stated above the component is useless without them.

src/Components/Toolbar/Toolbar.test.js
```diff
@@ -0,0 +1,41 @@
...
+                    jobType: [],
+                    orgId: [],
+                    clusterId: [],
+                    templateId: [],
+                    sortBy: []
+                } }
+                filters={ { status: null, quickDateRange: null } }
+            />);
+        const buttons = wrapper.find({ className: 'pf-c-select__toggle' });
+        expect(buttons.length).toBe(4);
```
kialam, line 35: Should this test for 2 or 4 buttons?

src/Components/Toolbar/ToolbarFilterItem.js
```diff
@@ -0,0 +1,94 @@
...
+    filter = null,
+    values = [],
+    visible = true,
+    setFilter
+}) => {
+    const [ expanded, setExpanded ] = useState(false);
+    const options = optionsForCategories[categoryKey];
+
+    const onDelete = (name, valueToDelete) => {
+        const single = optionsForCategories[categoryKey].single;
```
kialam, line 25: Should be `isSingle` since the value is a boolean.

src/Components/Toolbar/ToolbarFilterItem.js
```diff
@@ -0,0 +1,94 @@
...
+
+const ToolbarFilterItem = ({
+    categoryKey,
+    filter = null,
+    values = [],
+    visible = true,
+    setFilter
+}) => {
+    const [ expanded, setExpanded ] = useState(false);
+    const options = optionsForCategories[categoryKey];
```
kialam, line 22: Should this be singular `option`?

src/Components/Toolbar/Toolbar.test.js
```diff
@@ -0,0 +1,41 @@
...
+        const buttonTexts = buttons.map((b) => {
+            return b.text().trim();
+        });
+        expect(buttonTexts).toEqual(expect.arrayContaining([ 'Status' ]));
```
kialam, line 39: Do you need another `expect` statement for `quickDateRange` here?

src/Components/Toolbar/ToolbarFilterItem.js
```diff
@@ -0,0 +1,94 @@
...
+
+    const onDelete = (name, valueToDelete) => {
+        const single = optionsForCategories[categoryKey].single;
+
+        if (single) {
+            setFilter(null);
```
kialam, line 28: Would like to see a unit test to make sure `onDelete`, `onSelect`, and `handleChips` behave as expected.  I see different ways of calling `setFilters` with either a single `null` or a double `null` argument. Seems kinda odd to me.

src/Components/Toolbar/ToolbarFilterItem.js
brumik: I think no, there are multiple options for the same key

src/Components/Toolbar/Toolbar.test.js
brumik: Wrong naming maybe, it checks how many selects are there, not buttons (category, filter, date, sort)

src/Components/Toolbar/Toolbar.test.js
brumik: I update this test to check for date and sort too if possible

src/Components/Toolbar/ToolbarFilterItem.js
brumik: The component is created in null and `setFilter(type, value)` is used. When called `setFilter(null, null)` then the filter resets. When called `setFilter('status', value)` then the status is set to the value. To this component, there is a wrapper function which does reduces the function to `setFilter(value)` and presets the type of it (since the ToolbarFilterItem handles only ONE specific filter at a time)

brumik: If you want to see the implementation check out `useQuerryParam` reducers.

src/Components/Toolbar/CustomDateSelector.js
brumik: Shouldn't we wait until it comes out of beta? Updating this one component is pretty easy later too.

src/Components/Toolbar/CategoryDropdown.js
kialam: True, let's leave it for now. I prefer setting default values as opposed to marking props as required anyway - less prone to falling over unnecessarily.

src/Components/Toolbar/CategoryDropdown.js
kialam: Since `selected` is not a required prop ([see here](https://www.patternfly.org/v4/components/select#props)) the component will not throw an error or fail to mount. You call `map` on `categories` so if it's `undefined` it will throw an error.  I suppose it's fine to leave out. The default value would be `[]`.

src/Components/Toolbar/CustomDateSelector.js
kialam: @trahman73 thoughts on this? Should we start using the new date picker linked above? Seems like since it's not that hard to use we might as well do it. This is mostly a now vs later argument. There will be additional work to get `beta` tagged components to work in our code base I think.

src/Utilities/useQueryParams.js
```diff
@@ -7,126 +7,134 @@ import moment from 'moment';
...
+            case 'SET_END_DATE': {
+                let newValues = {};
+                Object.entries(value).forEach(([ key, value ]) => {
+                    newValues[key] = formatDate(value);
+                });
+                return { ...state, ...newValues };
+            }
 
-                return { ...state, end_date: formatDate(action.date) };
+            case 'RESET_FILTER':
```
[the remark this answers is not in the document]
kialam: Or, maybe this needs to be renamed `RESET_PARAMS`,

src/Components/Toolbar/ToolbarFilterItem.js
kialam: Ok, I think I need more clarity on the `setFilters()` method, since I'm seeing different implementations in the code:
```
setFilter(selection); <-- array

setFilter(event.target.checked <-- expression eval that returns either an array or calls another method
    ? [ ...filter, selection ]
    : filter.filter(value => value !== selection)
);

setFilter(null); <-- null one param

setFilter(null, null); <-- null 2 params
```
As a user, it is confusing to me how to use this method in the appropriate context.

src/Components/Toolbar/CategoryDropdown.js
```diff
@@ -0,0 +1,53 @@
...
+                variant={ 'single' }
+                aria-label={ 'Categories' }
+                onToggle={ () => setIsExpanded(!isExpanded) }
+                onSelect={ (_event, selection) => {
+                    setSelected(selection);
+                    setIsExpanded(false);
```
kialam, line 30: This is thoughtful -> close the select if it's a single selection but keep it open if it's multi. 👍

src/Components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,202 @@
...
+    setFilters,
+    pagination,
+    hasSettings = false
+}) => {
+    const [ currentCategory, setCurrentCategory ] = useState(
+        Object.keys(categories)[0]
+    );
+    const [ settingsExpanded, setSettingsExpanded ] = useState(false);
+
+    const { quickDateRange, sortBy, ...filterCategories } = categories;
```
kialam, line 54: 💅

src/Components/Toolbar/ToolbarFilterItem.js
brumik: Okey, you are looking at `setFilter` in different components.

1. `setFilter(null, null)` is not used.... the name is `setFilters(null, null)` - **mind the plural form** - and it is called in the `Toolbar` where we work with all the filters passed to the toolbar.
2. `setFilter(any value)` sets the value for ONE filter out of the multiple ones we use in the toolbar and it is only used in the `ToolbarFilterItem` component which takes care of only one of the filters. This function doens't have to do anything with the `setFilters` method in the toolbar since they are defined in different components.

You look at how the `ToolbarFilterItem` is called from the `Toolbar` you see this line:
```js
setFilter={ value => setFilters('quickDateRange', value) }
```

Here you can see the connection between the usage with one or two params. Just keep in mind, that we are talking about different components. I know on GitHub page it is hard to see through it, but when you need to open a different file you don't assume that the two names are connected.


3. About the expression: that code is mostly copied but lets look at it:
```js
setFilter(
// Ternary operator, depending if event.target.checked is true or false 
// returns the first or second expression supplied
event.target.checked
    // If checked is true, we return an array with the new selection
    ? [ ...filter, selection ]
    // If checked is false we return the filter array without the selection
    // the Array.filter() method returns a new array with values where the supplied function returns true
    : filter.filter(value => value !== selection)
);
```

So in the end we basically do `setFilter(array)` where the array is a new array containing the new values for the filter.

I hope I cleared it up for you. If you still have questions I am happy to answer. And maybe I recommend checking out the pr and look at it in your favorite text editor.

src/Utilities/useQueryParams.js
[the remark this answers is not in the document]
brumik: I was thinking about it. It was named first `RESET` only but I had to rename it. It seems to me we will have different variables in the `queryParam` per pages which don't have to include only filters. So we want to be able to reset only the filters. Since it is a reducer I think this should go here, because the logic is more easily looked through (when calling an action we say what we want and we control from the reducer the action)

I was also thinking about if `RESET_FILTER` should really clear all filters or restore the default ones?

I was thinking about making multiple reducers maybe for different components which would all add to query params, but I am not sure if we want it (at least now) and if it is feasible with `React.useReducer()`.
Something like this: https://redux.js.org/recipes/structuring-reducers/using-combinereducers

src/Components/Toolbar/CategoryDropdown.js
brumik: Thanks :)

src/Components/Toolbar/Toolbar.js
brumik: :confused: I have no nails to paint, but I guess you think it's pretty??

src/Components/Toolbar/Toolbar.js
```diff
@@ -0,0 +1,202 @@
...
+                clearAllFilters={ () => { setFilters(null, null); } }
+                collapseListedFiltersBreakpoint="xl"
+            >
+                <ToolbarContent>
+                    <ToolbarToggleGroup toggleIcon={ <FilterIcon /> } breakpoint="xl">
+                        { Object.keys(filterCategories).length > 0 && <FilterCategoriesGroup /> }
+                        { quickDateRange && <QuickDateGroup /> }
+                        { sortBy && <SortByGroup /> }
+                        {
+                            hasSettings &&
```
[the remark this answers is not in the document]
brumik: I think it would be confusing if there were no settings and the user could click on the hog and nothing would happen

src/Components/Toolbar/CustomDateSelector.js
brumik: It is not doable right now, the DatePicker does not expose onChange callback. There is an open PR for it in PF already but it will take some time (https://github.com/patternfly/patternfly-react/pull/5011)

kialam (approved):
🎉

PR #340, opened 2020-11-04
Fixed redirect problem: Converting url keys for queryParam to camelCase

brumik:
The `queryParam` is camelCase now so we need to convert the URL parsed params to camel case also.


kialam:
@brumik Can we add some quick unit tests to keep coverage up on helpers.js?

kialam (approved):
👍

PR #351, opened 2020-11-16
Job Explorer update with tests

brumik:
What this does:
* Updatest the dev deps for testing + adding `fetchMock` library
* Futureproofing: On filter value change query also the `jobExploererOptions` endpoint. As we talked about it, depending on the filter value some values may change in the future.
* Remove unused css code for pagination (double defining same styling)
* Adding tests to bump up test coverage
* Adding default value to the `meta` object from `job-explorer` API call


benthomasson (approved):

appuk:
@brumik Can you please clarify what changes does QE need to verify for this? does it require a regression test?

brumik:
@appuk I am not sure, but this is basically added testing (which doesn't need QE if i am right), and 2 little refractors on the Job Explorer page.

appuk:
yeah thought so too, thanks for the clarification!

Issue #354, opened 2020-11-16
Browser's back button behaviour with filters change


brumik:
@trahman73 @benthomasson @kialam @jctanner 
I didn't get a definitive answer yet so...:

What is the expected behaviour when using the browser's back button on a page where there is a filter?
1. Go back to the previous page
2. Go back to the previous filter state

In the case of 2. option, we need to synchronize the `history` and the `queryParams`.

benthomasson:
In general we should go to the previous filter state.  Except for the initial state of page should go to the previous page.  If the implementation adds filter states after landing on a page those states should be ignored for the purpose of the back button.

jctanner:
migrated to https://issues.redhat.com/projects/AA/issues/AA-197

Issue #128, opened 2020-04-21
Expose query params in URL


kialam:
# Description
As we add more list views to our app with multiple filters and sort options, we need a way to preserve the user's selection history as they navigate throughout various sections of the app.

**Is your feature request related to a problem? Please describe.**
This feature would allow users to save specific urls/views as bookmarks and reference points.

**Describe the solution you'd like**
Configure React Router to update browser history as users navigate, toggle, interact with various knobs that control the query params of a view.

**Describe alternatives you've considered**

**Additional context**
Related #127, #124, 

**Mock up**

Add mock up here when ready

**Related PRs**

Add PRs here when ready

# Verification

**Screenshot**

Add screenshot of implementation here when implemented

**Steps**

Add verification steps here when ready for QE

unlikelyzero:
@appuk when testing this, there is a string length limit to the length a url that is often smaller than the spec. The cloud.redhat.com team may have a better idea of what the realistic limit will be.

When projects/products implement this functionality, it's good to do an end-to-end test with the maximum number of characters allowed by a query filter, stacked with the max number of filters (plus some extra for future work).

Ladas:
The limit is usually about 2000 chars. So it should work for our first try of filtering/bookmarking. (and there will be limits e.g. for how many multiselect items you can select)

As a second round, we'll need to move queries to POST from GET and utimately and we'll need to make the query filter saveable (although, we'll still limit what you can filter, since it affects query speed).

unlikelyzero:
@Ladas thats the spec limit, but from my experiences there can be other limitations:

1. Some random nginx / tomcat config controlled by another team.
2. User Tracking Data can be appended (depending on the platform)

Ladas:
@unlikelyzero btw. if we'll go with a path having all parameters after `#`, then this is purely browser issue, correct? Since anything after # shouldn't be sent to server (the actual parameters will be send via POST request)

So that is usual way of solving this (e.g. kibana and others do this)

Ladas:
@kialam @jeanetteobr @benthomasson @bronaghs @jameswnl @AparnaKarve @unlikelyzero I am still not sure where do discuss if we change something in the API still. :-) 

This is the API example using POST for filters:
https://github.com/RedHatInsights/tower-analytics-backend/pull/353/files#diff-01963bb7191a6e399d529db848a57f0aR4

[...]

PR #355, opened 2020-11-18
Two way queryParam synch for 'back' in filters 

brumik:
@kialam @benthomasson 

#354 
#128 

It synchronizes the queryParams variable with the URL without polluting history so the back button works as expected.

**Problem:**
* Needed to update the `react-router-dom` package a major version, I didn't find any issues but probably needs more trough testing

**Gain:**
* Back button working as expected
* No need to update the URL from the pages, just the `queryParam`
* No need to parse URL and pass it to `queryParam` on page load as the initial value.

**How it does:**
* on the `queryParam` mounting checks if there is a search in the URL:
  * if there is then updates `queryParam` variable to match it
  * if not then updates the URL to match the query params (with REPLACING the URL so no pollution...)
* Further lifecycle
  * on URL change updates the `queryParam`
  * on `queryParam` change pushes to history.

**Todo**:
* [ ] Does not work right now with the clusters page (it has some problem with the cluster id)
* [ ] I don't like how the code looks in the `useQueryParams` component...
  * Maybe another hook from it to synch?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9992,25 +9992,16 @@
       "dev": true
     },
     "history": {
-      "version": "4.7.2",
-      "resolved": "https://registry.npmjs.org/history/-/history-4.7.2.tgz",
-      "integrity": "sha512-1zkBRWW6XweO0NBcjiphtVJVsIQ+SXF29z9DVkceeaSLVMFXHool+fdCZD4spDCfZJCILPILc3bm7Bc+HRi0nA==",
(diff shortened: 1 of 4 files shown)


brumik:
@kialam @benthomasson will we need this? Or in this form? Because if not we can close this

jlmitch5:
After discussing with @brumik, we're gonna close this pr for now, while keeping the branch in tact.

PR #359, opened 2020-11-23
Added useApi hook to simplify fetching

brumik:
@kialam It was a long time ago that we speak about this, but I had to made fetching research, and while at it I come up with this solution. (Now I understand why there were `ignore` at API fetching)

I think this hook could simplify fetching API. I can fetch one endpoint or multiple (`Promise.all`). It accepts a postprocessing function, allowing it to remap or modify data from API. It is also preventing memory leaks when the component is unmounted before API is fetched.

Returns and object with the following keys: 
* (bool) isLoading
* (string) errror - null if no error
* (bool) isSuccess - loaded without error --> this variable replaces the `!isLoading && !error`, making the code more readable.
* (object) data - the loaded data from the API

**Note**: Needed to update `JobExplorerList` component to use camelCase instead of snake_case, the `useApi` hook automatically converts the fetched data to it.

**New dev package**: https://react-hooks-testing-library.com/ for testing hooks (async hooks too ^^) with ease.

cc @benthomasson

### DIFF
--- a/src/Containers/JobExplorer/JobExplorer.js
+++ b/src/Containers/JobExplorer/JobExplorer.js
@@ -154,7 +154,7 @@ const JobExplorer = ({
                                 }
                                 hasSettings
                             />
-                            { error && <ApiErrorState message={ error } /> }
+                            { error && <ApiErrorState message={ error.error } /> }
                             { isLoading && <LoadingState /> }
                             { isSuccess && data.length <= 0 && <NoResults /> }
                             { isSuccess && data.length > 0 && <JobExplorerList jobs={ data } /> }
--- a/src/Containers/JobExplorer/JobExplorer.js
+++ b/src/Containers/JobExplorer/JobExplorer.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { parse, stringify } from 'query-string';
 
 import { useQueryParams } from '../../Utilities/useQueryParams';
+import useApi from '../../Utilities/useApi';
 import { Paths } from '../../paths';
 
 import LoadingState from '../../Components/LoadingState';
@@ -45,17 +46,24 @@ const initialQueryParams = {
     attributes: jobExplorer.attributes
 };
 
+const optionsMapper = options => {
+    const { groupBy, attributes, ...rest } = options;
+    return rest;
+};
+
 const JobExplorer = ({
     location: { search },
     history
 }) => {
     const [ preflightError, setPreFlightError ] = useState(null);
-    const [ apiError, setApiError ] = useState(null);
-    const [ isLoading, setIsLoading ] = useState(true);
-    const [ jobExplorerData, setJobExplorerData ] = useState([]);
-    const [ meta, setMeta ] = useState({});
+    const [{
+        isLoading,
+        isSuccess,
+        error,
+        data: { meta = {}, items: data = []}
+    }, setData ] = useApi({ meta: {}, items: []});
     const [ currPage, setCurrPage ] = useState(1);
-    const [ explorerOptions, setExplorerOptions ] = useState({});
+    const [ options, setOptions ] = useApi({}, optionsMapper);
 
     let initialSearchParams = parse(search, { arrayFormat: 'bracket', parseBooleans: true });
     let combined = { ...initialQueryParams, ...initialSearchParams };
@@ -87,32 +95,9 @@ const JobExplorer = ({
     }, []);
 
     useEffect(() => {
-        setApiError(null);
-        setIsLoading(true);
-        let didCancel = false;
-        window.insights.chrome.auth.getUser()
-        .then(() => {
-            Promise.all([
-                readJobExplorer({ params: queryParams }),
-                readJobExplorerOptions({ params: queryParams })
-            ]).then(([
-                { items: jobExplorerData = [], meta = {}},
-                options
-            ]) => {
-                if (didCancel) { return; }
-
-                setJobExplorerData(jobExplorerData);
-                setMeta(meta);
-
-                setExplorerOptions(options);
-            })
-            .catch(e => setApiError(e.error))
-            .finally(() => {
-                updateURL();
-                setIsLoading(false);
-            });
-        });
-        return () => didCancel = true;
+        setData(readJobExplorer({ params: queryParams }),);
+        setOptions(readJobExplorerOptions({ params: queryParams }));
+        updateURL();
     }, [ queryParams ]);
 
     const returnOffsetVal = page => (page - 1) * queryParams.limit;
@@ -131,7 +116,7 @@ const JobExplorer = ({
     };
 
     return (
-        <>
+        <React.Fragment>
             <PageHeader>
                 <PageHeaderTitle title={ 'Job Explorer' } />
             </PageHeader>
@@ -147,7 +132,7 @@ const JobExplorer = ({
                     <Card>
                         <CardBody>
                             <FilterableToolbar
-                                categories={ explorerOptions }
+                                categories={ options.data }
                                 filters={ queryParams }
                                 setFilters={ setFromToolbar }
                                 pagination={
@@ -169,10 +154,10 @@ const JobExplorer = ({
                                 }
                                 hasSettings
                             />
-                            { apiError && <ApiErrorState message={ apiError } /> }
-                            { !apiError && isLoading && <LoadingState /> }
-                            { !apiError && !isLoading && jobExplorerData.length <= 0 && <NoResults /> }
-                            { !apiError && !isLoading && jobExplorerData.length > 0 && (<JobExplorerList jobs={ jobExplorerData } />) }
+                            { error && <ApiErrorState message={ error } /> }
+                            { isLoading && <LoadingState /> }
+                            { isSuccess && data.length <= 0 && <NoResults /> }
+                            { isSuccess && data.length > 0 && <JobExplorerList jobs={ data } /> }
                             <Pagination
                                 itemCount={ meta && meta.count ? meta.count : 0 }
                                 widgetId="pagination-options-menu-bottom"
@@ -192,7 +177,7 @@ const JobExplorer = ({
                     </Card>
                 </Main>
             ) }
-        </>
+        </React.Fragment>
     );
 };
 
--- a/src/Utilities/useApi.js
+++ b/src/Utilities/useApi.js
@@ -30,7 +30,7 @@ const dataFetchReducer = (state, action) => {
 };
 
 const useApi = (initialData, postprocess = d => d) => {
-    const [ url, setUrl ] = useState(null);
+    const [ request, setRequest ] = useState(null);
 
     const [ state, dispatch ] = useReducer(dataFetchReducer, {
         isLoading: false,
@@ -41,7 +41,7 @@ const useApi = (initialData, postprocess = d => d) => {
 
     useEffect(() => {
         // Prevent fetching nothing
-        if (!url) {
+        if (!request) {
             return;
         }
 
@@ -50,7 +50,7 @@ const useApi = (initialData, postprocess = d => d) => {
         dispatch({ type: 'FETCH_INIT' });
 
         // Fetching
-        window.insights.chrome.auth.getUser().then(() => url.then(data => {
+        window.insights.chrome.auth.getUser().then(() => request.then(data => {
             if (!didCancel) {
                 dispatch({
                     type: 'FETCH_SUCCESS',
@@ -64,9 +64,9 @@ const useApi = (initialData, postprocess = d => d) => {
         }));
 
         return () => { didCancel = true; };
-    }, [ url ]);
+    }, [ request ]);
 
-    return [ state, setUrl ];
+    return [ state, setRequest ];
 };
 
 export default useApi;
--- a/src/Utilities/useApi.js
+++ b/src/Utilities/useApi.js
@@ -1,0 +1,74 @@
+import { useState, useEffect, useReducer } from 'react';
+
+const dataFetchReducer = (state, action) => {
+    switch (action.type) {
+        case 'FETCH_INIT':
+            return {
+                ...state,
+                isLoading: true,
+                isSuccess: false,
+                error: null
+            };
+        case 'FETCH_SUCCESS':
+            return {
+                ...state,
+                isLoading: false,
+                isSuccess: true,
+                error: null,
+                data: action.payload
+            };
+        case 'FETCH_FAILURE':
+            return {
+                ...state,
+                isLoading: false,
+                isSuccess: false,
+                error: action.payload
+            };
+        default:
+            throw new Error();
+    }
+};
+
+const useApi = (initialData, postprocess = d => d) => {
+    const [ url, setUrl ] = useState(null);
+
+    const [ state, dispatch ] = useReducer(dataFetchReducer, {
+        isLoading: false,
+        isSuccess: false,
+        error: null,
+        data: initialData
+    });
+
+    useEffect(() => {
+        // Prevent fetching nothing
+        if (!url) {
+            return;
+        }
+
+        // Initialize
+        let didCancel = false;
+        dispatch({ type: 'FETCH_INIT' });
+
+        // Fetching
+        window.insights.chrome.auth.getUser().then(() => Promise.all(
+            Array.isArray(url) ? url : [ url ]
+        ).then(data => {
+            if (!didCancel) {
+                dispatch({
+                    type: 'FETCH_SUCCESS',
+                    payload: postprocess(Array.isArray(url) ? data : data[0])
+                });
+            }
+        }).catch(e => {
+            if (!didCancel) {
+                dispatch({ type: 'FETCH_FAILURE', payload: e.error });
+            }
+        }));
+
+        return () => { didCancel = true; };
+    }, [ url ]);
+
+    return [ state, setUrl ];
+};
+
+export default useApi;
--- a/src/Utilities/useApi.js
+++ b/src/Utilities/useApi.js
@@ -50,13 +50,11 @@ const useApi = (initialData, postprocess = d => d) => {
         dispatch({ type: 'FETCH_INIT' });
 
         // Fetching
-        window.insights.chrome.auth.getUser().then(() => Promise.all(
-            Array.isArray(url) ? url : [ url ]
-        ).then(data => {
+        window.insights.chrome.auth.getUser().then(() => url.then(data => {
             if (!didCancel) {
                 dispatch({
                     type: 'FETCH_SUCCESS',
-                    payload: postprocess(Array.isArray(url) ? data : data[0])
+                    payload: postprocess(data)
                 });
             }
         }).catch(e => {
--- a/src/Utilities/useApi.js
+++ b/src/Utilities/useApi.js
@@ -59,7 +59,7 @@ const useApi = (initialData, postprocess = d => d) => {
             }
         }).catch(e => {
             if (!didCancel) {
-                dispatch({ type: 'FETCH_FAILURE', payload: e.error });
+                dispatch({ type: 'FETCH_FAILURE', payload: e });
             }
         }));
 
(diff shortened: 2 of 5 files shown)


brumik:
@benthomasson Do we still need a preflight error? Can we do the preflight error check and the `auth.getUser()` in the App component itself instead of every page? (We could add a page to the `react-router` for authentication errors and that way we don't need to deal with it on all other pages.)

src/Components/JobExplorerList.js
```diff
@@ -105,14 +105,14 @@ const buildListRow = (items, ariaLabel, ariaLabelledBy, windowWidth) => {
                             <DataListItemCells
                                 dataListCells={ [
                                     <DataListCell key={ count++ }>
-                                        <a href={ item.id.tower_link } target='_blank' rel='noopener noreferrer'>
+                                        <a href={ item.id.towerLink } target='_blank' rel='noopener noreferrer'>
```
benthomasson, line 108: Are these changes unrelated to the useApi hook?

benthomasson:
This is similar to useRequest in AWX:  https://github.com/ansible/awx/blob/5e51dd2ff7cb2db3f2b56435f47c1054bcc8b614/awx/ui_next/src/util/useRequest.js#L21

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -91,30 +100,9 @@ const JobExplorer = ({
...
-                const { attributes, groupBy, ...rest } = keysToCamel(options);
-                setExplorerOptions(rest);
-            })
-            .catch(e => setApiError(e.error))
-            .finally(() => {
-                updateURL();
-                setIsLoading(false);
-            });
-        });
+        setData(readJobExplorer({ params: urlMappedQueryParams }),);
```
benthomasson, line 103: Does this line block or does it fetch the data and options in parallel?

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -91,30 +100,9 @@ const JobExplorer = ({
...
+        setOptions(readJobExplorerOptions({ params: urlMappedQueryParams }));
+        updateURL();
```
benthomasson, line 105: If the two lines above don't block I am worried that this gets called too early.

src/Utilities/useApi.js
```diff
@@ -0,0 +1,80 @@
...
+
+    useEffect(() => {
+        // Prevent fetching nothing
+        if (!url) {
+            return;
+        }
+
+        // Initialize
+        let didCancel = false;
+        dispatch({ type: 'FETCH_INIT' });
```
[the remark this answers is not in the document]
brumik: Yes

src/Utilities/useApi.js
```diff
@@ -0,0 +1,80 @@
...
+
+        // Fetching
+        window.insights.chrome.auth.getUser().then(() => Promise.all(
+            Array.isArray(url) ? url : [ url ]
+        ).then(data => {
+            if (!didCancel) {
+                dispatch({
+                    type: 'FETCH_SUCCESS',
+                    payload: postprocess( // Apply the postprocess function.
```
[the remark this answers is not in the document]
brumik: It is optional, it gives an option to run maps or other postprocessing without needing to destructure and save the variables elsewhere (see the job explorer options which we can pass to the toolbar itself, but we need to throw out or add new keys to it)

I don't understand the second comment.

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -91,30 +100,9 @@ const JobExplorer = ({
...
     }, [ queryParams ]);
```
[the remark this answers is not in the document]
brumik: It is, the function updates every time queryParam is changed and returns a new object reference. React is capable to detect it, so it can be passed to use effect.

src/Containers/JobExplorer/JobExplorer.js
brumik: Yes, it does, since the two calls are not dependent on each other, there is no problem not using `Promise.all` (which basically does the same, but waits for both of them to finish before calling the callback function)

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -91,30 +100,9 @@ const JobExplorer = ({
...
-                setExplorerOptions(rest);
-            })
-            .catch(e => setApiError(e.error))
-            .finally(() => {
-                updateURL();
-                setIsLoading(false);
-            });
-        });
+        setData(readJobExplorer({ params: urlMappedQueryParams }),);
+        setOptions(readJobExplorerOptions({ params: urlMappedQueryParams }));
```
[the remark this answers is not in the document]
brumik: I don't think we handle the option fail yet in the URL, I would say that the toolbar remains empty since it is passed there. Otherwsie the api sets the `options = { error: 'Some error', isLoading: false, isSuccess: false }` so we could do a more proper error handling if we wish.

src/Containers/JobExplorer/JobExplorer.js
brumik: No, `updateURL` is dependent only on the `queryParams`, we don't need the options or data to know what goes to the URL. So it could be updated before the fetching is finished.

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -46,17 +47,25 @@ const initialQueryParams = {
     attributes: jobExplorer.attributes
 };
 
+const optionsMapper = options => {
+    /* eslint-disable-next-line */
```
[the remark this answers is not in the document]
brumik: Because `no-unused-var` rule. We are basically throwing away parts of the object. We could do it with the `delete` keyword, but it's much more verbose and much less clear.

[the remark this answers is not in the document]
benthomasson: Try this: https://eslint.org/docs/rules/no-unused-vars#ignorerestsiblings

src/Utilities/useApi.js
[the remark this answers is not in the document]
benthomasson: I was probably wrong in the second comment now that I think about it again. So don't worry about it.

benthomasson:
@brumik If we want to make the same API call twice in a row with the same params would this be able to do that?

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -46,17 +47,25 @@ const initialQueryParams = {
...
-    const [ apiError, setApiError ] = useState(null);
-    const [ isLoading, setIsLoading ] = useState(true);
-    const [ jobExplorerData, setJobExplorerData ] = useState([]);
-    const [ meta, setMeta ] = useState({});
+    const [{
+        isLoading,
+        isSuccess,
+        error,
+        data: { meta = {}, items: data = []}
+    }, setData ] = useApi(null, { meta: {}, items: []});
```
benthomasson, line 66: The defaults values and the destructuring on the previous line are almost the same.    Is it necessary to do the rename?

src/Utilities/useApi.js
```diff
@@ -0,0 +1,80 @@
...
+            };
+        default:
+            throw new Error();
+    }
+};
+
+const useApi = (initialUrl, initialData, postprocess = d => d) => {
+    const [ url, setUrl ] = useState(initialUrl);
+
+    const [ state, dispatch ] = useReducer(dataFetchReducer, {
```
benthomasson, line 36: This is a cool way to use useReducer.

brumik:
@benthomasson
TL;DR: Yes, but forcing always feels hacky.

If `React's` shallow compare detects the change in the URL (which can be done, but feels really anti-pattern) it can refetch. It is using a `useEffect` hook internally to detect changes and refetch, so as long as it fires a `useEffect` it will refetch.

I cannot really imagine how we would force it without changing a reference that would fire a `useEffect` without this component. We are currently using `effects` as well for fetching, so it would be the same - force to fire an effect - problem as with this hook.

src/Containers/JobExplorer/JobExplorer.js
brumik: No, this use case is more of a showcase how we can use it. (Okay, maybe yes, I am not sure, I wrote lots of tests to check if API returns an error, and for that maybe we need the destructuring part too - so ui doesn't break)

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -46,17 +47,25 @@ const initialQueryParams = {
...
     history
 }) => {
     const [ preflightError, setPreFlightError ] = useState(null);
-    const [ apiError, setApiError ] = useState(null);
-    const [ isLoading, setIsLoading ] = useState(true);
-    const [ jobExplorerData, setJobExplorerData ] = useState([]);
-    const [ meta, setMeta ] = useState({});
+    const [{
+        isLoading,
+        isSuccess,
```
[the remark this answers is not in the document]
brumik: It is redundant, but it is really simple and using a reducer so the redundancy is not painful to manage. On the other hand, we are already checking for too many variables *inline the render functions* and this variable could clean it up a little bit. 

It is more straighforward to write: `if isSuccess` than `if !loading && error === null`

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -46,17 +47,24 @@ const initialQueryParams = {
...
+        error,
+        data: { meta = {}, items: data = []}
+    }, setData ] = useApi(null, { meta: {}, items: []});
     const [ currPage, setCurrPage ] = useState(1);
-    const [ explorerOptions, setExplorerOptions ] = useState({});
+    const [ options, setOptions ] = useApi(null, {}, optionsMapper);
```
benthomasson, line 67: Why do we need the initial URL?  In all examples here it is passed in as null.   Can we remove it?

benthomasson (changes requested):

kialam:
Related #376

src/Containers/JobExplorer/JobExplorer.js
kialam: I'm having trouble figuring out how passing `null` as your initial URL calls the hook given [these lines](https://github.com/RedHatInsights/tower-analytics-frontend/pull/359/files#diff-31bc939331c6e5b934a01d1d52249903bd424c43e01a8aa64eff3613715ca286R45-R47).

src/Containers/JobExplorer/JobExplorer.js
kialam: Same as comment above.

src/Utilities/useApi.js
```diff
@@ -0,0 +1,80 @@
...
+
+        // Fetching
+        window.insights.chrome.auth.getUser().then(() => Promise.all(
+            Array.isArray(url) ? url : [ url ]
+        ).then(data => {
+            if (!didCancel) {
+                dispatch({
+                    type: 'FETCH_SUCCESS',
+                    payload: postprocess( // Apply the postprocess function.
+                        keysToCamel(
```
kialam, line 61: I don't understand the purpose of this helper function in this hook. IMO the parent component level should be responsible for handling the data and the fetch hook should just be responsible for passing the data returned.

src/Utilities/useApi.js
```diff
@@ -0,0 +1,80 @@
...
+                            // If single url give return a single object instead of array.
+                            Array.isArray(url) ? data : data[0]
+                        )
+                    )
+                });
+            }
+        }).catch(e => {
+            if (!didCancel) {
+                dispatch({ type: 'FETCH_FAILURE', payload: e.error });
```
kialam, line 70: Shouldn't you pass just `e` here? Not all API responses have an `error` attribute.

src/Utilities/useApi.js
```diff
@@ -0,0 +1,80 @@
...
+            }
+        }));
+
+        return () => { didCancel = true; };
```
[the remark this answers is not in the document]
kialam: Yes, glad you are reusing existing logic in the code that I wrote.

kialam (changes requested):
I would like to see this implemented on the Clusters page to see whether it simplifies fetching or not.

kialam:
@brumik how would this hook handle multiple API calls to the `job_explorer` endpoint with various different params?

src/Utilities/useApi.js
brumik: I was not sure how the error is returned in all the cases.

src/Utilities/useApi.js
brumik: We often do mapping in the components. We can pass the mapping function here. This would ensure that the same mapping function is used, no matter how many multiple places we query the API.

src/Containers/JobExplorer/JobExplorer.js
brumik: Id does not fetch the API, but the variables are initialized nevertheless.

brumik:
@kialam You can create to the same endpoints as many hooks you need. Also, the hook supports promise all, so if you wish you can pass an array of endpoints and receive an array of results (However I don't think in our app there is a place where we need to synch API call returns.)

brumik:
@kialam I made the example for the clusters page, however, I get the feeling you were thinking about your PR with the v1 api right?

src/Utilities/useApi.js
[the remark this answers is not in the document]
benthomasson: We should do more code sharing.  Feel free to borrow code and patterns from any open source project.  That is the main point of open source after all: learning and understanding.

src/Utilities/useApi.js
benthomasson: I do think the error handling here needs some work.  Maybe we pass in an error handler function to useAPI with you did with `postProcess`.

src/Containers/JobExplorer/JobExplorer.js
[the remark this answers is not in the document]
kialam: Agree with @benthomasson that this could live within the parent component level since it handles the rendering logic.

```
const isSuccess = !isLoading && !error;
```

src/Utilities/useApi.js
kialam: The mapping should be handled on the parent component level since each child component handles the data in their own unique way. Then again, all this does is convert snake case key value pairs to camel case so we avoid lengthy eslint ignore comments so 🤷 .

src/Utilities/useApi.js
```diff
@@ -0,0 +1,80 @@
...
+            return;
+        }
+
+        // Initialize
+        let didCancel = false;
+        dispatch({ type: 'FETCH_INIT' });
+
+        // Fetching
+        window.insights.chrome.auth.getUser().then(() => Promise.all(
+            Array.isArray(url) ? url : [ url ]
```
kialam, line 55: Do we need this logic? None of the implementation examples pass more than one request to the hook.

brumik: I know, but I think it's nice to have an option to use the hook as Promise.all. It feels futureproof for me :D.

brumik: But I can remove it if you insist

src/Utilities/useApi.js
benthomasson: Never mind about passing the error handler function.   It is not necessary.

src/Utilities/useApi.js
benthomasson: Add one example of its use somewhere to see how it works with multiple urls.

src/Utilities/useApi.js
brumik: @benthomasson Said he likes the idea of some kind of mapping function passed down. It is not required to use but I like the possibility too. Can save a few parent component's state destructurings.

src/Utilities/useApi.js
kialam: Often what we think of "futureproofing" is just needless abstraction/complication. How does your `postprocess` handle mapping of data from multiple asynchronous sources?

kialam:
A lot of this uses `useReducer` to return an object from `useAPI` hook. I can also see a much more simplified version of this without the `useReducer` hook:

```
import { useState, useEffect } from 'react';

const useApi = (request) => {
	const [ res, setRes ] = useState([]);
	const [ isLoadinig, setIsLoading ] = useState(true);
	const [ error, setError ] = useState(null);

    useEffect(() => {
        // Initialize
        let didCancel = false;
		
        // Fetching
        window.insights.chrome.auth.getUser().then(() => Promise.all(
            [ request ]
        ).then(data => {
            if (!didCancel) {
            	// post process here if you want
                setRes(data);
            }
        }).catch(e => {
            if (!didCancel) {
                setErrorr(e);
            }
        })).finally(() => {
        	setIsLoadiing(false);
        });

        return () => { didCancel = true; };
    }, []);

    return {
        resp,
        isLoading,
        error
    };;
};

export default useApi;
```

src/Utilities/useApi.js
brumik: Well, the data is passed to the postprocess as we get it in the promise all, so if we know that we are passing an array of urls to the hook then we await an array of objects as a result. So the postprocess accepts an array of objects as an argument and works with them.

brumik:
> A lot of this uses `useReducer` to return an object from `useAPI` hook. I can also see a much more simplified version of this without the `useReducer` hook:
> 
> ```
> import { useState, useEffect } from 'react';
> 
> const useApi = (request) => {
> 	const [ res, setRes ] = useState([]);
> 	const [ isLoadinig, setIsLoading ] = useState(true);
> 	const [ error, setError ] = useState(null);
> 
>     useEffect(() => {
>         // Initialize
>         let didCancel = false;
> 		
>         // Fetching
>         window.insights.chrome.auth.getUser().then(() => Promise.all(
>             [ request ]
>         ).then(data => {
>             if (!didCancel) {
>             	// post process here if you want
>                 setRes(data);
>             }
>         }).catch(e => {
>             if (!didCancel) {
>                 setErrorr(e);
>             }
>         })).finally(() => {
>         	setIsLoadiing(false);
>         });
> 
>         return () => { didCancel = true; };
>     }, []);
> 
>     return {
>         resp,
>         isLoading,
>         error
>     };;
> };
> 
> export default useApi;
> ```

I don't see why is this "simpler", expect that this is rather a helper function than a hook (this solution doesn't expose URL change and probably would operate on calling the function every time when we want to fetch the data).

I favored reducer in this case because you can see much clearer in the reducer which action how affects the different states in one place. The loading, error, success, and data states we look at as basically one object. Reducers are great for these situations in my opinion. This is why I decided to have the `isSuccess` in because it is so easy to see if I made a mistake in the reducer.

src/Utilities/useApi.js
brumik: And it is still done in the parent, you have to define the function in the parent.

src/Containers/JobExplorer/JobExplorer.js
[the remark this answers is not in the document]
kialam: Frankly `urlMappedQueryParams` adds unnecessary obfuscation in its usage and implementation. The snake to camel conversion should happen at the API level (Api.js) rather than being sprinkled everywhere in the code.

src/Utilities/useApi.js
kialam: v1 API returns a `details` object for any errors returned.

src/Containers/JobExplorer/JobExplorer.js
kialam: We shouldn't be using `updateURL` at all.

kialam:
We do not have parameterized URLs in our other views so how would this hook work?

benthomasson (approved):

brumik:
@benthomasson @kialam Should I finish up this hook?
We need to remove keys to the camel function from it and resolve conflicts.

I would delete the usage on the Clusters page since @kialam works on that page anyway. Then @kialam could rebase and use the hook

kialam:
@brumik let's try and merge this PR this week so we can start using it. Nice work 👍

kialam (approved):

PR #323, opened 2020-10-26
ROI change to v1 + little refactor

brumik:
#199 
Solves #131 
Solves #134
Rebased after merging: #314. :heavy_check_mark: 

**Sorry for whoever will look at this. It is not changed THAT much but it is all green since I moved pieces of codes to different files :woman_dancing:**

# Changes
* Filter toolbar! Added endpoint options from ROI Templates
* Change API to v1 - ROI Templates endpoint

# TODO
* [x] If possible remove the mutation of data.
* [x] Add API error handling to the ROI page
* [x] Toolbar selects from the --> everything the ROIOptions endpoint will offer
* [x] Crosslinking date to job explorer: just 30 day + change text to `List of jobs for this template for past 30 days`

PR #364, opened 2020-12-01
Convert Clusters view to V1 API

kialam:
This PR converts the Clusters view to use the V1 `job_explorer` and `event_explorer` endpoints.

Todo:

  - [x] switch to clusters options endpoint
  - [x] fix chart dates to accommodate custom date ranges
  - [x] custom dates for clusters page not hooked up
  - [x] map pertinent filters to top templates, workflows, and modules
  - [x] ~~fix job explorer custom dates~~ this is a toolbar related bug
  - [x] API not returning "name" for group_by: "template" when passing cluster_id
  - [x] pass job_type to details modal
  - [x] update chart click through to include relevant params (org_id, cluster_id, template_id)
  - [x] ensure usage numbers for top templates and top workflows match in the detail modal
  - [x] ~~clearing filters should reset to default params and be page-specific~~ toolbar bug
  - [x] ~~switching from custom back to quick date range should clear out custom & start and end dates~~ again, a toolbar related bug
  - [x] add APIErrorState component to clusters view
  - [x] add most_failed_steps to workflow detail modal
  - [x] API is taking a long time to return aggregate data. Sometimes it does not return `total_failed_count` attribute when passed `most_failed_steps` for workflows. Issue opened here: https://github.com/RedHatInsights/tower-analytics-backend/issues/619. PR here: https://github.com/RedHatInsights/tower-analytics-backend/pull/626
  - [x] fix memory leak from templateslist update function when clicking away from clusters to a different view
  - [x] Ensure "most failed step/task" percentage is correct. Issue opened here: https://github.com/RedHatInsights/tower-analytics-backend/issues/622. PR here: https://github.com/RedHatInsights/tower-analytics-backend/pull/626
  - [x] Update and write new integration tests

Screen Shot 2020-11-30 at 12 34 50 PM


src/Containers/Clusters/Clusters.js
```diff
@@ -23,121 +25,161 @@ import {
...
-        );
+            });
+            fetchEndpoints().then(
+                ([
+                    { items: jobExplorerData = []},
+                    { items: templatesData = []},
+                    { items: workflowData = []},
+                    { items: modulesData = []},
+                    {
+                        cluster_id,
```
brumik, line 118: You can make these lines much more compact, like here:

https://github.com/RedHatInsights/tower-analytics-frontend/blob/ff81a72308f9fd5b5aca5c46c501e2a1ae2fbfa5/src/Containers/OrganizationStatistics/OrganizationStatistics.js#L128

```js
    useEffect(() => {
        ...
        window.insights.chrome.auth.getUser().then(() => Promise.all(
            [
                readJobExplorerOptions({ params: urlMappedQueryParams }),
                ...
            ]
        ).then(([
            options, // Don't destructure, we eventually want to be this object the default settings for the toolbar probably
            ...
        ]) => {
         
            /* eslint-disable-next-line */ // Exlude options which you don't want to display
            const { sortBy, groupBy, attributes, ...rest } = keysToCamel(options);

           // Add extra options, set ONE state
            setOptions({ ...rest, sortBy: constants.sortBy });
            ...
        })

```

src/Containers/Clusters/Clusters.js
```diff
@@ -23,121 +25,161 @@ import {
...
+        setIsLoading(true);
+        new Promise(update).finally(setIsLoading(false));
     }, [ queryParams ]);
 
     return (
         <React.Fragment>
             <PageHeader>
                 <PageHeaderTitle title={ 'Clusters' } />
+                <FilterableToolbar
+                    categories={ {
```
brumik, line 172: Then you can do simething like this with the options:
```js
<FilterableToolbar
    categories={ options }
    filters={ queryParams }
    setFilters={ setFromToolbar }
/>
```

src/Containers/Clusters/Clusters.js
```diff
@@ -23,121 +25,161 @@ import {
...
-                    setClusterOptions(clusterOptions);
-                    setIsLoading(false);
-                })
-            ).catch((error) => {
+        let ignore = false;
+
+        const fetchEndpoints = () => {
+            return Promise.all(
+                [
+                    readJobExplorer({ params: urlMappedQueryParams(queryParams) }),
```
brumik, line 96: Fetching the same data in the `mounting` effect and the `queryParam` effect doesn't fetch it twice when mounted?

Anyway, I think having the same functionality in the code as a duplicate is not too nice. So maybe we can have API calls which we need only load once in the mount and the others in their effect? So when we need to change something on the call we don't need to change it in two places.

brumik (commented):
I came here to see what I need to use for the Org toolbar, but here is a tip: (I know this is a WIP)

src/Containers/Clusters/Clusters.js
kialam: Yes thanks for catching this. I had some logic around `firstRender` that would skip the queryParams useEffect hook entirely and it must have gotten lost during one of my rebases. I've now added it back :).

brumik: If you are taking this approach (fetching in multiple places the same), I am pretty sure you don't need a `state` for `firstRender`, simple `let firstRenderd` is enough, since you don't need to tell react to rerender stuff.

kialam: I have found that using a simple variable declaration and initialization (e.g. `let firstRender = true`) will not work. The variable gets added to the lexical environment and is assigned its initial value each time the component gets rendered. Since we have a hook that updates our component whenever `queryParams` gets updated, `firstRender` needs to be a state variable.

brumik: Oh, sorry for the misleading info. Thanks for the explanation.

Wouldn't it be easier to fetch the `chart30, modules, templates` only in the `queryParam` effect and avoid all `firstRender` logic? I know that `mount` effect takes precedence on the other effects, but we have only an async call, so I don't think it is delaying render by much.

Is there also some performance reason that there is double code for fetching of some endpoints? (chart30, modules, templates in the mount and in the queryParam use effect)

kialam: The queryParam hook handles updating data for the chart, top templates, top workflow, and top modules portion of that page. The hook that fetches on mount gets the initial data for those sections, as well as making additional calls to the `dashboard_clusters_options` and `authorized` endpoint. This optimizes performances by limiting the amount of API calls we make based on various parameters changing to the ones we need.

kialam:
@Ladas should the `total_count` for the "Top Templates" or "Top Workflows" list also match the `total_count` attribute for the top templates/top workflows detail modal?

Eg:
Screen Shot 2021-01-04 at 10 52 52 AM
Screen Shot 2021-01-04 at 10 52 44 AM

The current request payloads are as following:

(Top Templates/Workflows list)
```
{"quick_date_range":"last_30_days","group_by":"template","limit":10,"job_type":["workflowjob"],"group_by_time":false}
```

Detail Modal:
```
{"group_by":"template","template_id":[2],"attributes":["elapsed","job_type","successful_count","failed_count","total_count","most_failed_tasks","most_failed_steps"],"status":["successful","failed"],"quick_date_range":"last_30_days","job_type":["workflowjob"]}
```
Sidenote: most_failed_step percentage seems off. I will look into this.

Ladas:
@kialam yeah the counts should match. I see that the detail modal gets extra status filter? That is probably why the numbers don't match

Ladas:
@kialam can you print here the body for most failed step? It should be normalized to 100% from what I remember (because the test data has the same template in workflow multiple times I think, so the step can fail more times than the workflow)

kialam:
@Ladas yep, here's the response body:

```
failed_count: 100
failed_host_count: 3525900
host_count: 366693600
job_type: "job"
passed_host_count: 356115900
template_id: 2
template_name: "template_name_0"
total_canceled_count: 0
total_error_count: 0
total_failed_count: 10
total_failed_host_count: 20
total_host_count: 2080
total_successful_count: 12
total_unfinished_count: 0
total_unreachable_host_count: 40
unreachable_host_count: 7051800
```

Ladas:
@kialam hm, there is something fishy in there, saying the step ran on 366M hosts (might explain why the query is so slow)

It's saying the step template_name_0 (nested template or workflow) failed 100 times, but it only cause 10 failures of the parent workflow. (so this step failed 10 times in each worflow)

I'll need to look into why the dedup is not working (it should be showing unique template failing the top workflow, so it should say it failed 10 times, if it failed in all workflows)

brumik (changes requested):
Removing the template filter or org filter does as incorrect querry:
Screenshot from 2021-01-22 13-18-40

src/Charts/BarChart.js
```diff
@@ -27,11 +27,14 @@ class BarChart extends Component {
...
         const formattedDate = formatDate(date);
+        const { orgId, templateId } = this.props;
         const initialQueryParams = {
             start_date: formattedDate,
             end_date: formattedDate,
             quick_date_range: 'custom',
-            status: [ 'failed', 'successful' ]
+            status: [ 'failed', 'successful' ],
+            templateId,
+            orgId
```
brumik, line 37: I think the previous problem may be caused by these lines: (not sure) but it seems we should pass:
```js
...
template_id: templateId,
org_id: org_id
```

src/Charts/LineChart.js
```diff
@@ -27,38 +25,34 @@ class LineChart extends Component {
...
+        const { clusterId, templateId, orgId } = this.props;
         const initialQueryParams = {
             start_date: formattedDate,
             end_date: formattedDate,
             quick_date_range: 'custom',
             status: [ 'failed', 'successful' ],
-            cluster_id: [ clusterId ]
+            clusterId,
+            orgId,
+            templateId
```
brumik, line 36: If the previous one is causing problems, probably this one will too

src/Charts/LineChart.js
```diff
@@ -208,10 +206,10 @@ class LineChart extends Component {
...
             'transform',
             'translate(' +
-                width / 2 +
-                ' ,' +
-                (height + this.props.margin.top + 20) +
-                ')'
+          width / 2 +
+          ' ,' +
+          (height + this.props.margin.top + 20) +
+          ')'
```
brumik, line 212: Indent

src/Charts/LineChart.js
```diff
@@ -332,13 +329,14 @@ class LineChart extends Component {
...
+    templateId: PropTypes.array,
+    orgId: PropTypes.array,
     data: PropTypes.array,
-    value: PropTypes.number,
     margin: PropTypes.object,
     getHeight: PropTypes.func,
     getWidth: PropTypes.func,
-    history: PropTypes.object,
-    clusterId: PropTypes.number
+    history: PropTypes.object
```
brumik, line 339: **Just a note**: You can pull history from the `react-router` as an import and remove `withRouter` import and it would make the charts testing easier (for one) and then there is no need to pass history.

(I remember doing this some time ago)

src/Components/TemplatesList.js
```diff
@@ -98,36 +100,96 @@ const formatTopFailedTask = data => {
...
+                    setRelatedJobs(relatedJobs);
+                }
             });
         };
 
         if (selectedId) {
             update();
         }
 
+        return () => (ignore = true);
```
brumik, line 192: Can we use `didCancel` here too for consistency?

src/Containers/Clusters/Clusters.js
```diff
@@ -23,208 +22,226 @@ import {
...
+                        setJobTypes(job_type);
+                        setQuickDateRanges(quick_date_range);
+                    }
+                })
+            .catch(e => setApiError(e.error));
+        }
+
+        initializeWithPreflight();
+
+        return () => (ignore = true);
```
brumik, line 145: Oh sorry, I thought you changed these to `didCancel` that's why the comment ^

src/Containers/Clusters/Clusters.js
```diff
@@ -23,208 +22,226 @@ import {
...
+        let ignore = false;
+
+        async function initializeWithPreflight() {
+            await window.insights.chrome.auth.getUser();
+            await preflightRequest().catch(error => {
                 setPreFlightError({ preflightError: error });
-            })
-        );
+            });
+            readClustersOptions({ params: optionsQueryParams }).then(
```
brumik, line 122: I was talking with @Ladas and I think we will need to refetch these too on queryParam change since in the future options like `granularity` will change the `quick date ranges` and possibly other options too.

So I think it would be futureproofing if we would do the refetching of the options too when changing values instead of fetching once on mount.

@Ladas Correct me if I am wrong.

src/Containers/Clusters/Clusters.js
```diff
@@ -23,208 +22,226 @@ import {
...
                 </Main>
             ) }
-            { !preflightError && (
+            { apiError && (
+                <Main>
+                    <ApiErrorState message={ apiError } />
+                </Main>
+            ) }
+            { !preflightError && !apiError && (
         <>
```
brumik, line 218: Can we use `React.Fragment` instead of the sorthand. The shorthand messes up the Eslint indent rules.

brumik (changes requested):
Its a nice work, I had mostly nitpicking and code format in the review.

Howver there is one issue (which may be connected to the API - I am using up to date devel backend) where a top module has no name. Maybe we can write `No module name` or something when there is no name?

Screenshot from 2021-01-22 13-44-40

@Ladas

src/Containers/Clusters/Clusters.js
Ladas: @brumik yeah, but we can do this in separate PR. Also we should get together with @AparnaKarve and see how to enable `show only eligible options`, where filtering always produces some results. 

So e.g. from start you can filter 10 templates and 10 orgs, you'll filter 2 orgs and now you can see only 5 templates (since only 5 templates are being used by those 2 orgs).

We'll need to design together how this should work.

Ladas:
> Its a nice work, I had mostly nitpicking and code format in the review.
> 
> Howver there is one issue (which may be connected to the API - I am using up to date devel backend) where a top module has no name. Maybe we can write `No module name` or something when there is no name?
> 
> Screenshot from 2021-01-22 13-44-40
> 
> @Ladas

@kialam @brumik the empty module name shouldn't appear in prod data. But later I'd be just for some generic empty name decorator we'd use everywhere?

src/Containers/Clusters/Clusters.js
kialam: Until there is an API implementation of this, I will not be changing the UI to include extra calls.

Ladas: yeah, it can wait until it's needed

kialam:
@brumik after using the `useAPI` hook for `TemplatesList`, I am seeing this error in the console when the modal gets closed:

Screen Shot 2021-02-02 at 5 46 27 PM

brumik:
@kialam Can you show me? It is really vague like this (you didn't push the non-working version it seems)

src/Components/TemplatesList.js
```diff
@@ -201,119 +247,128 @@ const TemplatesList = ({ history, clusterId, templates, isLoading, queryParams }
...
-          </Modal>
-      ) }
+            <Modal
+                aria-label="modal"
+                width={ '80%' }
+                title={ stats.name ? stats.name : 'no-template-name' }
+                isOpen={ isModalOpen }
+                onClose={ () => {
+                    setIsModalOpen(false);
+                    setRelatedJobs([]);
```
kialam, line 262: @brumik using the returned `setState` from `useAPI` here fails with a `request.then is not found` error message. The hook expects the `setState` method to be called with a request, however I am using it to reset to the initial state of `[]`.

src/Components/TemplatesList.js
```diff
@@ -201,119 +247,128 @@ const TemplatesList = ({ history, clusterId, templates, isLoading, queryParams }
...
+                    setSelectedId(null);
+                } }
+                actions={ [
+                    <Button
+                        key="cancel"
+                        variant="secondary"
+                        onClick={ () => {
+                            setIsModalOpen(false);
+                            setRelatedJobs([]);
```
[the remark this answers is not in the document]
brumik: @kialam Oh yes, you cannot do that now. The `setRequest` expects a `Promise` returned. Right now you cannot reset the state. However, if you remember yesterday's conversation, this is solved here: https://github.com/RedHatInsights/tower-analytics-frontend/pull/323/commits/01b7b6f4928aa3619e823bf1324cd79ab1d4ddbe#diff-31bc939331c6e5b934a01d1d52249903bd424c43e01a8aa64eff3613715ca286

Then you can use the `setData` function instead of `setRequest` from the `useApi` hook. I had a different use case for this change but this is a perfectly valid one.

src/Charts/BarChart.js
```diff
@@ -57,12 +60,12 @@ class BarChart extends Component {
...
         d3.selectAll('#' + this.props.id + ' > *').remove();
         const parseTime = d3.timeParse('%Y-%m-%d');
-        let { data: unformattedData, value } = this.props;
-        const data = unformattedData.reduce((formatted, { created, successful, failed }) => {
-            let DATE = parseTime(created) || new Date();
-            let RAN = +successful || 0;
-            let FAIL = +failed || 0;
-            let TOTAL = +successful + failed || 0;
+        let { data: unformattedData } = this.props;
+        const data = unformattedData.reduce((formatted, { created_date, successful_count, failed_count }) => {
```
brumik, line 64: I know this is a "legacy" code but these two lines could be simplified as:
```js
const data = this.props.data.map(({ created_date, successful_count, failed_count, ...rest }) => (
    ...rest,
    DATE: parseTime(created_date) || new Date(),
    RAN: +successful_count || 0,
    FAIL: +failed_count || 0,
    TOTAL: +successful_count + failed_count || 0
));

In the current code there are too much variables and most of them is not even `constant`
```

src/Charts/LineChart.js
```diff
@@ -112,16 +106,19 @@ class LineChart extends Component {
             svg: '#' + this.props.id,
             colors
         });
-        const { data: unformattedData, value } = this.props;
+        const { data: unformattedData } = this.props;
         const parseTime = d3.timeParse('%Y-%m-%d');
 
-        const data = unformattedData.reduce((formatted, { created, successful, failed }) => {
```
brumik, line 118: The previous comment could be applied here too

src/Components/TemplatesList.js
```diff
@@ -98,36 +101,82 @@ const formatTopFailedTask = data => {
...
+    };
 
+    useEffect(() => {
         if (selectedId) {
-            update();
+            setStats(readJobExplorer({ params: agreggateTemplateParams }));
+            setRelatedJobs(readJobExplorer({ params: relatedTemplateJobsParams }));
         }
-
     }, [ selectedId ]);
```
brumik, line 180: @kialam I was thinking about this: could we fetch the data in the modal where it is used?

So what I am thinking:
* On button/link click -> open modal (`setIsModalOpen(true)`)
* Modal would have a React Component `X`  as a child with the `selectedID` param
* The `X` would on mount parse the data from the API with its own loading screen, API error, etc...

This hierarchy would make the `TemplateList` component smaller and would move the parsing logic where it is used for the modal component. Also, you would not need to reset the modal component data on close because on open it would display the `loading` screen not the old data. 

I hope I could explain it well, if not ping me

src/Charts/LineChart.js
kialam: https://reactrouter.com/web/api/Hooks/usehistory :)

src/Containers/Clusters/Clusters.js
kialam: I think this is an issue with our linter.

src/Components/ModalContents.js
```diff
@@ -0,0 +1,313 @@
...
+
+const success = (
+    <CircleIcon
+        size="sm"
+        key="5"
+        style={ { color: 'rgb(110, 198, 100)', marginRight: '5px' } }
+    />
+);
+const fail = (
+    <>
```
brumik, line 33: Can we use <React.Fragment> to help eslint?

src/Components/ModalContents.js
```diff
@@ -0,0 +1,313 @@
...
+                                    <PFDataListCell key="start time">
+                                        { formatDateTime(job.started) }
+                                    </PFDataListCell>
+                                    <PFDataListCell key="total time">
+                                        { formatSeconds(job.elapsed) }
+                                    </PFDataListCell>
+                                </DataListItem>
+                            )) }
+            </DataList>
+            { /* <ModalContents selectedId={ selectedId } qp={ qp } jobType={ jobType } /> */ }
```
brumik, line 299: Forgot to delete?

src/Components/Toolbar/ToolbarFilterItem.js
```diff
@@ -56,6 +56,7 @@ const ToolbarFilterItem = ({
 
     return (
         <ToolbarFilter
+            data-cy = { categoryKey }
```
brumik, line 59: Nice ^^

src/Containers/Clusters/Clusters.js
```diff
@@ -23,208 +23,184 @@ import {
...
-                { modules: modulesData = []},
-                { templates: templatesData = []}
-            ]) => {
-                queryParams.id ? setLineChartData(chartData) : setBarChartData(chartData);
-                setModulesData(modulesData);
-                setTemplatesData(templatesData);
-                setIsLoading(false);
-            }).catch(() => [])
-        );
+        const fetchEndpoints = () => {
```
brumik, line 146: No need for extra function here

src/Containers/Clusters/Clusters.js
```diff
@@ -23,208 +23,184 @@ import {
...
                 </Main>
             ) }
-            { !preflightError && (
+            { error && (
+                <Main>
+                    <ApiErrorState message={ error.error } />
+                </Main>
+            ) }
+            { !preflightError && !error && (
         <>
```
brumik, line 177: This is why I try to use `React.Fragment` instead of the shorthand, eslint doesn't detect indent mismatch

brumik (approved):
Great work, it is really extensive, I had a few nitpicks, but nothing that is blocking this PR to get merged. If you want you can change them, if not, this is good to be merged.

Additionally discovered another issue with the Toolbar's `Clear All Filter` but I believe that issue is not specific to this PR and I will fix it in a different PR.

Issue #333, opened 2020-11-02
Legend in charts has wrong size and the has the % symbol where it is not a % for sure


brumik:
The legend now has % --> this is misleading and most probably wrong see:

![Screenshot from 2020-11-02 14-20-11](https://user-images.githubusercontent.com/8531681/97872699-c36b3080-1d16-11eb-9a71-8b620c3665c8.png)

Also, the legend font sizes are different for the label and the value: 12px for the title 15 for the text. I believe this was introduced in UXD audit changes: https://github.com/RedHatInsights/tower-analytics-frontend/commit/192ed854ca893dd7d201406ea0cd61875505b244

@jeanetteobr

benthomasson:
@jeanetteobr Did the % come from the UXD feedback?  I would prefer to keep the raw counts instead of percentages on the legend.

Ladas:
yeah just noticed this, this needs to be recomputed if we are showing % (afaik the UXD feedback was showing both percentage and absolute number?)

Ladas:
@kialam @jeanetteobr  hm, my comment about showing also the absolute number is gone now https://docs.google.com/presentation/d/1FTNmLVUwbXWX1NxcN31yvtzYBrVDlRNioQN1Cehfx7k/edit#slide=id.g9376ae2f39_0_46

I believe we were talking about that and it was similar to what Cost team is doing (also formulation of the others)

Ladas:
I believe we were talking about something like this  (just pie chart)

![Screenshot from 2020-11-04 14-26-43](https://user-images.githubusercontent.com/1737058/98117368-ee39be00-1ea9-11eb-8a12-b36554abfc8b.png)

Without the absolute numbers, the percentages are kind of without meaning
----
screenshot from cost mgmt demo https://drive.google.com/file/d/1R8Hm_cRYYhxjObVNK2_rub8TY-BGUtSu/view

Ladas:
@kialam I think we agreed that the `X others` name would be added in the API? If that is the case I'll create a backend issue for it :-)

jeanetteobr:
@benthomasson The percentages were part of the UXD Audit. I admit that I may have misinterpreted what the issue related to the change was asking for. I'll have a PR up that fixes this sometime today.

jeanetteobr:
Here's the PR that addresses this issue: https://github.com/RedHatInsights/tower-analytics-frontend/pull/371

PR #371, opened 2020-12-09
Jobs Run by Org Chart Bugfix

jeanetteobr:
This PR is a bugfix for issue #333.

- Per the request of @wenottingham, we will only show raw count on the job runs by org pie chart.
- The chart labels will now be a consistent font size of 15px for better legibility. 

Before:
image

Verification:
Screen Shot 2020-12-09 at 11 07 39 AM


src/Utilities/Legend.js
```diff
@@ -35,7 +35,7 @@ const Color = styled.div.attrs(props => ({
 `;
 
 const Title = styled.span`
-  font-size: 12px;
+  font-size: 15px;
```
brumik, line 38: This makes the legend bigger than the title... are we sure we want this? I tried it out on a HDready screen and it is huge. If you say it's needed, then okey, just feels strange for me. I personally would remove the custom font size and leave the PF sizes. This is not an issue, just my opinion...

@trahman73 @wenottingham

brumik (dismissed):

src/Utilities/Legend.js
jeanetteobr: @brumik It looks like the card titles are 16px and the legend labels are 15px as you've pointed out above. I can adjust the size of the labels however, I'm not sure what the size is intended to be. I only had the verification image to cross-reference when I was working on this audit change originally.

src/Components/JobExplorerHostChart.js
```diff
@@ -0,0 +1,48 @@
...
+    console.log('data', data);
+
+    if (data.most_failed_tasks === null) {
+        return null;
+    }
+
+    return 'Success!'
+};
+
+export default JobExplorerHostChart;
```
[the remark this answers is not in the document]
jeanetteobr: @brumik Thanks for catching that. Not sure how that file found its way into this commit.

benthomasson (approved):

PR #372, opened 2020-12-10
Push to all the betas from devel

benthomasson:
This change will push to all beta environments from devel.     This supports a faster beta feedback cycle but makes production-beta more unstable.   Users can always go to the non-beta production site if they want a stable version.

Please discuss pushing to all betas at the same time from devel.   Point out any problems like the left nav needing matching PRs.

### DIFF
--- a/.travis/custom_release.sh
+++ b/.travis/custom_release.sh
@@ -2,12 +2,14 @@
 set -e
 set -x
 
-# If current dev branch is devel, push to build repo ci-beta
+# If current dev branch is devel, push to build repo ci-beta, qa-beta, and prod-beta
 if [ "${TRAVIS_BRANCH}" = "devel" ]; then
     .travis/release.sh "ci-beta"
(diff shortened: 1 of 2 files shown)


jctanner (approved):
tests are good now ... although not really affected by this change =)

kialam (approved):
👍 from me. I can't think of a reason not to do this. Navigation changes will be required for net-new features however maybe we can talk to @ryelo and @kruai if there's a way to avoid opening 3 PRs at once to add nav items to all the `beta` environments.

kruai:
@kialam For CSC PRs, you should just need to make one to ci-beta, followed by a PR to prod-beta, once you've validated.

jeanetteobr (approved):

brumik (approved):

brumik:
I think this change should update the README too (there is also a mismatch with going `master` -> `devel`)

PR #373, opened 2020-12-11
Tests: Updated the App tests to use the global insights object

brumik:
Removed the `App.test` mock of `insights`. We have it globaly set up. Let this test to use it as well.

### DIFF
--- a/config/setupTests.js
+++ b/config/setupTests.js
@@ -12,8 +12,11 @@ global.React = React;
 // For page API
 global.insights = {
     chrome: {
+        on() {},
+        init() {},
+        identifyApp() {},
         auth: {
-            getUser: () => new Promise((resolve) => resolve(''))
+            getUser: () => new Promise((resolve) => resolve('bob'))
         },
         appNavClick: jest.fn()
     }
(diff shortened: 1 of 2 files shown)


config/setupTests.js
```diff
@@ -12,8 +12,11 @@ global.React = React;
 // For page API
 global.insights = {
     chrome: {
+        on() {},
+        init() {},
+        identifyApp() {},
         auth: {
-            getUser: () => new Promise((resolve) => resolve(''))
+            getUser: () => new Promise((resolve) => resolve('bob'))
```
jctanner, line 19: Per https://github.com/RedHatInsights/tower-analytics-frontend/commit/066905b6f1f75f11ad5016179b341e156c0083bd it seems like this returned bob before the refactor, so I'm guessing this is okay.

src/App.test.js
```diff
@@ -1,36 +1,14 @@
-
-import { mount, shallow } from 'enzyme';
 import App from './App.js';
 import { Router } from 'react-router-dom';
-
-import configureStore from 'redux-mock-store';
-import { Provider } from 'react-redux';
-
+import { mountPage } from './Containers/tests/helpers';
```
jctanner, line 3: :+1:

jctanner (approved):

config/setupTests.js
brumik: I liked bob, so I put it back ^^

PR #375, opened 2020-12-14
useFetch hook POC

kialam:
Borrows AWX's [useRequest](https://github.com/ansible/awx/blob/devel/awx/ui_next/src/util/useRequest.js) hook to create an AA useFetch hook.

Usage:

```
const {
        result: { jobExplorerData, meta },
        error: jobsError,
        isLoading: isJobExplorerLoading,
        request: fetchJobExplorer
    } = useFetch(
        useCallback(async params => {
            const [ jobs ] = await Promise.all([ readJobExplorer({ params }) ]);
            return {
                jobExplorerData: jobs.items,
                meta: jobs.meta
            };
        }, []),
        {
            jobExplorerData: [],
            meta: {}
        }
    );
```
- destructure & and reassign the object that this hook returns:
```
{
   result::any,
   error::any,
   isLoading::boolean,
   request::function
}
```

- pass function as first argument to hook. This function is a [callback hook](https://reactjs.org/docs/hooks-reference.html#usecallback) that is referenced as `makeRequest` and is expected to return an object.

```
useCallback(async params => {
    const [ jobs ] = await Promise.all([ readJobExplorer({ params }) ]);
    return {
        jobExplorerData: jobs.items,
        meta: jobs.meta
    };
}, []),
```

- second argument is the initial default values of your result(s)

```
{
    jobExplorerData: [],
    meta: {}
}
```

Lastly use it like you would within a regular `useEffect` hook:

```
useEffect(() => {
    fetchJobExplorer(urlMappedQueryParams);
}, [ queryParams ]);
```

Question:
1. Should `preFlight request` be called before each API request or just once when component mounts?

### DIFF
--- a/src/Containers/JobExplorer/JobExplorer.js
+++ b/src/Containers/JobExplorer/JobExplorer.js
@@ -1,17 +1,19 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import PropTypes from 'prop-types';
-import { parse, stringify } from 'query-string';
+import {
+    parse
+    // stringify
+} from 'query-string';
 
 import { useQueryParams } from '../../Utilities/useQueryParams';
+import useFetch from '../../Utilities/useFetch';
 import { keysToCamel } from '../../Utilities/helpers';
-import { Paths } from '../../paths';
 
 import LoadingState from '../../Components/LoadingState';
-import EmptyState from '../../Components/EmptyState';
 import NoResults from '../../Components/NoResults';
 import ApiErrorState from '../../Components/ApiErrorState';
 import {
-    preflightRequest,
+    // preflightRequest,
     readJobExplorer,
     readJobExplorerOptions
 } from '../../Api';
@@ -47,16 +49,11 @@ const initialQueryParams = {
 };
 
 const JobExplorer = ({
-    location: { search },
-    history
+    location: { search }
+    // history
 }) => {
-    const [ preflightError, setPreFlightError ] = useState(null);
-    const [ apiError, setApiError ] = useState(null);
-    const [ isLoading, setIsLoading ] = useState(true);
-    const [ jobExplorerData, setJobExplorerData ] = useState([]);
-    const [ meta, setMeta ] = useState({});
+    // const [ preflightError, setPreFlightError ] = useState(null);
     const [ currPage, setCurrPage ] = useState(1);
-    const [ explorerOptions, setExplorerOptions ] = useState({});
 
     let initialSearchParams = keysToCamel(
         parse(search, { arrayFormat: 'bracket' })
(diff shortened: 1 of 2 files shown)


kialam:
Related #376

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -70,51 +67,56 @@ const JobExplorer = ({
         setFromToolbar
     } = useQueryParams(combined);
 
-    const updateURL = () => {
```
brumik, line 73: We don't need this anymore?

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -47,16 +49,11 @@ const initialQueryParams = {
...
-    history
+    location: { search }
+    // history
 }) => {
-    const [ preflightError, setPreFlightError ] = useState(null);
-    const [ apiError, setApiError ] = useState(null);
-    const [ isLoading, setIsLoading ] = useState(true);
-    const [ jobExplorerData, setJobExplorerData ] = useState([]);
-    const [ meta, setMeta ] = useState({});
+    // const [ preflightError, setPreFlightError ] = useState(null);
```
brumik, line 55: We don't need preflight check anymore? I don't see it moved and I am a bit puzzled

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -133,68 +135,71 @@ const JobExplorer = ({
...
+                      <LoadingState />
+                  ) }
+                  { !jobsError &&
+              !isJobExplorerLoading &&
+              !isJobExplorerOptionsLoading &&
+              jobExplorerData.length <= 0 && <NoResults /> }
+                  { !jobsError &&
+              !isJobExplorerLoading &&
+              !isJobExplorerOptionsLoading &&
+              jobExplorerData.length > 0 && (
```
brumik, line 181: Indent mismatch.

Also, I don't see why we need to wait **here** for job explorer options to load? We could indicate it in the toolbar or if we need the data and the options to synch maybe we can use Promise.all to fetch them.

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -70,51 +67,56 @@ const JobExplorer = ({
...
-                setExplorerOptions(rest);
-            })
-            .catch(e => setApiError(e.error))
-            .finally(() => {
-                updateURL();
-                setIsLoading(false);
-            });
-        });
+        fetchJobExplorerOptions();
+    }, [ fetchJobExplorerOptions ]);
```
brumik, line 116: Maybe this is just for showcasing the options of usage?

I guess this is needed in case of the `useCallback` reference changes when the `urlMappedQueryParams` change. However then why we cannot do it for the `fetchJobExplorer` too?

Personally, I think this is a bit less readable than having them both be called in the `useEffect` with `queryParams`.

brumik (changes requested):
This is very similar in the use case to the other PR. It is utilizing a lot of memoization, which is nice since it allows to pass initial URL when calling the hook without firing the hook multiple times.

benthomasson:
@keithjgrant wrote useRequest.

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -70,51 +67,56 @@ const JobExplorer = ({
...
-            search
-        });
-    };
+    const {
+        result: { jobExplorerData, meta },
+        error: jobsError,
+        isLoading: isJobExplorerLoading,
+        request: fetchJobExplorer
+    } = useFetch(
+        useCallback(async params => {
```
benthomasson, line 76: It seems like useCallback is always used with useFetch.  Can it be added to useFetch so we don't need the boilerplate code?

src/Containers/JobExplorer/JobExplorer.js
brumik: I am not sure how indicating loading and parallel data loading is connected?

src/Containers/JobExplorer/JobExplorer.js
kialam: The job explorer options should only be fetched once when the toolbar component renders.

src/Containers/JobExplorer/JobExplorer.js
kialam: I think we can remove this since bookmarking is not yet implemented. Leftover code.

src/Containers/JobExplorer/JobExplorer.js
kialam: I commented it out because I am not sure the order/prioritization of this call within the new hook. If it can only be called once on initial render then we can pass it like we do to the hook, however if it needs to be called for each call then it needs to be incorporated in the hook itself. @benthomasson thoughts on this?

src/Containers/JobExplorer/JobExplorer.js
kialam: So, this conditional ensures that there are no errors returned from the API before we render the jobs list. @brumik yeah any errors from the options endpoint should be moved to the filter toolbar block.

src/Containers/JobExplorer/JobExplorer.js
kialam: @benthomasson I am not sure I understand what you mean here...

keithjgrant: We tried to find a way to push `useCallback` down in AWX and ultimately decided against it. Primarily, it breaks linting, since the React hooks lint tools need to see the dependency array defined inline where `useCallback` is invoked (i.e. you can't pass the whole dependency array in as a parameter without linting errors). It would also basically go against the purpose of useCallback, which is to make a callback hooks-safe at the point where you define it.

src/Containers/JobExplorer/JobExplorer.js
brumik: Hm, okay, when we were talking about before I got the feeling that the options endpoint can return different values depending on the current options too (like changing in the future granularity would change date selection options).

src/Containers/JobExplorer/JobExplorer.js
brumik: Actually, bookmarking is working for JobExplorer right now with this code

brumik:
Closing in favour of https://github.com/RedHatInsights/tower-analytics-frontend/pull/359
@kialam delete the branch if you you also think we won't need this anymore

PR #379, opened 2020-12-15
Disable camel case check

benthomasson:
Since we have a convention of using snake case to match the attributes from the API turn off camel case checking.


kialam (approved):
👌

PR #393, opened 2021-01-18
Broken cypress tests into smaller per page files

brumik:


brumik:
@kialam Look at this and try to run the tests, I removed from notification and orgs the waits, if it runs with you locally and on github I remove the other waits as well, as it is an antipattern.

package.json
```diff
@@ -90,8 +90,8 @@
   "scripts": {
     "build": "webpack --config config/prod.webpack.config.js",
     "test": "jest --verbose",
-    "integration_headless": "cypress run --headless --browser chrome --spec cypress/integration/automation-analytics.js",
-    "integration": "cypress run --no-exit --browser chrome --spec cypress/integration/automation-analytics.js",
+    "integration_headless": "cypress run --headless --browser chrome --spec 'cypress/integration/*.spec.js'",
```
kialam, line 93: No need to pass the `--spec` flag. The cypress runner will know to look in the `/integration` directory. `cypress run --headless --browser chrome` should work.

cypress.json
```diff
@@ -1,5 +1,11 @@
 {
     "chromeWebSecurity": false,
     "waitForAnimations": true,
-    "videoCompression": 0
+    "videoCompression": 0,
+    "env": {
```
kialam, line 5: Leave out CLOUD_BASE_URL, CLOUD_USERNAME, and CLOUD_PASSWORD as it can differ from machine to machine or based on what you want to point cypress at.

cypress/integration/AutomationCalculator.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    cy.viewport(1600, 2000);
+    cy.getBaseUrl().then(url => cy.visit(url));
+
+    // sso login ...
+    cy.get('[data-ouia-component-id="1"]').click();
+    cy.getUsername().then(uname => cy.get('#username').type(`${uname}{enter}`));
+    cy.getPassword().then(password => cy.get('#password').type(`${password}{enter}`));
+});
+
+describe('Notification page smoketests', () => {
```
kialam, line 40: Automation Calculator not Notification

kialam (changes requested):
I can't seem to get the other test files to run although it looks like AutomationCalculator.spec.js runs fine. I would test these changes against prod before moving forward with removing the wait calls.

cypress.json
brumik: You can override them if you define them in the terminal:
https://docs.cypress.io/guides/guides/environment-variables.html#Setting
So this is the same as was before, a default value if you don't define them.

brumik:
@kialam I put the wait back, now we need to figure out where the package.json config is coming to the github action (docker image).

kialam (approved):

PR #394, opened 2021-01-22
Cypress testing: Changed env variable for baseURL

brumik:
For these changes we need to change the pipeline here: https://github.com/RedHatInsights/analytics-pipeline

## Env variable renames
Renamed:
* `CLOUD_PASSWORD --> PASSWORD` 
* `CLOUD_USERNAME --> USERNAME` 
* `CLOUD_BASE_URL --> BASE_URL`
 
Also, the base URL is now set in the `cypress.json` making running server checks available.  The `baseURL` is a standard convention in cypress and it is best practice to have it set up. Allows using visit, without specifying the base URL part all the time.

So instead of: `cy.visit('https://localhost:8000/index')` we can use just `cy.visit('/index')`.

## Additional global settings:
Added `viewport` to the config file. This sets the default viewport for all tests. If it is needed it can be overwritten with `cy.viewport(x, y)` from withing the tests.

## Removed waits:
In cypress, waits are [anti-pattern](https://docs.cypress.io/guides/references/best-practices.html#Unnecessary-Waiting). If needed we can increase timeouts for various events in the config:
https://docs.cypress.io/guides/references/configuration.html#Timeouts

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -2,10 +2,12 @@
     "chromeWebSecurity": false,
     "waitForAnimations": true,
     "videoCompression": 0,
+    "viewportHeight": 2000,
+    "viewportWidth": 1600,
+    "baseUrl": "https://prod.foo.redhat.com:8443",
     "env": {
         "appid":"#automation-analytics-application",
-        "CLOUD_BASE_URL": "https://prod.foo.redhat.com:8443",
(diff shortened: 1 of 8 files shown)


kialam (approved):
👏  Nice - this is much cleaner!

PR #395, opened 2021-01-25
'View All' Bugfix on Notifications page

jeanetteobr:
### Verification
This PR addresses a bug in prod on the Notifications page that causes an empty query string to be sent to the Notifications endpoint on the selection of the 'View All' filter. 

**Before:**
DMG1fsE3SY

**After:**
GFE0niCKt2

Jira Issue: https://issues.redhat.com/browse/AA-238

### DIFF
--- a/src/Utilities/useQueryParams.js
+++ b/src/Utilities/useQueryParams.js
@@ -28,12 +28,13 @@ export const useQueryParams = initial => {
                 return { ...state, limit: parseInt(value.limit) };
             /* v1 api reducers */
             case 'SET_OFFSET':
-            case 'SET_SEVERITY': {
-                const { severity: ignored, ...rest } = state;
-                if (value.severity === '') { return rest; }
-            }
+            case 'SET_SEVERITY':
+                if (value.severity === '') {
+                    const { severity: ignored, ...rest } = state;
+                    return rest;
+                }
 
-            // eslint-disable-next-line no-fallthrough
+                return { ...state, ...value };
             case 'SET_ATTRIBUTES':
             case 'SET_JOB_TYPE':
             case 'SET_STATUS':
--- a/src/Utilities/useQueryParams.js
+++ b/src/Utilities/useQueryParams.js
@@ -27,13 +27,6 @@ export const useQueryParams = initial => {
                 return { ...state, limit: parseInt(value.limit) };
             /* v1 api reducers */
             case 'SET_OFFSET':
-            case 'SET_SEVERITY':
-                if (value.severity === '') {
-                    const { severity: ignored, ...rest } = state;
-                    return rest;
-                }
-
-                return { ...state, ...value };
             case 'SET_ATTRIBUTES':
             case 'SET_JOB_TYPE':
             case 'SET_STATUS':
@@ -53,6 +46,13 @@ export const useQueryParams = initial => {
                 return newState;
             }
 
+            case 'SET_SEVERITY':
+                if (value.severity === '') {
+                    const { severity: ignored, ...rest } = state;
+                    return rest;
+                }
+
+                return { ...state, ...value };
             case 'SET_START_DATE':
             case 'SET_END_DATE': {
                 let newValues = {};
--- a/src/Utilities/useQueryParams.js
+++ b/src/Utilities/useQueryParams.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import { useReducer } from 'react';
 import moment from 'moment';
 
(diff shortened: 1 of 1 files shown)


src/Utilities/useQueryParams.js
```diff
@@ -28,6 +29,12 @@ export const useQueryParams = initial => {
             /* v1 api reducers */
             case 'SET_OFFSET':
             case 'SET_SEVERITY':
+                if (value.severity === '') {
```
brumik, line 32: You cannot do this, this way you break the `SET_OFFSET` rule here.

src/Utilities/useQueryParams.js
```diff
@@ -28,6 +29,12 @@ export const useQueryParams = initial => {
...
+                    const { severity: ignored, ...rest } = state;
```
brumik, line 33: We don't need `ignored` anymore, there was a rule added in eslint for destructuring + unused

src/Utilities/useQueryParams.js
```diff
@@ -94,7 +101,9 @@ export const useQueryParams = initial => {
         dispatch,
         setLimit: limit => dispatch({ type: 'SET_LIMIT', value: { limit }}),
         setOffset: offset => dispatch({ type: 'SET_OFFSET', value: { offset }}),
-        setSeverity: severity => dispatch({ type: 'SET_SEVERITY', value: { severity }}),
+        setSeverity: (severity) => {
```
brumik, line 104: I think here no functionality changed except not returning the dispatch return value?

src/Utilities/useQueryParams.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
brumik, line 1: Forgot eslint rule

brumik (changes requested):

brumik:
@kialam I remember when we could break this: when we were converting maybe from `urlMappedQueryParams` to `queryParams` only. Since in the `urlMappedQuerryParams` I was removing the empty params :thinking:

src/Utilities/useQueryParams.js
jeanetteobr: Could you please elaborate on how the logic on line 32 would break the `SET_OFFSET` rule?

src/Utilities/useQueryParams.js
jeanetteobr: We are doing something similar on line 21 with `SET_LIMIT`...

src/Utilities/useQueryParams.js
jeanetteobr: I was console logging some values here and neglected to revert the function to its original state...

src/Utilities/useQueryParams.js
brumik: Yes, it was needed before, but we changed the eslint rules a week or two back, so now we don't need it

src/Utilities/useQueryParams.js
brumik: Hm, hard to explain for me so sorry if it baroquesq:

When you write a list of cases then it will jump off the only on break or return from the switch block. So in this example right now from `SET_OFFSET` to `SET_ROOT_WORKFLOWS_AND_JOBS` were using the same line of code: `return { ...state, ...value };` since there were no other `return` or `break` statements.

Now you modified it sonow the `SET_OFFSET` which is before `SET_SEVERITY` uses the logic what you wrote for `SET_SEVERITY`. 

So basically if there is no break or return after a case the next case code is also run. 

So for example you can (and maybe should) delete your `return { ...state, ...value };` line and just leave the `if` block, since you won't have any return so the last `return { ...state, ...value };` would run.

brumik: So, yeah, I didn't tested the code and it may work, but it is not safe to leave it like this I think. My solution would be to move the `SET_SEVERITY` case block under `SET_ROOT_WORKLOFS_AND_JOBS` code block.

brumik: And sorry that I didn't explain it before, it was a mistake I made in a hurry.

src/Utilities/useQueryParams.js
brumik: It's not actually is a problem, I just don't like extra character typing :D but you can leave it

brumik (approved):
I approve, the bug is gone!

brumik:
I guess we need to test the `queryParam` hook better in jest, if you want to try you can look for my `useApi.test.js` where I test a hook. It's pretty simple :) But that is outside of the scope of PR so this is good to go

PR #397, opened 2021-01-27
Error notification/ filtered notification item timestamps

jeanetteobr:
This PR fixes the following bugs: 

- Timestamp for Error notifications appear encoded and are not human-readable (AA-239: https://issues.redhat.com/browse/AA-239)
- Timestamp not displayed for filtered notification items (AA-246: https://issues.redhat.com/browse/AA-246)
- Some stringified timestamps show negative integer 

### Verifcation

**Before:**
TfpS5AppR7

**After:**
m0ZVPz5Alt

### DIFF
--- a/src/Components/NotificationsList.js
+++ b/src/Components/NotificationsList.js
@@ -13,7 +13,6 @@ import {
 import { ExternalLinkAltIcon as PFExternalLinkAltIcon } from '@patternfly/react-icons';
 import LoadingState from '../Components/LoadingState';
 import { capitalize } from '../Utilities/helpers';
-import { stringify } from 'query-string';
 
 const ExternalLinkAltIcon = styled(PFExternalLinkAltIcon)`
   margin-left: 7px;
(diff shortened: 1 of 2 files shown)


src/Components/NotificationsList.js
```diff
@@ -40,19 +39,20 @@ const NotificationDrawerList = styled(PFNotificationDrawerList)`
 `;
 
 const stringifyDate = (date) => {
+    const day = moment(date);
```
brumik, line 42: :+1:

brumik (approved):
The bug was caused by the missing timestamp param?

Eg.: the modifications in the `stringifyDate` function were not necessary? (I like what you done with it just interested)

I could not break it anymore so it looks good to me :)

brumik:
@jeanetteobr You probably accidentally checked out from your other branch, but that one is good to go too so it should not be a problem.

We should merge that one first (#395)

src/Components/NotificationsList.js
jeanetteobr: I modified it because the Error notification item timestamp needed to be transformed into a moment object in order for the return value to be human-readable. Otherwise, the timestamp rendered as a string that looks like encoded JSON.

jeanetteobr:
@brum @kialam My PR is failing on OrganizationStatistics.spec.js. Not sure why -- any ideas?

PR #399, opened 2021-02-05
Reseting the filter to the `initial` state on that page

brumik:
This fix changes the behavior:
**Before**:
Clicking on the `Clear All Filters` removed ALL the filter from the query (meaning all the filters were set to `null`, `''` or `[]`)
**After**:
Clicking on the `Clear All Filters` restores the `initial` state of the filter on that page. This prevents accidental invalid values to the API, which produced errors on some pages.


kialam (approved):
LGTM! I had that on my list of things to address later when working on the clusters conversion:

Screen Shot 2021-02-05 at 8 19 43 AM

I am glad you got to it before me. :)

PR #404, opened 2021-02-09
Add switch to pie chart legend

brumik:
Working switches:

Screenshot from 2021-02-10 12-33-05

I also refactored the two components while trying to debug to functional components with the following process: 
* `method => arrow function`
* `state => useState`
* `onMount/onUnmount => useEffect(() => {}, []);`

It made easier to debug as we are usually much more familiar with functional components.


kialam (approved):
lgtm - this will be a nice feature to have for the pie charts - nice work!

PR #405, opened 2021-02-09
Expanded filter dropdown bugfix

jeanetteobr:
This PR fixes a bug that causes filter dropdowns to go off-page when there are several options. 

### Verification 

**Before**
Screen Shot 2021-02-09 at 1 16 30 PM

**Fixed**
Screen Shot 2021-02-09 at 1 12 53 PM

### DIFF
--- a/src/Components/Toolbar/ToolbarFilterItem.js
+++ b/src/Components/Toolbar/ToolbarFilterItem.js
@@ -71,6 +71,19 @@ const ToolbarFilterItem = ({
         }
     };
 
+    const onFilter = event => {
+        const textInput = event.target.value;
+        if (textInput === '') {
+            return renderValues(values);
+        } else {
+            return renderValues(
+                values.filter(({ value }) =>
+                    value.toLowerCase().includes(textInput.toLowerCase())
+                )
+            );
+        }
+    };
+
     return (
         <ToolbarFilter
             data-cy={categoryKey}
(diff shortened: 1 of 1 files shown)


brumik (approved):
:+1:

kialam (changes requested):
We need to add fix for when the dropdown is too wide as well. If you visit the Clusters page on CI/Beta and select the Clusters dropdown, you can see where the dropdown takes up more than half the width of the page.

kialam (commented):
Another way to approach this is to leverage the css variables already provided to us by PF:

```
overflow: auto;
max-height: calc(var(--pf-c-select__menu-item--LineHeight) * var(--pf-c-select__menu-item--FontSize) * 15);
```
^^ will give us 10 menu items

kialam:
@jeanetteobr I am going to leave this up to you to include the fix for the wide select dropdown as part of this PR or a new one. Just didn't want the issue to be forgotten. I'll approve this if it's ready to go.

kialam (approved):

jeanetteobr:
@kialam @brumik  I would like to include the fix for the wide expanded menu in this PR as well. I'm working on it now. Thanks!

brumik (changes requested):
You are removing my changes to the toolbar filter. 

`onFilter` function, `onFilter` and `hasInlineFilter` props from select (and maybe more, but I think these are the main removals)

kialam (changes requested):

jeanetteobr:
@kialam @brumik  I am so terribly sorry about that! I will add that functionality back. Thanks for pointing it out!

jeanetteobr:
@kialam @brumik The missing props and onFilter method have been added back to this PR.

brumik (approved):
LGTM

kialam (approved):

PR #406, opened 2021-02-10
Add host chart to org page

brumik:
Merge first: #404 

Results:
Screenshot from 2021-02-10 15-35-41


kialam (changes requested):
We should remove this header bar:
orgstatspanel

kialam:
Clicking through from the host chart to the job explorer gives an error:

hosts-to-jobs

brumik:
@kialam It should be fine now, however, note that clicking on the `other` category is disabled (if id === -1)

kialam:
2 things:
1. Should the tooltip say "81 hosts" instead of "81 jobs"?
Screen Shot 2021-02-10 at 11 27 01 PM

2. @Ladas what should the click through for the hosts by org chart to the Job Explorer show? Right now it's showing jobs but the numbers don't match up with the numbers from the hosts chart (e.g. 81 hosts -> click -> 16 jobs)

Ladas:
@kialam @brumik hm, the title could say `81 unique hosts` we can also show `X hosts` , so how many unique hosts were configured vs. how many hosts configured in total (including 1 host being configured multiple times)

The click through should be disabled now, until we can render more data types in the table view. The click through would point to host explorer API grouped by host, with org and date filtered. (so you can paginate through all of the the 81 unique hosts)

kialam:
@Ladas I'd rather we not add this hosts chart if that means we have to disable the click-through to the job explorer - seems like it's taking away a feature we already have in other areas of our app. Alternatively, we could build out a the host explorer view, but that's getting out of scope for a v1 refactor.

Ladas:
@kialam it still adds value, even with the click through disabled. We can add the destination for the click through later

brumik:
So we can have the tabs and on the hosts charts no clicktrough?

kialam:
@brumik yes, we can just disable the click through for the hosts chart. Sorry for the delay in getting back to you on this.

brumik:
* The tooltip now shows `81 Hosts` in the host chart.
* The clickthrough works on Orgs but is disabled on hosts chart --> on host chart the cursor is not changing to a pointer.

Should be there more changes to the tab name or tooltip? @kialam

kialam (approved):
🎉

PR #408, opened 2021-02-11
Added switches back to the top templates on the roi page

brumik:
Screenshot from 2021-02-11 15-48-21

The switches are not icons as they used to be. Will ot be fine with a normal switch component or should we find the original switch icons?

### DIFF
--- a/src/Containers/AutomationCalculator/TopTemplates.js
+++ b/src/Containers/AutomationCalculator/TopTemplates.js
@@ -12,7 +12,11 @@ import {
     Tooltip,
     Popover
 } from '@patternfly/react-core';
-import { InfoCircleIcon } from '@patternfly/react-icons';
+import {
+    InfoCircleIcon,
+    ToggleOnIcon,
+    ToggleOffIcon
+} from '@patternfly/react-icons';
 
 import {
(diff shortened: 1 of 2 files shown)


kialam (changes requested):
This should match what was there previously:

toggles

Placement is incorrect and the toggle component should be an icon.

src/Containers/AutomationCalculator/TopTemplates.js
```diff
@@ -101,46 +105,49 @@ QuestionIconTooltip.propTypes = {
...
     setDataRunTime = () => {},
+    setEnabled = () => {},
     redirectToJobExplorer = () => {}
 }) => (
     <Card style={{ overflow: 'auto', flex: '1 1 0' }} className="top-templates">
         <CardBody>
             <p>Enter the time it takes to run the following templates manually.</p>
-            {data.map(data => (
-                <div key={data.id}>
+            {data.map(item => (
```
kialam, line 114: Prefer destructuring here as opposed to dot notation.

kialam (approved):
Looks good. I had one request to use object destructuring when mapping over an array but it is not a blocker for merging this PR.

PR #410, opened 2021-02-11
Move filter toolbar to the top title section per UXD's suggestion.

kialam:
After changes:
Screen Shot 2021-02-11 at 10 36 12 AM
Screen Shot 2021-02-11 at 10 36 06 AM


brumik:
If we are moving it should not be moved on all pages? (Or is this PR still WIP?)

kialam:
So it needs to be moved on the Orgs and Automation calculator page because the filters are top level. For the Job Explorer it only applies to the table so it does not need to be moved. Same goes for the Notifications page.

brumik:
It seems a little strange for me, since we have nothing else on these pages, so even when the filter is applied only on the table, it is basically affecting the whole page. But I can accept that later it will be changing maybe and more content will be added to the pages which use different filters maybe.

brumik (approved):
LGTM

PR #411, opened 2021-02-12
Fixed clear filter button on job explorer

brumik:
It was cause by the fact that the initial query param was updated at 
every rerender from the search. Therefore the clear all button reseted 
the state to the CURRENT state.


kialam (approved):
LGTM! We should try and consolidate how the "Clear all" works but that will probably happen once search queries become bookmark-able.

PR #415, opened 2021-02-16
AA-352 Improve Sort UX in Automation Calculator

kialam:
Spoke with @trahman73 on this and the fix is to:

1. Dynamiclaly show in the UI on the right side Templates list column the sort by details:
sortBy_ROI_templates_list

2. Include all the sort by attributes/values returned by the API in the popover like so:
Screen Shot 2021-02-15 at 8 17 10 PM

### DIFF
--- a/src/Containers/AutomationCalculator/TopTemplates.js
+++ b/src/Containers/AutomationCalculator/TopTemplates.js
@@ -20,7 +20,8 @@ import {
 
 import {
     convertSecondsToMins,
-    convertMinsToSeconds
+    convertMinsToSeconds,
+    capitalize
 } from '../../Utilities/helpers';
 
 const TemplateDetail = styled.div`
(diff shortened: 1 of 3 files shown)


benthomasson:
https://issues.redhat.com/browse/AA-352

src/Containers/AutomationCalculator/TopTemplates.js
```diff
@@ -76,98 +73,93 @@ const InputAndText = styled.div`
...
+    sortBy = null,
     setDataRunTime = () => {},
     setEnabled = () => {},
     redirectToJobExplorer = () => {}
 }) => (
     <Card style={{ overflow: 'auto', flex: '1 1 0' }} className="top-templates">
         <CardBody>
             <p>Enter the time it takes to run the following templates manually.</p>
-            {data.map(({
-                id,
```
brumik, line 121: Did you do this by accident? I am pretty sure I just destructured in a recent commit this object on your advice.

brumik (changes requested):
Also if we merge the #416 first we could pass `sortAttr` directly instead of `sortBy`.

Except for the default value of the `sortBy` everything else is a nitpick (non-blockers).

src/Containers/AutomationCalculator/TopTemplates.js
kialam: No, this is intentional. We are now passing in the entire api response to show in the tooltip so it makes less sense to destructure everything out in this case.

brumik (approved):
Thanks for including my nitpicks ^^ :heart:

PR #416, opened 2021-02-16
Add toggle icon for sort order in filter toolbar.

kialam:
This PR is an attempt to hook up the toggle icon by the Sort By dropdown in the filter toolbar:
exhibit1

This would allow for the following changes:
1. Sort by options inside the drop down will be cut down by half, since the API will no longer be sending repeat options. So instead of `id:asc` and `id:desc`, the API will send the following sort options: `id`. The UI will be responsible for handling the sort order (`'asc', 'desc'`) and send this back to the API in the form of `sort_by: id:asc`.
2. The toggle icon will change depending on sort order selected:
Screen Shot 2021-02-26 at 12 08 12 PM <- correlates to `ascending` order.
Screen Shot 2021-02-26 at 12 07 55 PM <- correlates to `descending` order.

Automation Calculator page showcasing new changes:
toggle_asc_desc

**_### STOP, DO NOT PASS GO, DO NOT COLLECT $200_**
**_### STOP, DO NOT PASS GO, DO NOT COLLECT $200_**
**_### STOP, DO NOT PASS GO, DO NOT COLLECT $200_**
**_### STOP, DO NOT PASS GO, DO NOT COLLECT $200_**
**_### STOP, DO NOT PASS GO, DO NOT COLLECT $200_**
**_### STOP, DO NOT PASS GO, DO NOT COLLECT $200_**

Now that I have your attention, there is an issue that needs to be addressed in order to move forward:
1. Org Stats page sort options are inconsistent with rest of app. As implemented and in attempt to maintain _existing_ UX from the v1 Org Stats refactor, the dropdown is rendered useless as we only really care about the toggle dictating the order of the hard-coded sort attributes.
orgs_sort_weird

I have documented this very issue here in JIRA: https://issues.redhat.com/browse/AA-399
My recommendation is to remove the sort dropdown altogether on the Org Stats page. We would then need to make small refactors to our code to yank out all the edge-cases we've written for this one UX element in this one view.

I have also found a bug with the job explorer sort by options being broken as a result of this change.

**_TLDR;_**
https://issues.redhat.com/browse/AA-360 is blocked by https://issues.redhat.com/browse/AA-399


benthomasson:
https://issues.redhat.com/browse/AA-360

src/Components/Toolbar/Toolbar.js
```diff
@@ -102,13 +103,21 @@ const FilterableToolbar = ({
         <ToolbarGroup variant="filter-group">
             <ToolbarFilterItem
                 categoryKey="sort_by"
-                filter={filters.sort_by}
+                filter={filters.sort_attr}
                 values={sort_by}
-                setFilter={value => setFilters('sort_by', value)}
+                setFilter={value => setFilters('sort_attr', value.split(':')[0])}
```
brumik, line 108: We will need to remove the split

brumik (commented):
I think this PR should wait for the backend PR to return the sort values without the :asc/:desc values?

src/Components/Toolbar/Toolbar.js
kialam: No, iirc sort_by will always be `attr:desc/:asc`.

brumik: Oh, wasn't @AparnaKarve PR was going to remove the `:asc/:desc` from the sort by options? (I cannot find the link to the backend pr, maybe somebody could link it here?)

kialam: Yes, but we still need to send it as `attr:asc/:desc` to the API and our queryParams do that (in PR #416)

brumik: This is PR #416 :confused: ? I am confused now

AparnaKarve: backend PR - https://github.com/RedHatInsights/tower-analytics-backend/pull/670

kialam: @AparnaKarve thanks

src/Api.js
```diff
@@ -101,7 +101,7 @@ export const readJobEventsByOrg = ({ params = {}}) => {
         include_others: true,
         granularity: 'daily',
         attributes: [ 'host_task_count' ],
-        sort_by: `host_task_count:${params.sort_by}`
+        sort_by: `host_task_count:${params.sort_order}`
```
jlmitch5, line 104: I guess we will move these changes to org stats based on the other pr doing that.

src/Api.js
```diff
@@ -164,6 +164,31 @@ export const readJobExplorer = ({ params = {}}) => {
     }).then(handleResponse);
 };
 
+export const readJobsGroupByTemplate = ({ params = {}}) => {
```
jlmitch5, line 167: I feel like these two functions should be updated to "readJobExplorer" and the params should be moved to clusters like you did with the org stats page in the other pr.

src/Components/Toolbar/Toolbar.js
```diff
@@ -102,13 +103,21 @@ const FilterableToolbar = ({
...
+                filter={filters.sort_attr}
                 values={sort_by}
-                setFilter={value => setFilters('sort_by', value)}
+                setFilter={value => setFilters('sort_attr', value)}
                 hasChips={false}
             />
-            <Button variant="control">
-                <SortAmountDownIcon />
+            <Button variant="control"
+                onClick={() =>
```
jlmitch5, line 112: woo!  this is definitely feeling how it should work I think!

src/Containers/OrganizationStatistics/OrganizationStatistics.js
```diff
@@ -77,8 +77,8 @@ const TopCard = styled(Card)`
 const colorFunc = scaleOrdinal(pfmulti);
 
 const optionsMapper = options => {
-    const { meta, inventory_id, ...rest } = options;
-    return { ...rest, sort_by: meta.sort_by };
+    const { inventory_id, ...rest } = options;
+    return { ...rest, sort_by: [{ key: 'Order', value: 'Ascending/Descending' }]};
```
jlmitch5, line 81: it sounds like we decided to remove this in the backlog grooming meeting.  should we just do that as part of this pr or as a follow up?  either is cool.

src/Utilities/useQueryParams.js
```diff
@@ -63,19 +67,48 @@ export const useQueryParams = initial => {
                 return { ...state, ...newValues };
             }
 
+            // I know this is ugly, we could put it into a middleware but I had
```
jlmitch5, line 70: I would remove this comment, and add a comment as to why we are keeping this around in the two formats (which makes sense now that it's been explained to me, thanks @kialam. colon-separated for api use, split out into separate attrs for toolbar use)

jlmitch5 (commented):
This is looking great!  I think this squares with how I would expect the sorting functionality to work as a user.

I think there will need to be some updates to this based on the api.js work in https://github.com/RedHatInsights/tower-analytics-frontend/pull/424

Otherwise, I think this all makes sense and looks great!

jlmitch5:
I have updated this PR and I believe it is ready for review, testing and merge.  Things I did:
- rebased to devel, updating new fetch calls to using the authenticatedFetch wrapper
- updated to getting new sort_options from api to display just the sortable attributes once (thanks @AparnaKarve for help understanding this).  In other words, this dropdown is no longer duplicated twice, appending each option with "ascending"/"descending".  **NOTE** that this requires api changes that are not yet pushed to cloud.redhat.com.  I had to run the api locally to get the options endpoint to update.
- updating all references of sort_attr to sort_options for consistency with the way this is phrased in the api
- fixed an issue where when sort_by is not set, updating the sort order or option could cause the page to error out.  I updated the useReducer actions to either look at "sort_by" if available, or the sort_options/sort_order keys on state.

src/Utilities/constants.js
```diff
@@ -56,7 +58,8 @@ export const organizationStatistics = {
         status: [],
         org_id: [],
         quick_date_range: 'last_30_days',
-        sort_by: 'desc',
+        sort_options: 'order',
```
jlmitch5, line 61: I'm not sure what the correct option is here.  Somehow, I had code that had it "Order" capitalized, which didn't look right.  @AparnaKarve do you know what the default sort option is supposed to be for the org stats page?

src/Utilities/useQueryParams.js
```diff
@@ -4,6 +4,11 @@ import moment from 'moment';
 import { formatDate } from '../Utilities/helpers';
 
 export const useQueryParams = initial => {
+    const initialWithCalculatedParams = {
```
jlmitch5, line 7: You would think this would make it so that sort_by is always there, but apparently not.

src/Utilities/constants.js
AparnaKarve: @jlmitch5 As far as I know, the dashboard pages (clusters and org stats) do not have sort options.

src/Utilities/useQueryParams.js
brumik: Can you describe the workflow how you were able produce queryPramas values without sort_by? I was only able to do `sort_by: 'undefined:undefined'

src/Utilities/useQueryParams.js
```diff
@@ -63,19 +67,59 @@ export const useQueryParams = initial => {
                 return { ...state, ...newValues };
             }
 
+            // options and order are used for toolbar filter display purposes
+            case 'SET_SORT_OPTIONS': {
+                let order;
+                if (state?.sort_by) {
```
brumik, line 73: * Can you clarify why is this logic needed here?
* I could not find it, what exactly the `?` does in the `state?`
* We could make it more compact by:
`const order = state.sort_by ? state.sort_by.split(':')[1] : state.sort_order`

src/Utilities/useQueryParams.js
```diff
@@ -63,19 +67,59 @@ export const useQueryParams = initial => {
...
+                return {
+                    ...state,
+                    sort_options: value.sort_options,
+                    sort_by: `${value.sort_options}:${order}` // Update sort by
+                };
+            }
+
+            case 'SET_SORT_ORDER': {
+                let attr;
+                if (state?.sort_by) {
```
brumik, line 88: The same as above

src/Utilities/constants.js
brumik: You don't have to, you can omit the whole line. The sort by is overwritten in the Api.js. As I told few days before we probably want to standardize and maybe a little rework the Api.js and the default query params (we may organize them by changeable and fixed for each call)

brumik (changes requested):
It looks good and works, I have some questions regarding the code.

Otherwise, we should wait for the backend to be merged, we should track it in JIRA but we can link the backend PR here too

brumik (approved):
LGTM

PR #417, opened 2021-02-17
[Showcase] Sort by discussion example

brumik:
Build upon #416


brumik:
Closing in favout of #416

PR #421, opened 2021-02-22
Typescript integration; update redux

kialam:


package.json
```diff
@@ -19,7 +19,8 @@
     "react-router-dom": "^5.2.0",
     "redux": "^3.7.2",
     "redux-logger": "^3.0.6",
-    "redux-promise-middleware": "^5.1.1"
+    "redux-promise-middleware": "^5.1.1",
+    "redux-toolkit": "^1.1.2"
```
brumik, line 23: What is this for?

kialam: It's the official library from redux for faster integration and includes a recommended way of writing redux logic. It includes these [helpers](https://redux-toolkit.js.org/introduction/getting-started#whats-included) . I was particularly interested in their `configureStore()` and `createReducer()` helpers to assist with readability and removing some boilerplate from the original redux library itself.

PR #424, opened 2021-02-26
Simplified Api.js and refactored Orgs page

kialam:
- Remove page-specific api methods in favor of generic ones
- Offload page-specific params to view component


src/Api.js
```diff
@@ -1,18 +1,21 @@
 import { stringify } from 'query-string';
-
+/* v0 endpoints */
 const clustersEndpoint = `/api/tower-analytics/v0/clusters/`;
 const notificationsEndPoint = `/api/tower-analytics/v0/notifications/`;
```
jlmitch5, line 4: while we are in here could we uncapitalize the "P"

src/Api.js
```diff
@@ -38,136 +41,41 @@ function handleResponse(response) {
...
-export const readNotifications = ({ params = {}}) => {
-    const formattedUrl = getAbsoluteUrl();
-    let url = new URL(notificationsEndPoint, formattedUrl);
-    Object.keys(params).forEach(key => url.searchParams.append(key, params[key]));
-    return fetch(url).then(handleResponse);
-};
-
-export const readJobExplorerOptions = ({ params = {}}) => {
+    const qs = stringify(paginationParams);
     const formattedUrl = getAbsoluteUrl();
```
jlmitch5, line 52: might be cool to get rid of these forrmattedUrl vars and just put getAbsoluteUrl() direct in the new URL call

src/Api.js
```diff
@@ -38,136 +41,41 @@ function handleResponse(response) {
...
-    let url = new URL(jobExplorerOptionsEndpoint, formattedUrl);
+    let url = new URL(jobExplorerEndpoint, formattedUrl);
+    url.search = qs;
     return fetch(url, {
```
jlmitch5, line 55: I think we will need to update these to authenticatedFetch

src/Api.js
```diff
@@ -228,3 +134,28 @@ export const readROIOptions = ({ params = {}}) => {
...
+export const readJobExplorerOptions = ({ params = {}}) => {
+    const formattedUrl = getAbsoluteUrl();
+    let url = new URL(jobExplorerOptionsEndpoint, formattedUrl);
+    return fetch(url, {
+        method: 'POST',
+        body: JSON.stringify(params)
+    }).then(handleResponse);
+};
+
+export const preflightRequest = () => {
```
jlmitch5, line 159: can we move preflightRequest above the v0 section, I think it's still in spec?

src/Containers/OrganizationStatistics/OrganizationStatistics.js
```diff
@@ -176,13 +207,14 @@ const OrganizationStatistics = ({ history }) => {
     }, []);
 
     useEffect(() => {
-        const apiPromise = chartMapper[activeTabKey].api;
-        setOrgs(apiPromise({ params: queryParams }));
+        const { api: readJobsOrHosts } = chartMapper[activeTabKey];
+        const params = activeTabKey === 0 ? jobsByDateAndOrgParams : hostAcrossOrgParams;
+        setOrgs(readJobsOrHosts({ params }));
```
jlmitch5, line 212: totally fine with how it is here.  inlining it `setOrgs(chartMapper[activeTabKey].api({ params }));` might be slightly easier to parse, but either works

jlmitch5 (commented):
I think this is very nice.  It keeps the api file a bit simpler and mapped more 1:1 with endpoints, and it moves these params to the page where it makes sense to define them (better "separation of concerns").  I think this refactor will help us with maintaining this page and keep the api file from getting bloated.

All my suggestions are just low priority, small tweaks to potentially help tighten things up

src/Api.js
```diff
@@ -1,18 +1,21 @@
...
+/* page options endpoints */
+const jobExplorerOptionsEndpoint =
+    '/api/tower-analytics/v1/job_explorer_options/';
 const ROITemplatesOptionsEndpoint = '/api/tower-analytics/v1/roi_templates_options/';
 const orgOptionsEndpoint = '/api/tower-analytics/v1/dashboard_organization_statistics_options/';
-const hostExplorerEndpoint = '/api/tower-analytics/v1/host_explorer';
 const clustersOptionsEndpoint = '/api/tower-analytics/v1/dashboard_clusters_options/';
-const eventExplorerEndpoint = '/api/tower-analytics/v1/event_explorer/';
 
 function getAbsoluteUrl() {
```
brumik, line 20: If we are here, can we convert these two functions to arrow functions, since we are using them everywhere.

The `getAbsoluteUrl` can be also simplified to onliner:
```js
const getAbsoluteUrl = () => window.location.href.split('/').slice(0, -1).join('/');
```

brumik: Or actually, we can just use `window.location.origin` which should give the same results more reliable.
Screenshot from 2021-03-14 16-37-50

src/Api.js
kialam: Good catch

src/Api.js
kialam: Yes, `window.location.origin` is much cleaner.

jlmitch5 (changes requested):
Again really loving that endpoints map 1:1 in the api file now, and queryParams are in the page files.  This is great!

src/Containers/OrganizationStatistics/OrganizationStatistics.js
```diff
@@ -166,6 +164,39 @@ const OrganizationStatistics = ({ history }) => {
         constants.defaultParams
     );
 
+    const jobEventsByOrgParams = {
+        ...queryParams,
+        group_by: 'org',
+        include_others: true,
+        granularity: 'daily',
+        attributes: [ 'host_task_count' ],
+        sort_by: `host_task_count:${queryParams.sort_by}`
```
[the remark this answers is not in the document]
kialam: Ah yes, I lost my changes after the merge/rebase. Good catch!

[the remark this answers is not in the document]
kialam: @jlmitch5 We also decided to hardcode those values. :)

jlmitch5 (approved):
nice!

PR #430, opened 2021-03-04
utilize centralized authenticated fetch api helper

jlmitch5:
this fixes https://issues.redhat.com/browse/AA-353

For whatever reason, the useApi hook did not seem to be successfully gating requests until they were finished logging in.  I also found a number of places (that I guess have not been converted to this useApi hook yet).  So here is an attempt at a simple and consistent solution for gating api requests to make sure they are logged in first.

This pr does two things:
- removes `window.insights.chrome.auth.getUser()` across the app
- wraps fetch calls in api.js (which seemed to be the centralized place that stores alll api requests) in a simple wrapper which calls `window.insights.chrome.auth.getUser()`, waits for a response, and then returns fetch

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -38,14 +38,18 @@ function handleResponse(response) {
     });
 }
 
+function authenticatedFetch(endpoint, options) {
+    return window.insights.chrome.auth.getUser().then(() => fetch(endpoint, options));
+}
+
 export const preflightRequest = () => {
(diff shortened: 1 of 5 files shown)


benthomasson (approved):

brumik (approved):
I don't like that the cypress is not passing but it is caused by the fact that your forked repo does not have the required secrets.

brumik:
The cypress tests are passing locally, merging.

PR #431, opened 2021-03-05
Codecov upload moving to Github Actions

brumik:


jlmitch5 (approved):
Nice, this move from travis to github actions continues to be easy to follow in these config files.

PR #437, opened 2021-03-15
Fix reset filter so that it returns initial values.

kialam:


src/Utilities/constants.js
```diff
@@ -20,8 +20,7 @@ export const jobExplorer = {
         template_id: [],
         sort_options: 'created',
         sort_order: 'desc',
-        start_date: null,
-        end_date: null,
+        sort_by: 'created:desc',
```
jlmitch5, line 23: sort_by is now purely computed in useQueryParams so you don't need it here

jlmitch5 (changes requested):
Looks good, just think we don't need sort_by in the constants file

jlmitch5 (approved):
woo!

PR #441, opened 2021-03-17
update cypress nav selector to one that works both locally and with CI

jlmitch5:


kialam (approved):
LGTM. We can run cypress tests locally and wait until there's a better element selector.

PR #442, opened 2021-03-18
Fix AA-429 pie chart values

kialam:
There's a lot of noise in this PR as I re-ordered the list of params for each of the charts so they're ordered alphabetically for ease of readability, but the main fix for this bug is to [change out the `attribute` param in `jobRunsByOrgParams ` from `host_count` to `total_count`](https://github.com/RedHatInsights/tower-analytics-frontend/commit/9d269f3fec018eb31f0148aaca73da7959b90c40).


Ladas (approved):
:+1:

jlmitch5 (approved):
looks good

PR #443, opened 2021-03-18
scripted tmux config for fully local ui dev setup

jlmitch5:
After discussing my thoughts around documenting the fully local setup with @benthomasson, he had the idea that trying to automate it may help make this stuff more impervious to being out of date.  So here's my very rudimentary attempt at doing that.  It definitely needs beefing up before it is useful and ready to merge, I think.

Basically, it is a new phony make target which starts and attaches to a tmux session which contains all the main pieces for a fully local UI dev setup:
- cloud-services-config for custom side nav
- frontend running webpack dev (i.e. always updating the bundle as changes are made to the frontend src)
- backend ran using headless (i.e. no ui cotainer) make ui target
- insights-proxy running using the fast-api config to connect all these pieces.

Iteration 1 looks like this:
Screen Shot 2021-03-18 at 4 34 03 PM

---

**Some things to note and that need iterating below:**

sometimes (about 50% of the time for me) the kafka container fails.  this is recognizable through the backend log output consistently saying that it is trying to connect to kafka indefinitely.  luckily, this has an easy fix, either restarting on the cli in another terminal shell, or starting the container in the docker UI:

Screen Shot 2021-03-18 at 4 54 38 PM

Luckily, you don't need to bounce anything else...once kafka is started back up everything magically connects and works.  Because of this, I don't think this runner needs to implement anything fancy like retry logic.  maybe it could print out something somewhere that if this happens that you might need to restart it.

---

ports are current non configurable.

---

this is just for **running** the processes needed.  all bootstrapping commands (npm install on frontend, various make build/pull/data/etc. commands on backend) need to be done before this target will work.  maybe there's some sort of local build and clean targets that should be written to handle all of that mangement?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,12 @@
 .PHONY: all build fully_local_dev_start
 
 BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
+GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
+
+REL_BACKEND_PATH ?= ../tower-analytics-backend
+REL_PROXY_PATH ?= ../insights-proxy
+REL_PROXY_CONFIG_PATH ?= ../tower-analytics-frontend/config/fast-api.spandx.config.js bash scripts/run.sh
+REL_CRC_CONFIG_PATH ?= ../cloud-services-config
 
 
 all: build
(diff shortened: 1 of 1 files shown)


jlmitch5:
Note that you will need tmux for this to work.  `brew install tmux` for mac, I believe it's a standard linux package, but `yum/apt-get install tmux` should work if not.

jlmitch5:
Screen Shot 2021-03-19 at 1 42 48 PM
Screen Shot 2021-03-19 at 1 28 20 PM

Now with better layout!

jlmitch5:
I've made all locations able to be changed by setting environment variables

jlmitch5:
I've updated the target to remove the unit test runner (thought about it and it's probably best just to have the target be as minimal as possible so I didn't add the extra shell), and I added the mac tmux installing target.  I think this is ready for merge now if y'all are cool with it.

brumik:
Hey, I think it is good to go (but I don't feel comfortable approving this without @kialam). Maybe it would be nice to put a little description of this feature in the Readme?

Actually, we can do it in a follow-up PR, because our readme doesn't really reflect our workflow anymore and we are still open source. If you are doing the initial setup scripts too then we can do in that PR

jlmitch5:
that all sounds good, thanks @brumik.  Agreed with updating the readme, I'm good with waiting until we get setup in here as well.

kialam (approved):

PR #449, opened 2021-04-07
update chart axis ticks:

jlmitch5:
This updates the number of labeled ticks on the x axis of the org stats chart.  For grouped bar I go to every 2 for month view, and every 4 for 2 month view.

Screen Shot 2021-04-07 at 2 06 24 PM

I also removed timeFrame data prop to clean things up for consistency across all the different charts (unused in pie charts, and we were just finding the length of the data on line/bar).


kialam (approved):
Looks great! Nice work and good call on every 4 for 2 month datasets :)

PR #451, opened 2021-04-10
Updated orgs page legend from the api

brumik:
The API now provides a legend field in the meta, where the `Others` has the correct number before it. This PR making use of this extra information, displaying the correct number of others in the legend.

**Additionally:**
- The `name` prop could not be used for colors coding, since the `X Other` string could be different in the charts, so the `id` is used instead
- The grouped bar chart groups used the `name` as the domain for ordering, but since the `name` is now different in the series (since in every series there can be a different amount of `Others`) the `id` is used for specifying the subdomain bandwidth.

**CSS Update:**
Played around with the CSS in the legend to make the count align nicely and consistently to the right. 

**Before**:
Screenshot from 2021-04-10 11-33-12

**After**:
Screenshot from 2021-04-10 11-30-54

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "webpack-cli": "^3.2.3"
   },
   "scripts": {
-    "proxy": "SPANDX_CONFIG=./config/spandx.config.js bash ./insights-proxy/scripts/run.sh",
+    "proxy:ci": "SPANDX_CONFIG=./config/spandx.config.js bash ./insights-proxy/scripts/run.sh",
     "proxy:local": "SPANDX_CONFIG=./config/spandx.config.js bash ../insights-proxy/scripts/run.sh",
     "build": "webpack --config config/prod.webpack.config.js",
     "test": "jest --verbose",
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
   },
   "scripts": {
     "proxy": "SPANDX_CONFIG=./config/spandx.config.js bash ./insights-proxy/scripts/run.sh",
+    "proxy:local": "SPANDX_CONFIG=./config/spandx.config.js bash ../insights-proxy/scripts/run.sh",
     "build": "webpack --config config/prod.webpack.config.js",
     "test": "jest --verbose",
     "test_dev": "jest --coverage=false --watch",
--- a/src/Charts/GroupedBarChart/GroupedBarChart.js
+++ b/src/Charts/GroupedBarChart/GroupedBarChart.js
@@ -21,19 +21,18 @@ const formatDate = date => {
 const GroupedBarChart = ({
     onClick = null,
     TooltipClass,
+    legend,
     ...props
 }) => {
-    const orgsList = props.data[0].items;
-    const colors = orgsList.map(org => {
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -91,7 +91,8 @@
     "webpack-cli": "^3.2.3"
   },
   "scripts": {
-    "proxy": "SPANDX_CONFIG=./config/spandx.config.js bash ./insights-proxy/scripts/run.sh",
+    "proxy:ci": "SPANDX_CONFIG=./config/spandx.config.js bash ./insights-proxy/scripts/run.sh",
```
jlmitch5, line 94: nice

src/Charts/GroupedBarChart/GroupedBarChart.js
```diff
@@ -236,6 +234,7 @@ const GroupedBarChart = ({
 GroupedBarChart.propTypes = {
     id: PropTypes.string,
     data: PropTypes.array,
+    legend: PropTypes.array,
```
jlmitch5, line 237: all these changes look really good.  the only thing I'm curious about is if anything throws errors at the "there is no data returned fo legend" case.

src/Charts/PieChart.js
```diff
@@ -171,13 +171,12 @@ const PieChart = ({
     getWidth,
     getHeight
 }) => {
-    const colors = data.map(org => {
-        const name = org.id === -1 ? 'Others' : org.name;
+    const colors = data.map(({ id, name, count }) => {
```
jlmitch5, line 174: nice destructuring in this map, I think it makes the return a bit more readable/understandable.

src/Utilities/Legend.js
```diff
@@ -73,13 +65,13 @@ const Legend = ({
...
-                            <SubTitle>{ count }</SubTitle>
-                        ) }
+                        {count && (
+                            <SubTitle>{count}</SubTitle>
+                        )}
                         { selected && (
                             <Switch
                                 isChecked={ selected.some(selection => selection === id) }
-                                onChange={ handleChange }
+                                onChange={ () => onToggle(id) }
```
jlmitch5, line 74: well that's definitely simpler that what was happening before, trying to parse the value, nice!

jlmitch5 (approved):
Great PR!  Everything looks good to me.  The one thing I want to make sure is that the case where the api returns no data (and thus no items for the legend) doesn't cause a bunch of errors to spit out/break things.

src/Charts/GroupedBarChart/GroupedBarChart.js
brumik: I am pretty sure if the legend is missing then all the charts fail to load with some `calling map on undefined` errors. There is no sensible fallback option. Since the API returns the legend in the object with the data, I think we can be pretty sure that it will be always there if the call to the API is successful. Unsuccessful calls we are already handling so I think this is fine.

brumik: https://github.com/RedHatInsights/tower-analytics-frontend/blob/devel/src/Containers/OrganizationStatistics/OrganizationStatistics.js#L248

PR #452, opened 2021-04-14
First cut of changes to display RBAC error message and stop loading of other elements on the screen when user does not have access to the screen.

h-kataria:
Added `AlertMessage` component that uses patternfly Alert component, this component can be used/enhanced further to be used as common component to display other notifications to user on screen.

https://issues.redhat.com/browse/AA-322

@benthomasson pls review first cut, hopefully i am on right track.

before
User without RBAC acces to Automation Calculator screen
Screen Shot 2021-04-13 at 10 42 40 PM

User with RBAC acces to Automation Calculator screen
Screen Shot 2021-04-13 at 10 43 14 PM

after
User without RBAC acces to Automation Calculator screen
Screen Shot 2021-04-14 at 4 26 14 PM

User with RBAC acces to Automation Calculator screen
Screen Shot 2021-04-13 at 10 41 42 PM

### DIFF
--- a/src/Containers/AutomationCalculator/AutomationCalculator.js
+++ b/src/Containers/AutomationCalculator/AutomationCalculator.js
@@ -3,10 +3,12 @@ import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import {
     Main,
+    NotAuthorized,
     PageHeader,
     PageHeaderTitle
 } from '@redhat-cloud-services/frontend-components';
 import { Card, CardBody } from '@patternfly/react-core';
+import { notAuthorizedParams } from '../../Utilities/constants';
 
 // Imports from custom components
 import LoadingState from '../../Components/LoadingState';
@@ -218,6 +220,10 @@ const AutomationCalculator = ({ history }) => {
             </Main>
         </WrapperRight>
     );
+    console.log('options', preflight);
+    if (preflight?.error?.status === 403) {
+        return <NotAuthorized { ...notAuthorizedParams } />;
+    }
 
     return (
         <React.Fragment>
--- a/src/Containers/Clusters/Clusters.js
+++ b/src/Containers/Clusters/Clusters.js
@@ -16,9 +16,11 @@ import useApi from '../../Utilities/useApi';
 
 import {
     Main,
+    NotAuthorized,
     PageHeader,
     PageHeaderTitle
 } from '@redhat-cloud-services/frontend-components';
+import { notAuthorizedParams } from '../../Utilities/constants';
 
 import {
     Card,
@@ -168,6 +170,10 @@ const Clusters = () => {
 
         fetchEndpoints();
     }, [ queryParams ]);
+
+    if (preflightError?.preflightError?.status === 403) {
+        return <NotAuthorized { ...notAuthorizedParams } />;
+    }
 
     return (
         <React.Fragment>
--- a/src/Containers/JobExplorer/JobExplorer.js
+++ b/src/Containers/JobExplorer/JobExplorer.js
@@ -10,6 +10,7 @@ import LoadingState from '../../Components/LoadingState';
 import EmptyState from '../../Components/EmptyState';
 import NoResults from '../../Components/NoResults';
 import ApiErrorState from '../../Components/ApiErrorState';
+
 import {
     preflightRequest,
     readJobExplorer,
@@ -19,9 +20,11 @@ import { jobExplorer } from '../../Utilities/constants';
 
 import {
     Main,
+    NotAuthorized,
     PageHeader,
     PageHeaderTitle
 } from '@redhat-cloud-services/frontend-components';
+import { notAuthorizedParams } from '../../Utilities/constants';
 
 import {
     Card,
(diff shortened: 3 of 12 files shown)


benthomasson:
@h-kataria this is the behavior that I'd like to see. @trahman73 do we have a mockup for the RBAC error messages?

brumik:
Is this error specific to this one page? If not we should maybe do a redirect to a specific page where we would display the error to avoid duplicated code.

h-kataria:
> Is this error specific to this one page? If not we should maybe do a redirect to a specific page where we would display the error to avoid duplicated code.

no @brumik this will be applied to all pages evntually, i will look into redirect approach and see how that works out.

trahman73:
@benthomasson RBAC error messages should use the PatternFly empty state component using the example below. 
image

h-kataria:
@brumik i havent been able to get the redirect working, but made some further changes to simplify changes that i made in previous commit, to match the mockup that @trahman73 suggested. I would be happy if you can help me figure out the redirect implementation or if there is an example i can look at, that would be great.

brumik:
@h-kataria We don't have an existing example and all my google searches resulted in using redux store (I was doing this kind of thing with redux too in my personal projects). I tried to play with using redirect in the `Api.js` or in the `useApi.js` hook, but without any success. Maybe @jlmitch5 or @kialam can help without actually starting to use redux. 

I would go (if we all agree) with using redux. Then we can catch the error in the `react router` with a simple rule. (If you search `react-router redux authentication` there are plenty of examples of how it can be done easily).

jlmitch5:
> I would go (if we all agree) with using redux

At this point, I am strongly against adding redux in to this app.  In my opinion and based on my experience, I think that the scale the app is at we would not benefit from the features enough to justify the added overhead of all the boilerplate needed to utilize redux in a best-practice way.

We are already using the useReducer hook in react to deal with more complicated state management in the few key places it is warranted in the app.  While there could be some benefits to having all state at a global area (most of which would be to be able to use the redux dev tools), trying to figure out where state is defined is not a big issue we face in this app during the course of development, in my experience.

@brumik I would encourage you to read this [section](https://redux.js.org/faq/general#when-should-i-use-redux) of the redux docs and the articles linked to from it.  Based on this information from the original creator and current maintainers of the redux project, I think it becomes pretty clear that we don't fit a solid usecase for introducing redux to our app.  Hooks were created to solve a lot of the problems inherit with react and managing state, and in my opinion, we are using them pretty effectively in the codebase at the current moment.

jlmitch5:
@h-kataria and I just met with @trahman73.  The overall flow approach is good here imo.  Yes, it will be done at each page, but that is okay.  I feel like it is similar to how errors and loading is handled at the page level.

I don't think the stuff in this pr should be handled at the api.js level.  The way that api.js works is that it is dealing with managing request/response with the backend purely in json.  I don't think we should introduce the concept of JSX component stuff in that file to maintain proper separation of concerns.

We did look to see if we could add error handling one level above the pages (at the InsightsRoute level where they are defined), but we would have to somehow pass the error "up", which would just be additional indirection and still need to be handled at every page.  The approach in the pr is the most straightforward way of doing this.

 Just one tweak:
- we're going to switch to using the insights-based NotAuthorized component, which basically is all the EmptyState stuff needed for these types of errors so that it is consistent across other apps.

src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -3,10 +3,12 @@ import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import {
     Main,
+    NotAuthorized,
     PageHeader,
     PageHeaderTitle
 } from '@redhat-cloud-services/frontend-components';
 import { Card, CardBody } from '@patternfly/react-core';
+import { notAuthorizedParams } from '../../Utilities/constants';
```
jlmitch5, line 11: nice use of constants!

src/Containers/Clusters/Clusters.js
```diff
@@ -169,6 +170,10 @@ const Clusters = () => {
         fetchEndpoints();
     }, [ queryParams ]);
 
+    if (error && error.error) {
```
jlmitch5, line 173: you can use the optional chaining connector (`error?.error`) to simplify this conditional slightly.

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -121,6 +123,10 @@ const JobExplorer = ({ location: { search }, history }) => {
         setCurrPage(page);
     };
 
+    if (error && error.error) {
```
jlmitch5, line 126: same here

src/Containers/OrganizationStatistics/OrganizationStatistics.js
```diff
@@ -220,6 +222,10 @@ const OrganizationStatistics = ({ history }) => {
         setTasks(readJobExplorer({ params: jobEventsByOrgParams }));
     }, [ queryParams ]);
 
+    if ((jobs && jobs.error) || (options && options.error) || (tasks && tasks.error)) {
```
jlmitch5, line 225: same on these

jlmitch5 (commented):
This looks great, thanks for doing this!

Code-wise, the only thing I see that could be made slightly more compact is the use of the optional chanining operator in a few places so we don't have to && parent objects.  More info about that here:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

jlmitch5 (approved):
awesome stuff @h-kataria ! 🥳

PR #456, opened 2021-04-21
add breadcrumbs

jlmitch5:
as a prerequisite of my savings planner add and @h-kataria's savings planner detail work, this adds support for breadcrumbs based on the pf-core breadcrumb components.

Screen Shot 2021-04-21 at 4 15 02 PM

I initially implemented this using the crc breadcrumb component, but after seeing deprecation messages when writing the tests, I changed to using the pf-core components.  Since the core components utilize href, I followed the entry pattern I found in the codebase to grab the base path using the utility function made for it.

As part of this, it seemed like "window.location.pathname" felt like it should be the default param for that rather than passed across the app so I made that small tweak.

### DIFF
--- a/src/Components/Breadcrumbs.js
+++ b/src/Components/Breadcrumbs.js
@@ -1,0 +1,34 @@
+/* eslint-disable */
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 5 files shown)


jlmitch5:
(the cypress failure is the flake we've identified and not an actual error)

kialam (approved):
LGTM!

PR #460, opened 2021-04-28
update crc config

jlmitch5:
this is an attempt to update @redhat-cloud-services/frontend-components-config to 4.0.28 in order to fix development malformed webpack chunk issues.


brumik (changes requested):
I tried to see trough the changes and have some questions to understand better what is happening. 

Also `Use proxy: false`, 
`You can use webpack proxy (instead of using insights-proxy) by setting "useProxy". Check config documentation to see more details.` - when running `npm start`. We could use that but which documentation it is referring to?

If we could use the proxy from webpack we can reduce the setup for the CI cypress tests which would be really nice (not to mention starting one less services during development) 

It can be that cypress is not passing since it is written for the beta and the `npm start` does not have BETA defined? (I am not sure what that variable does).

I am not sure if this is the scope of the PR but also we can setup in this PR the `don't fail the compile on linting error but write the linting errors to the browser console`

brumik:
Closing in favour https://github.com/RedHatInsights/tower-analytics-frontend/pull/464

PR #461, opened 2021-04-29
Bugfix: the toolbar not closes anymore on change unexpected

brumik:
Does what the title says. Additionally because of the new changes put some components into folders, so it is a little more intuitive.


src/Components/Toolbar/Groups/SettingsPanel.js
```diff
@@ -0,0 +1,73 @@
...
+                    variant="plain"
+                    onClick={() => setSettingsExpanded(!settingsExpanded)}
+                >
+                    <TimesIcon />
+                </Button>
+            </CardActions>
+            <CardTitle>Settings</CardTitle>
+        </CardHeader>
+        <CardBody>
+            <Switch
```
jlmitch5, line 40: this feels like something that would be specific to the job explorer list, and would probably be better defined there, but I understand that is out of the scope of the pr.

jlmitch5 (approved):
I like this PR, I think it is a nice clean up of the Toolbar component.  The overall Toolbar.js is tightened, and each section of the toolbar contains things relevant to it.

I think that the settings panel feels like it is hiding some page-specific code in it, but I understand that the scope of this PR wasn't actually changing the behavior or how the code in Toolbar was structured, just moving to some sub components.

src/Components/Toolbar/Groups/SettingsPanel.js
brumik: I am happy to look at this after this PR, we can talk about it in the standup maybe

PR #462, opened 2021-05-03
Savings Planner List View

kialam:
This PR adds the Savings Planner List view to our application.

Screen Shot 2021-05-03 at 1 05 39 PM


src/Utilities/useQueryParams.js
```diff
@@ -46,6 +46,10 @@ export const useQueryParams = initial => {
             case 'SET_ORG':
             case 'SET_CLUSTER':
             case 'SET_TEMPLATE':
+            case 'SET_AUTOMATION_STATUS':
```
jlmitch5, line 49: Somewhat out of scope of this pr (though I think it'd be fine to clean it up while we're in here too, up to you).  I wonder if we should just have a SET_VALUE action that is the default reducer action and delete all these that are just doing { ...state, ...value }.  Only ones that do something more complex to the state get a custom action.

src/Utilities/constants.js
```diff
@@ -77,6 +77,16 @@ export const clusters = {
     }
 };
 
+export const savingsPlanner = {
+    defaultParams: {
+        template_id: [],
```
jlmitch5, line 82: I think these are all only capable of single values, so I'm not sure why they default to arrays

src/Containers/SavingsPlanner/PlanCard.js
```diff
@@ -0,0 +1,147 @@
...
+            <CardFooter>
+                <Label>{category}</Label>
+            </CardFooter>
+        </Card>
+    );
+};
+
+PlanCard.propTypes = {
+    isSuccess: PropTypes.bool,
+    templates: PropTypes.array
```
jlmitch5, line 144: is there a reason why these specific props are getting propTypes but the others aren't?  If we can add them all aa well as add default props for them, I think that would be cool

src/Containers/SavingsPlanner/PlanCard.js
```diff
@@ -0,0 +1,147 @@
...
+            template_id: [templateId]
+        };
+        const search = stringify(initialQueryParams, { arrayFormat: 'bracket' });
+        history.push({
+            pathname: jobExplorer,
+            search
+        });
+    }
+
+    const showTemplate = template => {
```
jlmitch5, line 65: can we rename this to renderTemplateLink or something like that?  Show to implies it's a show hide type thing, but I think really, it's just a sub-component to pull the rendering out of the main render

jlmitch5 (changes requested):
Code is looking really good so far!  I like the separation of concerns with PlanCard and the SavingsPlanner component.  Just a few little things I noticed here and there.

src/Utilities/useQueryParams.js
brumik: I would keep this more verbose. Makes the logic more readable and explicit, not to mention how easy is to update logic for one of the values.

src/Containers/SavingsPlanner/PlanCard.js
```diff
@@ -0,0 +1,147 @@
...
+        };
+        if (isSuccess) {
+            return (
+                <a onClick={() => redirectToJobExplorer(template.id)}>{template.name}</a>
+            );
+        }
+    }
+
+    const kebabDropDownItems = [
+        <React.Fragment key={id}>
```
brumik, line 77: <3 for using `React.Fragment`

src/Containers/SavingsPlanner/PlanCard.js
brumik: We should add all props if we don't want to get some kicking from react :D also if the prop is needed we should add `.isRequired` to the end, and for all other props we should define default values while destructuring

src/Containers/SavingsPlanner/SavingsPlanner.js
```diff
@@ -0,0 +1,164 @@
...
+const perPageOptions = [
+    { title: '5', value: 5 },
+    { title: '10', value: 10 },
+    { title: '15', value: 15 },
+    { title: '20', value: 20 },
+    { title: '25', value: 25 }
+];
+
+const qp = {
+    limit: 5,
```
brumik, line 36: Just curious: Why are the limit and sort options here and not in the `constants.savingsPlanner.defaultParams`?

src/Containers/SavingsPlanner/SavingsPlanner.js
```diff
@@ -0,0 +1,164 @@
...
+    const [ currPage, setCurrPage ] = useState(1);
+
+    const combinedOptions = {
+        ...options.data,
+        name: [{ key: 'name', value: null }]
+    };
+
+    const returnOffsetVal = page => (page - 1) * queryParams.limit;
+
+    const handleSetPage = page => {
```
brumik, line 72: **Outside PR:** 
I start to see a pattern in pagination, maybe we can create our custom pagination component which would wrap the `Pagination` component from PF with our custom functions for handling the API params change.

src/Containers/SavingsPlanner/SavingsPlanner.js
```diff
@@ -0,0 +1,164 @@
...
+                    <ApiErrorState message={error.error} />
+                </Main>
+            )}
+            {isLoading && (
+                <Main style={{ height: '100vh' }}>
+                    <LoadingState />
+                </Main>
+            )}
+            {isSuccess && (
+                <Main style={{ height: '100vh' }}>
```
brumik, line 132: The `Main` can go outside the if statements since all the possible results from API are covered.

src/Containers/SavingsPlanner/SavingsPlanner.js
```diff
@@ -0,0 +1,164 @@
...
+                }}
+                onSetPage={(_event, pageNumber) => {
+                    handleSetPage(pageNumber);
+                }}
+                isSticky
+            />
+        </React.Fragment>
+    );
+};
+
```
brumik, line 163: Missing `PropTypes`.

src/Utilities/constants.js
brumik: No, they support filters by multiple values. (Well I could not check for the savings planner endpoint, but in every other place it is an array)

brumik (changes requested):
This is great and really extensive work (took me more than an hour to review :dancers:). I had lot's of improvement comments.

**The blockers:**
* Forgotten `eslint-disable` (and probably after removing some linting errors)
* Missing `propTypes` from components.

**Future PR Suggestions:**
* ToolbarFilterItem breakup
* Pagination wrapper component

src/Components/Toolbar/CustomFormElements/ToolbarFilterItem.js
```diff
@@ -51,6 +60,23 @@ const ToolbarFilterItem = ({
         }
     };
 
+    const onFilter = event => {
+        if (!event) {
```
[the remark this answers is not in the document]
jlmitch5: Looks like @Hyperkid123 also made a change to this in the chrome 2 pr...so something is up about onFilter.  https://github.com/RedHatInsights/tower-analytics-frontend/pull/464/commits/b519aefc8ccf55f25f79d3582bec5b6ce1e61e6d

src/Containers/SavingsPlanner/SavingsPlanner.js
jlmitch5: 👍  I was thinking a similar thing, it feels like there is some "generic" logic that could probably be excised out into a component...though I feel a pagination refactor would be good on its own and out of scope of this pr.

src/Components/Toolbar/CustomFormElements/ToolbarFilterItem.js
```diff
@@ -71,44 +97,68 @@ const ToolbarFilterItem = ({
...
+
+    const handleInputChange = e => {
+        setSearchVal(e);
+    };
+
+    const renderSearch = () => {
+        return (
+            <ToolbarItem>
+                <InputGroup>
+                    <TextInput name="textInput1" id="textInput1" type="search" aria-label="search input example" onChange={handleInputChange} />
```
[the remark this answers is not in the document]
kialam: 👍

src/Containers/SavingsPlanner/PlanCard.js
kialam: 👍

src/Api.js
```diff
@@ -9,13 +9,15 @@ const jobExplorerEndpoint = '/api/tower-analytics/v1/job_explorer/';
 const hostExplorerEndpoint = '/api/tower-analytics/v1/host_explorer/';
 const eventExplorerEndpoint = '/api/tower-analytics/v1/event_explorer/';
 const ROIEndpoint = '/api/tower-analytics/v1/roi_templates/';
+const PlansEndpoint = '/api/tower-analytics/v1/plans/';
```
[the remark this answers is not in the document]
kialam: 👍

src/Containers/SavingsPlanner/PlanCard.js
```diff
@@ -0,0 +1,147 @@
+/* eslint-disable */
```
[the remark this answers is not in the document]
kialam: Good call :)

src/Containers/SavingsPlanner/SavingsPlanner.js
kialam: I don't think this component has any proptypes.

src/Containers/SavingsPlanner/SavingsPlanner.js
kialam: Changing the code to move <Main> inside the if statements causes React to complain about a null react element being passed to the <Main> component, probably due to the async call to the API. I'll have to wrap it in the preFlightError if statement (I need to add this anyway) for it to work like you suggested.

src/Containers/SavingsPlanner/SavingsPlanner.js
brumik: Oh, my bad I mistook query params for props

src/Containers/SavingsPlanner/PlanCard.js
```diff
@@ -52,26 +49,19 @@ const PlanCard = (
...
-            return;
-        };
-        if (isSuccess) {
-            return (
-                <a onClick={() => redirectToJobExplorer(template.id)}>{template.name}</a>
-            );
-        }
-    }
+    const renderTemplateLink = template => {
+        return (template && isSuccess ? < a onClick = {() => redirectToJobExplorer(template.id)}> { template.name }</a> : null);
```
brumik, line 63: Is this our lint or? Space around the `=`, around the first `a` and an ?extra? space before `template.name`.

src/Containers/SavingsPlanner/PlanCard.js
```diff
@@ -141,7 +131,14 @@ const PlanCard = (
...
     isSuccess: PropTypes.bool,
-    templates: PropTypes.array
+    name: PropTypes.string,
+    id: PropTypes.number,
+    description: PropTypes.string,
+    frequency_period: PropTypes.string,
+    template_details: PropTypes.object,
+    modified: PropTypes.string,
+    category: PropTypes.string,
+    automation_status: PropTypes.object
```
brumik, line 141: If no default value provided in the component we should put `.isRequired` to the props without a default value.

jlmitch5 (approved):
looks good awesome job!

src/Containers/SavingsPlanner/PlanCard.js
```diff
@@ -130,8 +130,8 @@ const PlanCard = (
 };
 
 PlanCard.propTypes = {
-    isSuccess: PropTypes.bool,
-    name: PropTypes.string,
+    isSuccess: PropTypes.bool.isRequired,
+    name: PropTypes.string.isRequired,
     id: PropTypes.number,
```
brumik, line 135: If it is not required we should define default value for them in the destructuing

src/Containers/SavingsPlanner/PlanCard.js
brumik: @kialam

kialam: This was probably a linter auto format. Fixed now.

src/Components/Toolbar/CustomFormElements/ToolbarFilterItem.js
```diff
@@ -71,44 +97,68 @@ const ToolbarFilterItem = ({
...
+    if (categoryKey === 'name') {
+        return (
+            <ToolbarFilter
+                data-cy={categoryKey}
+                key={categoryKey}
+                showToolbarItem={isVisible}
+                chips={[]}
+                categoryName={options.name}
+                deleteChip={null}
+            >
```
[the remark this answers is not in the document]
kialam: 👍

kialam:
https://issues.redhat.com/browse/AA-53

PR #463, opened 2021-05-03
Fix to render either error state or empty state component.

h-kataria:
Render error state or empty state component based upon `preflightError` or `error`, no need to render both components at the same time

Fixes https://issues.redhat.com/browse/AA-480


src/Containers/Clusters/Clusters.js
```diff
@@ -185,14 +185,9 @@ const Clusters = () => {
                     setFilters={setFromToolbar}
                 />
             </PageHeader>
-            {preflightError && (
+            {(preflightError || error) && (
                 <Main>
```
brumik, line 189: If we are optimizing we can move the `<Main>` outside of the `if` statements:
```
<Main>
{if error && somecode}
{if not error && page }
</Main>
```
(Then also the React Fragment can be removed)

h-kataria: @brum `<Main>` can be moved out for sure, but i still can not remove `<React.Fragment>` part because we do need that to wrap page header

brumik: I meant the fragment on line 194 and the `Main` on line 195

brumik (approved):
Except for that one not important at all thing this is good to go

h-kataria:
@brumik please review unit tests. I tried to consolidate code using `<Main>` outside the code block but that did not work, so i am leaving that part as it is.

src/Containers/Clusters/Clusters.js
```diff
@@ -185,18 +191,13 @@ const Clusters = () => {
...
-            )}
-            {error && (
-                <Main>
-                    <ApiErrorState message={error.error} />
+                    { preflightError ? <EmptyState {...preflightError} /> : <ApiErrorState message={error.error} /> }
                 </Main>
             )}
             {!preflightError && !error && (
-                <>
+                <Router history={history}>
```
brumik, line 200: I am pretty sure this is going to the tests, not to the tested file. You should do something like this function with the react test renderer: https://github.com/RedHatInsights/tower-analytics-frontend/blob/devel/src/Utilities/tests/helpers.js#L18

(Also the bottom `</Router>` and `</>` are in swapped order)

src/Containers/Clusters/Clusters.test.js
```diff
@@ -1,11 +1,89 @@
...
+    });
+
+    afterEach(() => {
+        global.fetch.mockClear();
+        delete global.fetch;
+    });
+
+    test('has rendered preflight/authorization error component', async () => {
+        await act(async () => {
+            global.fetch = jest.fn().mockImplementation(() =>
```
brumik, line 36: This is only working when there is only one call of fetch at the time on the page. When I was talking about dropping fetch-mock I thought something that we mock the `exports from the Api.js` with the help of function mocks:
https://jestjs.io/docs/es6-class-mocks. This would ensure we can return different data on different params passed to these functions and we would have full control of each type of call. However these mocks should be in a different file probably, and the data fo the mocks also somewhere in a folder called `fixtures` or similar.

brumik (changes requested):

brumik (approved):
@h-kataria It is good from my side, if you review it and you see it fit, merge it. (and sorry for messing up probably one of your commits when rebased, if you are keen on it I can go back and fix them)

PR #464, opened 2021-05-04
Chrome 2 updade

Hyperkid123:
cc @brumik 

### Changes
- update build config to chrome 2 compatible settings
- update dependencies 
- fix minor bugs that were causing runtime errors and tests failures

Note that chrome 2 rendering is still disabled via the `skipChrome2` flag in webpack config. I will circle back and remove it once we update the main.yml config with the correct module definitions. I was told that there are still some nav changes pending so I will test and enable chrome 2 rendering once this will be done.

Also, I would like to use the build-in webpack proxy once those changes are in place. That will remove the necessity of using insights proxy for development.


config/dev.webpack.config.js
```diff
@@ -1,23 +1,23 @@
...
-    output: {
-        ...webpackConfig.output,
-        filename,
-        chunkFilename
+
+plugins.push(
+  require('@redhat-cloud-services/frontend-components-config/federated-modules')(
+    {
+      root: resolve(__dirname, '../'),
+      useFileHash: false,
```
kialam, line 15: @Hyperkid123 How is cache busting implemented on the CRC side? We had to add hashes to our filenames and I'm curious as to why this has been removed as part of this PR.

kialam (changes requested):

brumik:
It's working for me, only getting warnings about
* "event imported as d3 was not found in d3"
*  `WARNING in shared module @patternfly/react-tokens
No required version specified and unable to automatically determine one. Unable to find required version for "@patternfly/react-tokens" in description file (/home/brum/Documents/AA/tower-analytics-frontend/package.json). It need to be in dependencies, devDependencies or peerDependencies.`

But the build, refresh, links, and clicks and tests are all fine

brumik:
Issue for chrome2 update in JIRA @Hyperkid123: https://issues.redhat.com/browse/AA-494

config/dev.webpack.config.js
Hyperkid123: Hashes are added to js assets in the config.

Hyperkid123:
@brumik 
> WARNING in shared module @patternfly/react-tokens No required version specified and unable to automatically determine one. Unable to find required version for "@patternfly/react-tokens" in description file (/home/brum/Documents/AA/tower-analytics-frontend/package.json). It need to be in dependencies, devDependencies or peerDependencies.

Well that is because you are importing `react-tokens` in your code, but it's not listed as a direct dependency in your package.json. I will add it so the error goes away.

The d3 thing is caused by an invalid module definition. Nothing I can do about, we will just have to live with it.

Hyperkid123:
@brumik looking the cypress issue right now. Also, the chunk loading errors are unfortunately a bug when using chromium-based browsers, source maps, and the browser debugger. We will most likely disable source maps inc the chroma adn see if it helps.

The d3 webpack erros are caused by different API in newer versions, you can check the chacngle log here: https://github.com/d3/d3/blob/master/CHANGES.md

Hyperkid123:
@brumik @kialam can you trigger the cypress run? I was able to run it locally and most of the tests were passing. There were only some selector errors etc, which can be tuned later on. I just want to see if we can at least start the tests in CI env.

Hyperkid123:
Oh I see there is a different proxy config for the cypress than in `local-frontend.js`. I need to update that as well.

Hyperkid123:
@brumik @kialam I was able to run the cypress test even in CI, however, it's very non-deterministic. Sometimes a test passes, sometimes it does not. Mostly it's failing on proxy timeout or some runtime error tied to user identity. I will try to tinker around with it a bit more to make it more stable.

rvsia:
@Hyperkid123 the insights-proxy never worked for me on mac, so I am not the right candidate to test it. 😄

brumik:
@Hyperkid123 We disable cypress for the time being since we need a working environment, and we will need to fix it up later, as long as the only problem with the update will be cypress, this is good to go.

kialam (approved):
Everything seems to be working with the new proxy configs. I will note that developing with a local sidebar (e.g. serving `main.yaml` from cloud-services-config) doesn't work with new routes - they don't get redirected to `ansible/insights/new_route` @Hyperkid123 @ryelo . Is there a step I'm missing or a config file I need to look at to get a local side nav item to route correctly?

Hyperkid123:
@kialam  correct me if I am wrong @ryelo, but if there are any redirects happening via Akamai (which I believe there are: https://github.com/RedHatInsights/cloud-services-config/blob/prod-beta/akamai/data/prod/base_rules.json#L1296) it won't apply to local development.

But you can add new entries to the proxy config to serve assets from desired routes. So you will basically have duplicate entries on different routes (as you have for stable/beta envs).
```jsx
// current
routes[`/ansible/insights`] = { host: `https://localhost:${FRONTEND_PORT}` };
// duplicates
routes[`/ansible/whatever`] = { host: `https://localhost:${FRONTEND_PORT}` };
routes[`/ansible/whatever/foo`] = { host: `https://localhost:${FRONTEND_PORT}` };
...
```

brumik (approved):
Works for me :+1:

ryelo:
Yes @Hyperkid123 that's correct

PR #465, opened 2021-05-05
Tryin to update cypress for chrome 2 update

brumik:


Hyperkid123:
@brumik I was able to run the tests locally and about 50% of the tests have passed. Seems that there were some breaking changes in the new d3 version. I am downgrading it and will run the suite again.

PR #466, opened 2021-05-05
End of the moment

brumik:


src/Utilities/helpers.js
```diff
@@ -28,7 +28,10 @@ export function formatDateTime(dateTime) {
 }
 
 export function formatDate(date) {
```
jlmitch5, line 30: I feel like we should go ahead and get rid of all these date functions in this file and just use a consistent date format  method.

brumik: What do you mean?

*I am not sure you meant what I understood*:
I think it is nice to use an own wrapper function for datum. I mean now that we want to get rid of moment we just need to change 3 functions instead of ~15. But we could have only one for formatting datum with some params, that would be ideal I think.

jlmitch5: What I meant was--
In all places, date formatting is done the same way.  It looks to be in this PR, that would be importing the DateFormat component from the crc repo in all the places we want to format dates.

I'm not too picky about _what_ we do specifically, I just think that if it is all done one way (rather than in some places importing from the CRC shared components, some places utilizing the utility, some other places doing some other thing), it will be easier to maintain.

brumik: Ah, yeah, unfortunately the CRC is not covering all of our use cases so we will need to find something, also Ladas has mentioned a relative date calculation, which would mean a bit more advanced date processing, so I am considering https://date-fns.org/

kialam:
https://media.giphy.com/media/xT1XGLm7CJknNZKVS8/giphy.gif

brumik:
We are taking a different approach. RIP RIP MOMENT....

PR #470, opened 2021-05-12
Post chrome 2 script update

brumik:
I updated the scripts for starting tmux... and then I did a little more :/

I didn't want to add anything to the chrome 2 update repo. Here is what I did here:
* Added the `main.yml` to the repo for local full-stack development and set the script to run the webserver from it
* Updated the `Makefile` to run tmux with the new updates, also made the command shorter (sorry @jlmitch5 but yours was an absolute devil to type out or autocomplete), now you can run the nice tmux widnow with `make start`
* Removed the unused spandx configs from the `config` folder
* Removed the unused `package.json` commands for cypress (probably won't need it anymore even after reenabling cypress)
* Removed the unused `.eslintignore`
* Updated the `Dockerfile` to use node 14 --> I am still not sure if we need it still or not, and where it is used if we need it @benthomasson 

I would like to store the `main.yml` in the repo for development:
**Pros:**
1. No need for cloning another repo
2. When working together on a new page, no need for everybody to individually edit the file
3. Easier to find it from an `npm run` command as it has a constant place not depending on the repo hierarchy

**Cons:**
1. We need to reflect changes happening in the `cloud-services-config` repo manually for local full-stack-development (but we were doing it on an individual basis until now too I think)


jlmitch5 (changes requested):
This pr is great.  I really like the tweaks you've made to a lot of these things.  I think the experience of using them as well as maintaining them is improved in many ways.  Specifically:

> name change

great!

> removing unused spandx configs

great!

> pulling main.yml into this repo (instead of manually editing the other one

great!

---

If we could add a kill line to the beginning of the start command, that would also be cool (just need to figure out a way that it doesn't error out the rest of the command if there's nothing to kill):

`tmux kill-session -t aa`

The one point I don't quite understand is this:

> We need to reflect changes happening in the cloud-services-config repo manually for local full-stack-development (but we were doing it on an individual basis until now too I think)

I don't really understand what this means or why we need to do this manually.  The goal with that tmux script was to be able to have _all_ the fully local things running in one command, so that you can run them all with a single command.  This includes the crc config repo.

The whole point of cloning (and running the server) on that repo was to have a manually configured nav.

If we no longer need to run that (it seems we are somehow including the local nav as part of this repo), great and we can just remove the info from the onboarding doc about pulling down that repo.

If that's not the case, I think we need to tweak this PR (specifically the tmux command) back to starting this repo, as well as any sort of proxy config needed to expect that.

brumik:
> I don't really understand what this means or why we need to do this manually. The goal with that tmux script was to be able to have _all_ the fully local things running in one command, so that you can run them all with a single command. This includes the crc config repo.
> 
> The whole point of cloning (and running the server) on that repo was to have a manually configured nav.
> 
> If we no longer need to run that (it seems we are somehow including the local nav as part of this repo), great and we can just remove the info from the onboarding doc about pulling down that repo.
> 
> If that's not the case, I think we need to tweak this PR (specifically the tmux command) back to starting this repo, as well as any sort of proxy config needed to expect that.

Yes, we including the main.yml in the project, but it means that if they change the config for AAA in the repo, we need to update our main.yml too (however I think usually it works like we already work with the changes that they are pushing - new navbar items) so yes, basically it makes it easier at virtually no cost.

> If we could add a kill line to the beginning of the start command, that would also be cool (just need to figure out a way that it doesn't error out the rest of the command if there's nothing to kill):


I am not sure what you want, I just updated the script but didn't see this kill command. What should it do?

jlmitch5:
Update: I pulled this down.  It all works great, but the side nav is not changing when I update the main.yml file you've added in aa when accessing the beta urls

brumik:
@jlmitch5 are you using the command `npm run proxy:fullstack` or `make start` to run it? If yes the problem will be probably with the mac and I will try to generalize the command for more board platform usage, since it is working in Ubuntu terminal.

Also make sure you have installed http-server (https://www.npmjs.com/package/http-server) which is serving the main.yml on the port 8889

The command which should run the service for main.yml is:
`"proxy:fullstack": "SPANDX_CONFIG=./profiles/local-frontend-and-api.js bash ../insights-proxy/scripts/run.sh & npx http-server ./config -p 8889",`

jlmitch5:
I was misreading...the crc config is still part of the command.  This looks great!  Just added a commit to try to kill the aa session if things somehow shutdown weird and it is still hanging around.  Approving and merging

jlmitch5 (approved):

PR #471, opened 2021-05-12
Fixing org page - AA-481

brumik:
* Fixes the issue with toggling different chart then is next to the legend.
* Fixes displaying empty string in the legend instead of the "No organization".


jlmitch5 (approved):
great!

h-kataria (approved):
verified in UI

PR #473, opened 2021-05-14
Broke up the tolbar filter item into smaller parts

brumik:
* Restructured the toolbar how it works.
* Fixed error with the `Search by name` field and chip adding/deletion.
* Fixed misaligned checkbox - this was introduced when limiting the SelectOption max-width probably. Disliking custom styles for PF components :(

* Future work: Working on a Chip wrapper for all the possible components for enabling and disabling chips for all kind of files from the `constants.js` file


src/Components/Toolbar/Groups/FiltersCategoriesGroup.js
```diff
@@ -2,8 +2,8 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { ToolbarGroup } from '@patternfly/react-core';
 
-import CategoryDropdown from '../CustomFormElements/CategoryDropdown';
-import ToolbarFilterItem from '../CustomFormElements/ToolbarFilterItem';
+import CategoryDropdown from '../ToolbarItems/CategoryDropdown';
+import ToolbarItem from '../ToolbarItems/';
```
jlmitch5, line 6: I think renaming the folder to ToolbarInput (non-plural), as well as naming the index.js component ToolbarInput will help to make things a little clearer.

Specifically, ToolbarItem is a concept in core patternfly that I think we might be overloading here.  I think basically this "ToolbarInput" component switches the input type based on the options it is passed.

This is more preference and I'm fine either way, but it could be potentially be clearer if the ToolbarInput folder is nested inside Groups, where it is mean to be used.

src/Components/Toolbar/Groups/FiltersCategoriesGroup.js
```diff
@@ -2,8 +2,8 @@ import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { ToolbarGroup } from '@patternfly/react-core';
 
-import CategoryDropdown from '../CustomFormElements/CategoryDropdown';
-import ToolbarFilterItem from '../CustomFormElements/ToolbarFilterItem';
+import CategoryDropdown from '../ToolbarItems/CategoryDropdown';
```
jlmitch5, line 5: CategoryDropdown feels to me as if it is a part of the FilterCategoriesGroup component, rather than a toolbar item.  

The components are tightly and exclusively coupled, and both are < 50 LOC, so I'd recommend de-componetizing the CategoryDropdown (or moving the component definition) and inlining it in this file.  I think I'd prefer it not be exported (as it feels like something internal), but it's cool if it needs to to aid in making the existing tests work.

src/Components/Toolbar/ToolbarItems/index.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import CustomSelect from './CustomSelect';
+import InputDate from './InputDate';
+import InputText from './InputText';
+import { optionsForCategories } from '../constants';
+
+const components = {
+  select: CustomSelect,
```
jlmitch5, line 9: Can we rename these to ToolbarSelectInput, ToolbarDateInput, ToolbarTextInput or some other consistent pattern.  I think it'd be helpful to call out they are specifically for use with ToolbarItem and not for sharing outside.

jlmitch5 (changes requested):
This is looking good.  I'm wondering what you think about doing some renaming and moving some things around (see in-line comments).  I think this might help reduce some indirection, make it clearer which pieces are meant to be used where, and also just generally make things a little easier to understand.

src/Components/Toolbar/ToolbarItems/InputText.js
```diff
@@ -0,0 +1,66 @@
...
+
+const InputText = ({ categoryKey, isVisible = true, value = '', setValue }) => {
+  const [searchVal, setSearchVal] = useState(value);
+  const options = optionsForCategories[categoryKey];
+
+  const onDelete = () => {
+    setValue('');
+  };
+
+  const handleChips = () => {
```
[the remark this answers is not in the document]
brumik: We have cases for the other input types where in some places we need the chips, in others not (sort, quick date), and since we have the boolean in some fields, I think it is more consistent having it evrywhere. I hope I can move the chip logic later into a HOC component.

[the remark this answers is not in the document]
brumik: Good one, I think I was planning it to handle more, but in this form, we can just inline it.

src/Components/Toolbar/Groups/FiltersCategoriesGroup.js
brumik: Good one, I think I was planning it to handle more, but in this form, we can just inline it.

src/Components/Toolbar/ToolbarItems/index.js
brumik: Thanks, naming stuff is the hardest :+1:

src/Components/Toolbar/Groups/FiltersCategoriesGroup.js
brumik: Yes, I can rename it and if they are not used anywhere else than groups I will move there as a subfolder

src/Components/Toolbar/Groups/FiltersCategoriesGroup.js
jlmitch5: 👍  ah yeah that makes sense.

src/Components/Toolbar/ToolbarItems/InputText.js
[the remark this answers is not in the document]
jlmitch5: > We have cases for the other input types where in some places we need the chips, in others not (sort, quick date)

Totally, my thought here is I was wondering if we could introspect if chips show up rather than have it be something stored in configuration.  That being said, it's not a big deal having it in the config, and I agree with you, for this one, I think having it explicitly everywhere is a good thing.

jlmitch5 (approved):
Nice!

PR #474, opened 2021-05-14
savings planner add flow

jlmitch5:
This PR adds the Savings Planner add flow.

Currently, the pr:
- Adds “Add plan” button to Savings Planner card view to get to add screen
- Adds breadcrumb link back to savings planner list
- Adds in wizard component with hash-linkable steps (the hash linking will be needed for edit view)
- Adds in grid-based form from PF
    - Some fields needed the ? Help popover link in the label of the field…I pulled in this component from awx.
- Adds in drag-orderable task adding list
    - Task add input allows both enter key-down and clicking “+” button to add
    - Clicking the “x” in the remove will remove the task from the list.  I talked with Taufique and we decided against any sort of confirmation dialog for this particular interaction.
- Adds in radio single-select template list
    - Note, this is a simple, non-paginated DataList currently.  Will need to figure out with @akarve on the backend about pagination, searching, etc. in a follow-up PR
- Make save POST things to the API.

This is also my attempt at utilizing the naming conventions we discussed in our Monday sync-up meeting.  Please let me know what y'all think about this...happy to tweak names and file locations!

These TODOs I will do in a follow-up PR:
- TODO: Add unit tests
- TODO: Any tweaks @kyleabenson wants to category/platform/cloud on details form screen.
- TODO: Disable save button when required fields aren’t entered, and have a tooltip of what needs to be added/where you need to go to do that.
- TODO: Show informative error message when API POST fails to add.
- TODO: Hide add button and do the redirection notice from the add view when user has read-only role (RBAC)

### DIFF
--- a/src/Components/Popover.js
+++ b/src/Components/Popover.js
@@ -1,0 +1,55 @@
+import React from 'react';
+import { node, string } from 'prop-types';
+import { Popover as PFPopover } from '@patternfly/react-core';
+import { HelpIcon } from '@patternfly/react-icons';
+import styled from 'styled-components';
+
+const PopoverButton = styled.button`
+  padding: var(--pf-global--spacer--xs);
+  margin: -(var(--pf-global--spacer--xs));
+  font-size: var(--pf-global--FontSize--sm);
+`;
+
+function Popover({ ariaLabel, content, header, id, maxWidth, ...rest }) {
+  if (!content) {
+    return null;
+  }
+
+  return (
+    <PFPopover
+      bodyContent={content}
+      headerContent={header}
+      hideOnOutsideClick
+      id={id}
+      maxWidth={maxWidth}
+      {...rest}
+    >
+      <PopoverButton
+        aria-label={ariaLabel ?? 'More information'}
+        aria-haspopup="true"
+        className="pf-c-form__group-label-help"
+        onClick={(e) => e.preventDefault()}
+        type="button"
+      >
+        <HelpIcon noVerticalAlign />
+      </PopoverButton>
+    </PFPopover>
+  );
+}
+
+Popover.propTypes = {
+  ariaLabel: string,
+  content: node,
+  header: node,
+  id: string,
+  maxWidth: string,
+};
+Popover.defaultProps = {
+  ariaLabel: null,
+  content: null,
+  header: null,
+  id: '',
+  maxWidth: '',
+};
+
+export default Popover;
--- a/src/Components/Toolbar/Toolbar.js
+++ b/src/Components/Toolbar/Toolbar.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import {
   Toolbar,
   ToolbarContent,
+  ToolbarGroup,
   Button,
   ToolbarItem,
 } from '@patternfly/react-core';
@@ -20,6 +21,7 @@ const FilterableToolbar = ({
   setFilters,
   pagination = null,
   hasSettings = false,
+  additionalControls = [],
 }) => {
   const [settingsExpanded, setSettingsExpanded] = useState(false);
   const { quick_date_range, sort_options, ...restCategories } = categories;
@@ -78,6 +80,13 @@ const FilterableToolbar = ({
               </Button>
             </ToolbarItem>
           )}
+          {additionalControls.length > 0 && (
+            <ToolbarGroup>
+              {additionalControls.map((control) => (
+                <ToolbarItem key={control.key}>{control}</ToolbarItem>
+              ))}
+            </ToolbarGroup>
+          )}
           {pagination && (
             <ToolbarItem
               variant="pagination"
@@ -106,6 +115,7 @@ FilterableToolbar.propTypes = {
   setFilters: PropTypes.func.isRequired,
   pagination: PropTypes.object,
   hasSettings: PropTypes.bool,
+  additionalControls: PropTypes.array,
 };
 
 export default FilterableToolbar;
--- a/src/Containers/SavingsPlanner/SavingsPlanner.js
+++ b/src/Containers/SavingsPlanner/SavingsPlanner.js
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react';
+import { useHistory, useLocation } from 'react-router-dom';
 
 import { preflightRequest, readPlanOptions, readPlans } from '../../Api';
 import FilterableToolbar from '../../Components/Toolbar/';
@@ -17,12 +18,17 @@ import {
   PageHeader,
   PageHeaderTitle,
 } from '@redhat-cloud-services/frontend-components/PageHeader';
-
 import NotAuthorized from '@redhat-cloud-services/frontend-components/NotAuthorized';
 
-import { Gallery, PaginationVariant } from '@patternfly/react-core';
+import { Button, Gallery, PaginationVariant } from '@patternfly/react-core';
+
+// TODO: update to fining this out from API RBAC
+const canAddPlan = true;
 
 const SavingsPlanner = () => {
+  const history = useHistory();
+  const { pathname } = useLocation();
+
   const { queryParams, setFromPagination, setFromToolbar } = useQueryParams(
     savingsPlanner.defaultParams
   );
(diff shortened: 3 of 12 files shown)


src/Containers/SavingsPlanner/Shared/Details.js
```diff
@@ -0,0 +1,178 @@
...
+            fieldId="name-field"
+          >
+            <TextInput
+              isRequired
+              placeholder="Example: Provision NGINX server"
+              type="text"
+              id="name-field"
+              name="name"
+              value={name}
+              onChange={(newName) => setName(newName)}
```
kialam, line 50: This would become `onChange={(newName) => { type: 'SET_PLAN_NAME', value: { newName } }}`

src/Containers/SavingsPlanner/Shared/Form.js
```diff
@@ -0,0 +1,140 @@
...
+    history.replace({
+      hash: newStep.id,
+    });
+  };
+
+  const onSave = () => {
+    setData(
+      createPlan({
+        params: {
+          name,
```
kialam, line 55: Here you could just pass in `planData`.

src/Containers/SavingsPlanner/Shared/Tasks.js
```diff
@@ -0,0 +1,180 @@
...
+    setLiveText('Dragging finsihed');
+    setTasks([...newItemOrder]);
+  };
+
+  const appendTask = () => {
+    setTasks([...tasks, taskToAdd]);
+    setTaskToAdd('');
+  };
+
+  const handleTextKeyDown = (e) => {
```
kialam, line 75: Nice

src/Containers/SavingsPlanner/Shared/Tasks.js
```diff
@@ -0,0 +1,180 @@
...
+    if (e.key && e.key === 'Enter') {
+      e.preventDefault();
+      appendTask();
+    }
+  };
+
+  const removeTask = (task) => {
```
kialam, line 82: Love the remove method - a very thorough and well-thought out UI for the tasks list!

src/Containers/SavingsPlanner/Shared/Templates.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
```
kialam, line 1: We are going to replace this in a follow up PR with a paginated data list of templates so works for now 👍

kialam (changes requested):
Overall this is looking great, @jlmitch5 ! I think creating a reducer method to help set/get plan data to POST to the API would be beneficial here, so that's my one main request. 

Fantastic job!

src/Components/Toolbar/Toolbar.js
```diff
@@ -78,6 +80,13 @@ const FilterableToolbar = ({
               </Button>
             </ToolbarItem>
           )}
+          {additionalControls.length > 0 && (
```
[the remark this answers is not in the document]
jlmitch5: yes.  additionalControls is a pattern I pulled from tower.  It is how things like Add, Delete, etc. buttons are added to the toolbar.  Since it has always seem to work well for providing a spot to put arbitrary buttons and things towards the end, I decided to introduce the pattern here.

My preference would be to do what is done there to be consistent, and if we run into things specific to this app, we can approach that way.  What do you think?

src/Containers/SavingsPlanner/SavingsPlanner.js
```diff
@@ -78,6 +84,24 @@ const SavingsPlanner = () => {
           categories={combinedOptions}
           filters={queryParams}
           setFilters={setFromToolbar}
+          additionalControls={[
+            ...(canAddPlan
```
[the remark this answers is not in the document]
jlmitch5: I can pull this up above as a constant.  I inlined it because that is how it is done in tower.

I would probably make it a function, just an array constant.

src/Components/Popover.js
```diff
@@ -0,0 +1,55 @@
...
+import styled from 'styled-components';
+
+const PopoverButton = styled.button`
+  padding: var(--pf-global--spacer--xs);
+  margin: -(var(--pf-global--spacer--xs));
+  font-size: var(--pf-global--FontSize--sm);
+`;
+
+function Popover({ ariaLabel, content, header, id, maxWidth, ...rest }) {
+  if (!content) {
```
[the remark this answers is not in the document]
jlmitch5: This is another one I pulled straight from Tower...I honestly didn't look very closely, just that pulling in the component worked for what I wanted out of the box.

I can see if there's a way to clean up.

jlmitch5:
Thanks for the great review!

I'm not as familiar with the custom hook/reducer pattern, but now that you've explained it and I've done a bit of research, I totally agree with you, it seems like it would be a great fit for this!  It felt weird passing in so many props to the wizard steps, but I knew I wanted all of the plan data to roll up to the root form component.  This seems like a way that would allow me to do that but keep the API of those step components much tighter.  Thanks for the suggestion!

Popover and Toolbar additionalControls I'll discuss with you offline to see how/if we should tackle those differently.

src/Utilities/useForm.js
```diff
@@ -0,0 +1,25 @@
+import { useReducer } from 'react';
+
+const useForm = (initial) => {
+  const formReducer = (state, { type, value }) => {
+    switch (type) {
+      /* v1 api reducers */
+      case 'SET_FIELD':
+        return { ...state, ...value };
+      default:
+        throw new Error();
```
brumik, line 10: If this gets here it breaks the application, can we add at least a string saying "useFormReducer `action.type` was not found"?

src/Utilities/useForm.js
```diff
@@ -0,0 +1,25 @@
+import { useReducer } from 'react';
+
+const useForm = (initial) => {
+  const formReducer = (state, { type, value }) => {
+    switch (type) {
+      /* v1 api reducers */
+      case 'SET_FIELD':
```
brumik, line 7: I definitely do not prefer this style. If we use a reducer we should create an action for each possible action we can do, even if it is the same handler. As I see it before for example "SET_TEMPLATE_ID", "SET_SOME_OTHER", "DELETE_TEMPLATE_ID", "RESET_FORM" etc...

These string constants could live as exportend constants, so we can just export the `formData`, `dispatch` and then we can pull the action name constants fomr other firles. This makes this reducer TypeScript ready.

src/Utilities/useForm.js
```diff
@@ -0,0 +1,25 @@
...
+        return { ...state, ...value };
+      default:
+        throw new Error();
+    }
+  };
+
+  const [formData, dispatch] = useReducer(formReducer, {
+    ...initial,
```
brumik, line 15: If there is a set of values, we should initialize all of the attributes here, even if they are not passed in the `initial` value.

src/Utilities/useForm.js
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  return {
+    formData,
+    setField: (field, value) =>
+      dispatch({ type: 'SET_FIELD', value: { [field]: value } }),
+  };
```
brumik, line 22: we definitely want to export dispatch too

src/Containers/SavingsPlanner/Shared/Form.js
```diff
@@ -49,22 +51,21 @@ const Form = ({ title, options }) => {
   };
 
   const onSave = () => {
+    if (formData.template_id === -2) {
+      delete formData.template_id;
```
brumik, line 55: When using reducers, this is not the best way (I am surprised it works) to do this. If you don't want to modify the data permanently then create a copy of the form data (`{ ...formData }`), if the `formData` should change permanently then create a dispatch action for removing the template id and updating the tasks. Objects in react should be immutable.

brumik (changes requested):
Kia did the hard work in the previous PR review and I could only agree with her (this is a really good job here). 

I just made some advice on the reducer best practices (last commit). These things will make the code more future proof and bring it closer to the TS if we want to migrate in the future maybe.

I would definitely like to see the reducer exporting only a `dispatch`, `formData` and lots of actions constants here, so it could be used like:

`dispatch({ type: ACTION_CONST_IPORTED_FROM_THE_HOOK_FILE, value })`

src/Containers/SavingsPlanner/Shared/Form.js
```diff
@@ -0,0 +1,127 @@
...
+      hash: newStep.id,
+    });
+  };
+
+  const onSave = () => {
+    if (formData.template_id === -2) {
+      delete formData.template_id;
+    }
+
+    formData.tasks = formData.tasks.map((task, index) => ({
```
kialam, line 58: Can the hook create the tasks array instead of having the logic here?

brumik (approved):
My suggested changes are non blocking and can be done in a later in a followup PR

kialam:
These changes can be made in a separate PR.

kialam (approved):

jlmitch5:
Merging now, will follow up from feedback in the next pr!

PR #476, opened 2021-05-18
Pagination refactor

kialam:


src/Components/Pagination/Pagination.test.js
```diff
@@ -0,0 +1,58 @@
...
+      handleSetOffset={bar}
+      handleSetCurrPage={baz}
+      currPage={1}
+    />
+  );
+  fireEvent.click(screen.getByRole('button', { name: 'Items per page' }));
+  fireEvent.click(screen.getByText('10 per page'));
+
+  expect(foo).toHaveBeenCalled();
+});
```
jlmitch5, line 58: these test feel so clean, loviing this lib!

src/Components/Pagination/Pagination.js
```diff
@@ -0,0 +1,60 @@
+import PropTypes from 'prop-types';
+import { perPageOptions } from './constants';
```
jlmitch5, line 2: I would inline this in this file...only being used here now.  If we end up using in more than 1 place, I think that would be the time to move this out to an exported constant

src/Components/Pagination/Pagination.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+Pagination.propTypes = {
+  limit: PropTypes.number.isRequired,
+  count: PropTypes.number.isRequired,
+  currPage: PropTypes.number.isRequired,
+  handleSetCurrPage: PropTypes.func.isRequired,
+  handleSetLimit: PropTypes.func.isRequired,
+  handleSetOffset: PropTypes.func.isRequired,
+  variant: PropTypes.string,
```
jlmitch5, line 57: very clean api!

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -143,18 +116,13 @@ const JobExplorer = ({ location: { search }, history }) => {
...
                 pagination={
                   <Pagination
-                    itemCount={meta && meta.count ? meta.count : 0}
-                    widgetId="pagination-options-menu-top"
-                    perPageOptions={perPageOptions}
-                    perPage={queryParams.limit}
-                    page={currPage}
+                    count={meta?.count ? meta.count : 0}
+                    limit={queryParams.limit}
                     variant={PaginationVariant.top}
```
jlmitch5, line 121: Do we need the top ones...it seems like it defaults to that

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -143,18 +116,13 @@ const JobExplorer = ({ location: { search }, history }) => {
                 setFilters={setFromToolbar}
                 pagination={
                   <Pagination
-                    itemCount={meta && meta.count ? meta.count : 0}
-                    widgetId="pagination-options-menu-top"
-                    perPageOptions={perPageOptions}
-                    perPage={queryParams.limit}
-                    page={currPage}
+                    count={meta?.count ? meta.count : 0}
+                    limit={queryParams.limit}
```
jlmitch5, line 120: this feels quite a bit cleaner!

src/Components/Pagination/Pagination.js
```diff
@@ -0,0 +1,60 @@
+import PropTypes from 'prop-types';
+import { perPageOptions } from './constants';
+
+import { Pagination as PFPagination } from '@patternfly/react-core';
+
+const Pagination = ({
+  limit,
```
jlmitch5, line 7: could we default the limit and make it a non-required prop?  Not sure if that will tighten things or not, it may be always defined at the list level where this is used.

jlmitch5 (commented):
This is fantastic!  Just a few nitpicky things

src/Components/Pagination/Pagination.js
kialam: Limit is tied to the api query param, e.g. `limit=5` so not sure if we should default it in favor of matching what we already query to display.

src/Containers/JobExplorer/JobExplorer.js
kialam: 👍

src/Components/Pagination/Pagination.js
kialam: Sounds good

src/Components/Pagination/Pagination.js
jlmitch5: yeah i guess since it is always queried we can expect it is always passed and a required prop.  let's leave it like you have it then 👍

jlmitch5 (approved):
looks great!

PR #478, opened 2021-05-20
Making pagination simpler

brumik:
Basically:
Calculate current page from the offset and limit -> one source of truth
Add an unified method to the queryParams so it is easier to pull in in pages
Showcase typescript


src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -132,13 +129,11 @@ const JobExplorer = ({ location: { search }, history }) => {
               {isSuccess && data.length <= 0 && <NoResults />}
               {isSuccess && data.length > 0 && <JobExplorerList jobs={data} />}
               <Pagination
-                count={meta?.count ? meta.count : 0}
-                limit={queryParams.limit}
+                count={meta?.count}
```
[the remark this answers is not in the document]
brumik: It is defaulted to 0 in the Pagination props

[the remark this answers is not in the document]
brumik: https://github.com/RedHatInsights/tower-analytics-frontend/pull/478/files#diff-6e5301f2c45b5471c2e369ebe0281c9babf1fb8d5ac1dc298bad3b29889ea557R7

src/Components/Pagination/Pagination.test.js
```diff
@@ -1,58 +1,45 @@
...
+    limit: 5,
+  };
+  const count = 15;
+  const setPagination = jest.fn();
+
+  beforeEach(() => {
+    setPagination.mockReset();
+  });
+
+  it('can render the pagination component', async () => {
```
kialam, line 16: Do you need `async` render here?

src/Components/Pagination/Pagination.test.js
```diff
@@ -1,58 +1,45 @@
...
+      <Pagination count={count} params={params} setPagination={setPagination} />
+    );
+  });
+
+  it('can render the pagination component with default params', async () => {
+    render(<Pagination params={params} setPagination={setPagination} />);
+  });
+
+  it('calls the pagination functions', async () => {
+    const { getByRole } = render(
```
kialam, line 27: Any reason not to use `fireEvent.click(screen.getByRole('button', { name: 'Items per page' }));` instead of destructuring it out?

src/Components/Pagination/Pagination.test.js
```diff
@@ -1,58 +1,45 @@
...
+      <Pagination count={count} params={params} setPagination={setPagination} />
+    );
+    fireEvent.click(getByRole('button', { name: 'Go to next page' }));
+
+    expect(setPagination).toHaveBeenCalledWith(5);
```
kialam, line 32: You can tie this more with the coupling we've created with useQueryParams to by passing `params.limit` instead of hardcoding `5`.

src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -116,12 +116,12 @@ const JobExplorer = ({ location: { search }, history }) => {
...
                   <Pagination
-                    count={meta?.count ? meta.count : 0}
-                    limit={queryParams.limit}
-                    handleSetLimit={setLimit}
-                    handleSetOffset={setOffset}
-                    handleSetCurrPage={setCurrPage}
-                    currPage={currPage}
+                    count={meta?.count}
+                    params={{
+                      limit: queryParams.limit,
```
kialam, line 121: do we need to use type coercion here? I saw you had `parseInt` in a draft PR. Incidentally, `+queryParams.limit` also works.

src/Containers/SavingsPlanner/SavingsPlanner.js
```diff
@@ -80,12 +69,12 @@ const SavingsPlanner = () => {
           setFilters={setFromToolbar}
           pagination={
             <Pagination
-              count={meta?.total_count ? meta.total_count : 0}
-              limit={queryParams.limit}
-              handleSetLimit={setLimit}
-              handleSetOffset={setOffset}
-              handleSetCurrPage={setCurrPage}
-              currPage={currPage}
+              count={meta?.count}
```
kialam, line 72: should be `total_count` here :)

src/Containers/SavingsPlanner/SavingsPlanner.js
```diff
@@ -121,13 +110,13 @@ const SavingsPlanner = () => {
         </Main>
       )}
       <Pagination
-        count={meta?.total_count ? meta.total_count : 0}
-        limit={queryParams.limit}
+        count={meta?.count}
```
kialam, line 113: Same as above

src/Utilities/constants.js
```diff
@@ -84,6 +86,11 @@ export const savingsPlanner = {
     category: [],
     frequency_period: [],
     name: '',
+    limit: 5,
+    offset: 0,
+    sort_options: 'modified',
+    sort_order: 'desc',
+
```
kialam, line 93: remove empty line

src/Utilities/useQueryParams.js
```diff
@@ -42,7 +35,14 @@ export const useQueryParams = (initial) => {
         return { ...state, ...value };
 
       /* v1 api reducers */
+      case 'SET_LIMIT':
+        return isNaN(value)
+          ? { ...state, limit: 5 } // Defaults back to 5
```
kialam, line 40: So one alternative is that instead of returning nothing if value passed is string (if submitting data via a form, the value can be a string sometimes which is annoying), we do type coercion here: `{ ...state, limit: +value }`

kialam (approved):
I like that the `currentPage` logic is no longer the responsibility of the parent component and is neatly tied to the Pagination component so good job cleaning that up 👍 

I had a few questions about tests (do we need async, why are we not importing screen and calling screen.getByRole, etc.), and some type conversions wrt to the `limit` and `offset` params.

The only thing that needs immediate fixing is that pagination is broken on the Savings Planner page.

kialam (changes requested):
Sorry, mean to request changes instead of approving it.

src/Components/Pagination/Pagination.test.js
brumik: Good catch, I didn't see this when updating the tests

src/Components/Pagination/Pagination.test.js
brumik: I received a tip from other ppl using it: it is supposed to be a good practice, since we know exactly to which container the given destructured method refers, instead of depending on a global state (I am guessing you can render to the global state multiple components, but not sure). However, I am not keen on using it if you deem the `screen` readable.

src/Containers/JobExplorer/JobExplorer.js
brumik: The `parseInt` lived in the reducer initially, so I guessed we could just parse as int from the URL too (after tracking down the errors I realized this is the only place we are getting the limit as string and not int). 

The new object creation is needed since I wanted the runtime checking with PropTypes, and also the static type check with TypeScript and there is a weird behaviour when `PropTypes.shape` is giving an error when using with typescript

src/Utilities/useQueryParams.js
brumik: I am sorry, I am not sure what you mean here. If we don't get something that is not a number or cannot be converted to a number I default it back to 5, otherwise we are converting to a number with `parseInt` on the next line (I checked, parseInt does nothing if a number is passed to it.)

src/Containers/JobExplorer/JobExplorer.js
kialam: Ah, if the limit as string is coming from the URL, then we can pass a `{parseNumbers: true}` option from the `query-strings` library to solve this problem as well :) https://github.com/sindresorhus/query-string#parsenumbers. We do something similar for booleans from the URL too.

brumik: Yes, I added it few lines above. Then I am not entirely sure what you meant here. We know that the initial params for `query params limit` are int, all the updates because the reducer will be int, so we don't need type coercion anywhere else.

src/Utilities/useQueryParams.js
kialam: Ok, sorry I misread the code. This logic works for me. 👍

src/Containers/JobExplorer/JobExplorer.js
kialam: Ok, good call. You can ignore my type conversion comments it seems like we are doing as much as possible to ensure limit is a number and handling any cases where it isn't by converting it to a number so 👍

src/Components/Pagination/Pagination.test.js
```diff
@@ -1,58 +1,45 @@
...
+describe('Components/Pagination', () => {
+  const params = {
+    offset: 0,
+    limit: 5,
+  };
+  const count = 15;
+  const setPagination = jest.fn();
+
+  beforeEach(() => {
+    setPagination.mockReset();
```
kialam, line 13: [You no longer need to do this.](https://testing-library.com/docs/react-testing-library/api/#cleanup)

src/Components/Pagination/Pagination.test.js
kialam: It is actually now considered [bad practice](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen) according to Kent C. Dodds. Since we are only rendering the Pagination component and not any wrapper associated with it, I also think it's a bit redundant to do it this way.

src/Components/Pagination/Pagination.test.js
brumik: I think the `cleanup` function from the react testing library is actually just unmounting the rendered component. This line resets the mocked function which is Jest's territory. I am not aware that Jest would do the function cleanup for you.

https://jestjs.io/docs/mock-function-api#mockfnmockreset

src/Components/Pagination/Pagination.test.js
brumik: Haha, okay, now I am making common mistakes :D. Definitely good to know this ^. We probably should add eslint to our testing library too, as I see the recommendations

src/Components/Pagination/Pagination.test.js
kialam: Right, but if react testing library is already unmounting the whole component, do we need an extra call to reset a mocked function? Seems unnecessary.

brumik: Yes, otherwise it retains information between tests about the function call history (number of calls, call parameters, or the modified return value). I think we would not need it if we would create locally the mock function in each test case, but mocking a function is much more expensive then resetting after each test the data inside the mock.

We are not using any of the functionality in our tests which would be broken if we would retain the data in the mock function between tests (eg. checking the number of calls), but I think this is a small price for less inconvenience in the future if we wish to expand on tests. (Also as I know myself I will copy this pattern multiple times when creating new tests, and then I would forget about resetting the mock, causing headaches for myself).

brumik: The mocked function has no ties to the rendered component in the react testing library to my knowedlege. Therefore unmounting a component would not mean resetting a function which is passed to this component as a prop. 

I will try to create later a counter example as i cannot find any article which would state one or another viewpoint on this topic (or you can try to create one if you get to it first).

brumik: I updated the tests by testing the number of calls of the mocked `setPagination` function. If you comment out [this line](https://github.com/RedHatInsights/tower-analytics-frontend/pull/478/commits/2b5be689fcd5d6abcdea33eacd62e0fc2625adcb#diff-568b6f2120c6c703706d8f00ffd324aadc652826b3635b1f57611ec751f4a112R13) you can see the tests will not pass as the last test with `toHaveBeenCalledTimes` will receive `2` instead of `1` making the test fail, and showing off that the mock function retains history between the tests.

kialam: Ok, good observation and thanks for the example. I'm fine with keeping the mockReset and it sounds like a good thing to have before all test cases. 👍

kialam (approved):

PR #486, opened 2021-06-01
Adjust aria label to either "template" or "workflow".

kialam:
https://issues.redhat.com/browse/AA-536

### DIFF
--- a/src/Components/TemplatesList.js
+++ b/src/Components/TemplatesList.js
@@ -35,10 +35,14 @@ const TemplatesList = ({ templates, isLoading, qp, title, jobType }) => {
   const [selectedId, setSelectedId] = useState(null);
   const [isModalOpen, setIsModalOpen] = useState(false);
 
+  const displayJobType = type => {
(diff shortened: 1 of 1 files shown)


jlmitch5 (approved):
looks great!

src/Components/TemplatesList.js
```diff
@@ -35,18 +35,22 @@ const TemplatesList = ({ templates, isLoading, qp, title, jobType }) => {
   const [selectedId, setSelectedId] = useState(null);
   const [isModalOpen, setIsModalOpen] = useState(false);
 
+  const displayJobType = type => {
+    return (type === 'job' ? 'templates' : 'workflows');
```
brumik, line 39: This could be done with object too (but this simple use case probably it doesn't matter which one we use):
```js
const displayJobType = { job: 'templates', xxx: 'workflows' };

displayJobType[jobType]
```

brumik (approved):
LGTM

PR #488, opened 2021-06-02
Added Delete button on List and Details view

h-kataria:
- This commit adds a delete button in toolbar if user has RBAC Access. Clicking on delete button will open up a confirmation modal with  All selected plan names displayed in modal. Upon confirmation, records will be deleted and list view will be refreshed to show a list of plans.
- Added Delete button on the details tab with same confirmation modal, upon confirmation, record will be deleted and user will be redirected back to refreshed list view screen

List screen
Screen Shot 2021-06-07 at 4 34 39 PM

Screen Shot 2021-06-07 at 4 34 55 PM

Details screen
Screen Shot 2021-06-07 at 4 35 09 PM

Screen Shot 2021-06-07 at 4 35 19 PM

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -143,6 +143,13 @@ export const createPlan = ({ params = {} }) => {
   }).then(handleResponse);
 };
 
+export const deletePlan = ({ params = {} }) => {
+  let url = new URL(`${planEndpoint}${params.id}`, window.location.origin);
+  return authenticatedFetch(url, {
+    method: 'DELETE'
+  }).then(handleResponse);
+};
+
 export const updatePlan = ({ id, params = {} }) => {
   let url = new URL(`${planEndpoint}${id}`, window.location.origin);
   return authenticatedFetch(url, {
--- a/src/Components/AlertModal/AlertModal.js
+++ b/src/Components/AlertModal/AlertModal.js
@@ -1,0 +1,84 @@
+import React from 'react';
+import { Modal, Title } from '@patternfly/react-core';
+import {
+  CheckCircleIcon,
+  ExclamationCircleIcon,
+  ExclamationTriangleIcon,
+  InfoCircleIcon,
+  TimesCircleIcon,
+} from '@patternfly/react-icons';
+import styled from 'styled-components';
+
+const Header = styled.div`
+  display: flex;
+  svg {
+    margin-right: 16px;
+  }
+`;
+
+function AlertModal({
+  isOpen = null,
+  title,
+  label,
+  variant,
+  children,
+  ...props
+}) {
+  const variantIcons = {
+    danger: (
+      <ExclamationCircleIcon
+        size="lg"
+        css="color: var(--pf-global--danger-color--100)"
+      />
+    ),
+    error: (
+      <TimesCircleIcon
+        size="lg"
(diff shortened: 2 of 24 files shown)


kialam:
@h-kataria I know this is still a WIP but wanted to just point out that unit tests should be with react-testing-library instead of enzyme.

h-kataria:
> @h-kataria I know this is still a WIP but wanted to just point out that unit tests should be with react-testing-library instead of enzyme.

@kialam that is why I still have this in WIP, I am working on converting them to use react-testing-library

src/Components/Toolbar/ToolbarDeleteButton.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+  test('should render button', () => {
+    render(
+      <ToolbarDeleteButton onDelete={() => {}} itemsToDelete={[]} />
+    );
+    expect(screen.getByRole('button')).toBeTruthy();
+  });
+
+  test('should open confirmation modal', async () => {
+    await act(async () => {
```
kialam, line 38: Do we need to wrap the render in both an `async` arrow function, as well as `act`? Can we just render the `ToolBarDeleteButton`, and then `await` the appearance of the Alert modal?

```javascript
render(<ToolbarDeleteButton />);
fireEvent.click(screen.getByRole('button'));
await screen.findByText('Alert modal');
expect(screen.getByLabelText('Alert modal').toBeInTheDocument());
```

https://testing-library.com/docs/dom-testing-library/api-async/#findby-queries

h-kataria: @kialam fixed

src/Api.js
```diff
@@ -143,6 +143,22 @@ export const createPlan = ({ params = {} }) => {
...
 
+export const deletePlan = ({ params = {} }) => {
+  const { limit, offset, sort_by } = params;
+  const paginationParams = {
+    limit,
+    offset,
+    sort_by,
+  };
+  const qs = stringify(paginationParams);
+  let url = new URL(planEndpoint, window.location.origin);
```
jlmitch5, line 154: Does this work?  Trying to figue out if we can get rid of that string concatenation on 156.
```
let url = new URL(`${planEndpoint}${params.id}`, window.location.origin);
url.search = qs;
```

src/Components/AlertModal/AlertModal.js
```diff
@@ -0,0 +1,84 @@
...
+      variant="small"
+      title={title}
+      {...props}
+    >
+      {children}
+    </Modal>
+  );
+}
+
+export default AlertModal;
```
jlmitch5, line 84: nice, i remember making this component in awx back in the day heh

src/Utilities/getRelatedResourceDeleteDetails.js
```diff
@@ -0,0 +1,42 @@
+import {
+  readPlan
```
jlmitch5, line 2: can we pass this method as a param to this utility.  i'd prefer if we didn't couple plan stuff to this generic utility.

jlmitch5 (commented):
looking good, some small things

PR #490, opened 2021-06-04
Add unit tests to Shared/Details component.

kialam:


src/Containers/SavingsPlanner/Shared/Details.test.js
```diff
@@ -0,0 +1,207 @@
...
+    name: '',
+    category: 'system',
+    description: '',
+    manual_time: 240,
+    hosts: 1,
+    frequency_period: 'weekly',
+    tasks: [],
+    template_id: -2,
+  };
+  const mockOptioins = {
```
jlmitch5, line 17: spelling

jlmitch5 (approved):
awesome tests!  thank so much for doing this.  just one tiny spelling error and I believe this is ready for merge!

brumik:
Wohooo 🎉

One note however: i see you add lately to your test names that 'user can' prefix. I find it slightly unnecessary since the 'it' keyword makes it read without the 'user' prefix fine: "it can do this and that" opposed "it (user can do this and that)". (And since the test should be able to do it when testing it gives sense)

jlmitch5 (approved):

PR #492, opened 2021-06-07
Visual updates to the statistic page

brumik:
I will do a follow-up PR to clean up a little when the chart builder supports all these things I added for the chart here.

This PR contains style fixes mostly around the chart. Additionally removed all of the custom style elements and replaced them with corresponding PF elements. 

The PF version update allowed to use of the more advanced `List` from PF which supports icons.

Added strict to the typescript config so it actually checks types.

Result:
Screenshot from 2021-06-08 17-11-26


src/Containers/SavingsPlanner/TotalSavings.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+const TotalSavings: FunctionComponent<Props> = ({ value = 0, isMoney = true }) => (
+  <Card isPlain>
+    <CardTitle style={{ paddingBottom: '0' }}>Total savings</CardTitle>
+    <CardBody>
+      <Title
+        headingLevel="h3"
+        size="4xl"
+        style={{ color: 'var(--pf-global--success-color--200)' }}
```
jlmitch5, line 23: this changes based on isMoney see mockups

src/Containers/SavingsPlanner/StatisticsTab.tsx
```diff
@@ -85,30 +59,161 @@ interface Props {
...
+  /**
+   * Calculate the y offset for an axis from height and from the number of ticks
+   * and number of negative ticks. Has the top + bottom padding encoded as constant (120).
+   * Changing the margin will cause the function to calculate the offset wrogly.
+   * 
+   * @param ticks All the ticks for the chart, with the end ticks to be the edges of the domain
+   * @param height The height of the chart
+   * @returns Offset for the axis even if it has negative values.
+   */
+  const getXOffsetForAxis = (ticks: number[], height = 600): number => {
```
jlmitch5, line 213: i see from your comment these will be moved out to the chart renderer (which I think will be good and help reduce the complexity in this file).

src/Containers/SavingsPlanner/StatisticsTab.tsx
```diff
@@ -85,30 +59,161 @@ interface Props {
...
-  const statsData = years.map(year => ({
-    year,
-    total_costs: data.projections.monetary_stats.total_costs[year] * -1,
-    total_benefits: data.projections.monetary_stats.total_benefits[year],
-    cumulative_net_benefits: data.projections.monetary_stats.cumulative_net_benefits[year],
-    total_hours_spent_risk_adjusted: data.projections.time_stats.total_hours_spent_risk_adjusted[year] * -1,
-    total_hours_saved: data.projections.time_stats.total_hours_saved[year],
-    cumulative_time_net_benefits: data.projections.time_stats.cumulative_time_net_benefits[year]
+// TODO move this logic to the chart renderer
+const formatNumberAsK = (n: string | number): string => {
```
jlmitch5, line 63: typo (upper-case K)

src/Containers/SavingsPlanner/StatisticsTab.tsx
```diff
@@ -85,30 +59,161 @@ interface Props {
...
+   */
+  const getDomainFromTicks = (): [number, number] => {
+    const ticks = getTickValues();
+    return [ticks[0], ticks[ticks.length -1]];
+  }
+
+  /**
+   * Removes the edge ticks.
+   * 
+   * @param ticks The ticsk for the chart where the edge ticks are the domain for the chart.
```
jlmitch5, line 194: typo ticsk

jlmitch5 (changes requested):
This is looking great!  A few things:
- change the big hours number to blue (inline comment)
- looks like the top of the elapsed line is getting cut off, with these values:
Screen Shot 2021-06-09 at 3 32 34 PM

Screen Shot 2021-06-09 at 3 31 10 PM

It's also a bit strange that the largest y tick is bigger than the others.  I feel like one tick worth of padding above the line would be ideal visually

- we should set radix (I think that's what it's called) to always 2 decimal points and have the $ prepending the values
Screen Shot 2021-06-09 at 3 29 30 PM

- We should have the same labels for these tooltips as the money (cost/benefits)...it is still cost and benefits, but just in terms of hours instead of money.  These labels were also chosen to be consistent with the legend on the right so it is easy for the user to connect the two together without thinking about it.  Append hours after the number.

Screen Shot 2021-06-09 at 3 29 30 PM

brumik:
@jlmitch5 thanks for your review, to go in order:
I can correct the typos and the hour's number color, however, the tooltips are a bigger chunk of work, and I am actively working on it in the chart builder. Currently what you want is not possible.

Also, the Y-axis size === maximum (it's not cut off, it is exactly as big as the axis :D) value is already addressed in my lib and I am preparing another PR for upgrading the version of the charting library.

src/Containers/SavingsPlanner/TotalSavings.tsx
brumik: Done :+1:

jlmitch5 (approved):

PR #495, opened 2021-06-10
Added EmptyList component to show message and Add button

h-kataria:
This component will be displayed when there are no records to display on the page, along with Add button(if user has access to Add new records)
https://issues.redhat.com/browse/AA-543
Screen Shot 2021-06-10 at 11 23 53 AM

### DIFF
--- a/src/Components/EmptyList.js
+++ b/src/Components/EmptyList.js
@@ -1,0 +1,49 @@
+import React from 'react';
+import PropTypes from "prop-types";
+
+import {
+  Title,
+  EmptyState,
+  EmptyStateIcon,
+  EmptyStateBody, Button,
+} from '@patternfly/react-core';
+import { AddCircleOIcon } from '@patternfly/react-icons';
+import {useHistory, useLocation} from "react-router-dom";
+
+const EmptyList = ({ label = '', title = 'No items found.', message = '', canAdd = false, path }) => {
+  const history = useHistory();
+  return (
+    <EmptyState variant="full">
+      <EmptyStateIcon icon={AddCircleOIcon}/>
+      <Title size="lg" headingLevel="h3">
+        {title}
+      </Title>
+      <EmptyStateBody>{message}</EmptyStateBody>
+      {canAdd && (
+          <Button
+            key="add-item-button"
+            variant="primary"
+            aria-label={label}
+            onClick={() => {
+              history.push({
+                pathname: path,
+              });
+            }}
+          >
+            {label}
+          </Button>
+        )
+        }
+    </EmptyState>
+  )
+};
+
+EmptyList.propTypes = {
+  canAdd: PropTypes.bool,
+  label: PropTypes.string,
+  message: PropTypes.string,
+  title: PropTypes.string,
+  path: PropTypes.string.isRequired
+};
+
+export default EmptyList;
--- a/src/Components/EmptyList.test.js
+++ b/src/Components/EmptyList.test.js
@@ -1,0 +1,33 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import reactRouterDom from 'react-router-dom';
+const pushMock = jest.fn();
+reactRouterDom.useHistory = jest.fn().mockReturnValue({ push: pushMock });
+
+jest.mock('react-router-dom', () => ({
+  useLocation: jest.fn().mockReturnValue({
+    pathname: '/another-route',
+    search: '',
+    hash: '',
+    state: null,
+    key: '5nvxpbdafa',
+  }),
+}));
+
+import EmptyList from './EmptyList';
+
+describe('EmptyList', () => {
+  it('should render successfully', async () => {
+    render(
+      <EmptyList
+        label={'Add plan'}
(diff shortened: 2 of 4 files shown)


src/Components/EmptyList.js
```diff
@@ -0,0 +1,41 @@
...
+          >
+            {label}
+          </Button>
+        )
+        }
+    </EmptyState>
+  )
+};
+
+export default EmptyList;
```
brumik, line 41: [nitpick] Missing newline

src/Components/EmptyList.test.js
```diff
@@ -0,0 +1,32 @@
...
+        label={'Add plan'}
+        title={'No plans added'}
+        message={'No plans have been added yet. Add your first plan.'}
+        canAdd={true} />
+    );
+    expect(screen.getByText('No plans added')).toBeTruthy;
+    expect(screen.getByRole('button')).toBeTruthy;
+    expect(screen.getByText('Add plan')).toBeTruthy();
+  });
+});
```
brumik, line 32: [nitpick] Missing new line

src/Components/EmptyList.js
```diff
@@ -0,0 +1,41 @@
...
+import {
+  Title,
+  EmptyState,
+  EmptyStateIcon,
+  EmptyStateBody, Button,
+} from '@patternfly/react-core';
+import { AddCircleOIcon } from '@patternfly/react-icons';
+import {useHistory, useLocation} from "react-router-dom";
+
+const EmptyList = ({ label = '', title = '', message = '', canAdd = 'false' }) => {
```
brumik, line 12: [nitpick] New lines (or use eslint fix command)

Also can add should be boolean instead of a string.

src/Components/EmptyList.js
```diff
@@ -0,0 +1,41 @@
...
+            }}
+          >
+            {label}
+          </Button>
+        )
+        }
+    </EmptyState>
+  )
+};
+
```
brumik, line 40: Missing prop validation

src/Components/EmptyList.js
```diff
@@ -0,0 +1,41 @@
...
+  EmptyState,
+  EmptyStateIcon,
+  EmptyStateBody, Button,
+} from '@patternfly/react-core';
+import { AddCircleOIcon } from '@patternfly/react-icons';
+import {useHistory, useLocation} from "react-router-dom";
+
+const EmptyList = ({ label = '', title = '', message = '', canAdd = 'false' }) => {
+  const history = useHistory();
+  const {pathname} = useLocation();
```
brumik, line 14: eslint fix ==> `{ pathname } = useLocation()`

src/Components/EmptyList.js
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <EmptyState variant="full">
+      <EmptyStateIcon icon={AddCircleOIcon}/>
+      <Title size="lg" headingLevel="h3">
+        {title || 'No items found.'}
```
brumik, line 19: Move this logic to the component title default value (eg.: Line 12 `... title = 'No items found' ...`)

src/Components/EmptyList.js
```diff
@@ -0,0 +1,41 @@
...
+      </Title>
+      <EmptyStateBody>{message}</EmptyStateBody>
+      {canAdd && (
+          <Button
+            key="add-item-button"
+            variant="primary"
+            aria-label={label}
+            onClick={() => {
+              history.push({
+                pathname: `${pathname}/add`,
```
brumik, line 29: Since it is located in the `Components` folder I think we want to reuse it, so I would probably pass the on-click handler or at least the pathname as a param to the component, so it is really general.

Also we could define the paths somewhere because composing the pathname from a variable and constant this deep is not too readable. Maybe the react router has something to offer.

brumik (changes requested):
It is nice a nice job, there are some linting errors (they do not show up as errors when running lint but locally the prettify wants to make some changes). The only real blocker is the on-click method, look in the comments :) 

Overall I like how it works :)

If you have some extra time, this is an easy component to make it work in typescript (you only need to add prop types to the component itself, so it is a good starter if you want to start to work with TS :)

brumik:
@h-kataria I was also able to compile it, but how you make no data so you can see it working? I was not able to do it.

h-kataria:
> @h-kataria I was also able to compile it, but how you make no data so you can see it working? I was not able to do it.

change https://github.com/RedHatInsights/tower-analytics-frontend/blob/AA-543-savings-planner-empty-view/src/Api.js#L29 to `return []`

brumik:
Verified, the component seems to work as it should. Thanks for the help ^^

h-kataria:
> Verified, the component seems to work as it should. Thanks for the help ^^

np, addressing other comments now

h-kataria:
> Verified, the component seems to work as it should. Thanks for the help ^^

@brumik thanks for testing/reviewing, addressed all the comment please re-review

brumik (approved):
Good to go :clap:

PR #497, opened 2021-06-14
Update Modal to meet Patternfly specifications

knightsean00:
image

Now includes a job breakdown bar, DescriptionList stats, and TableComposable related jobs.


src/Components/Breakdown.js
```diff
@@ -0,0 +1,92 @@
...
+  height: 0.75rem;
+  margin-right: 0.5rem;
+`;
+
+function title(str) {
+  return str[0].toUpperCase() + str.slice(1).toLowerCase();
+}
+
+const Breakdown = ({ categoryCount = {}, categoryColor }) => {
+  const totalCount = Object.keys(categoryCount).reduce(
```
brumik, line 43: Just for your information, you can use the [`Object.values()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values) here and similar situations.
(And probably could have names the `a` and `b` variables a little better :P but I am nitpicking here)

src/Components/ModalContents.js
```diff
@@ -219,10 +253,45 @@ const ModalContents = ({
   useEffect(() => {
     if (selectedId) {
       setStats(readJobExplorer({ params: agreggateTemplateParams }));
-      setRelatedJobs(readJobExplorer({ params: relatedTemplateJobsParams }));
+      setRelatedJobs(readJobExplorer({ params: relatedTemplateJobsParams }))
```
brumik, line 256: Eslint fix :) (on the whole file, there are also unused variables - mainly stylings). You could probably use an editor with linting highlights to make your work better ^^

src/Components/ModalContents.js
```diff
@@ -219,10 +253,45 @@ const ModalContents = ({
...
+      failed: stats.failed_count,
+      new: stats.new_count,
+      pending: stats.pending_count,
+      running: stats.running_count,
+      waiting: stats.waiting_count
+    }
+  }
+
+  const categoryColor = {
+    success: "#95D58E",
```
brumik, line 285: These colors don't match the chart color (bar and line), we need to update one or another to match. I think this can be outside of this PR but we should mark it up in JIRA then. @jlmitch5

brumik (changes requested):
As the previous PR reviews, please use some kind of lint fixer, I know that our CI lint checking is kinda broken. You have plenty of unused variables from imports and custom-styled components, please remove them before merging.

src/Components/ModalContents.js
brumik: And just out of curiosity, where did you get these exact colors in this list?

src/Components/ModalContents.js
```diff
@@ -219,10 +253,45 @@ const ModalContents = ({
...
+  const tableCols = [
+    'Id/Name',
+    'Status',
+    'Cluster',
+    'Finished',
+    'Total time'
+  ]
+
+  let categoryCount;
+  if (stats == null || stats == 0) {
```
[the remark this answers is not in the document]
knightsean00: The tests have been updated in the most recent commits.

src/Components/ModalContents.js
knightsean00: I was given these colors:
image

I kind of just guessed the rest, but I assumed that someone else would probably know much more than I would about the correct color usage. I was going to bring that up as an issue whenever this pull request was approved.

brumik: Oh, okay, then I think we can pull these variables from `patternfly-tokens`. And if this is the pf coloring then we probably should update the other charts to match up.

src/Components/ModalContents.js
```diff
@@ -103,20 +54,20 @@ const formatTopFailedTask = (data) => {
   return `Unavailable`;
 };
 
-const formatTopFailedStep = (data) => {
-  if (!data) {
-    return;
-  }
+// const formatTopFailedStep = (data) => {
```
brumik, line 57: Why is this function just commented out?

brumik, line 57: Why is this function just commented out?

src/Components/ModalContents.js
```diff
@@ -223,6 +180,71 @@ const ModalContents = ({
...
+        failed: stats.failed_count,
+        new: stats.new_count,
+        pending: stats.pending_count,
+        running: stats.running_count,
+        waiting: stats.waiting_count,
+      }
+    : null;
+
+  const categoryColor = {
+    success: '#95D58E',
```
brumik, line 199: As I mentioned before we could use the PF tokens to get these values from https://www.npmjs.com/package/@patternfly/react-tokens. I am fine if this is in a followup PR.

brumik (changes requested):
It is great and nearly there. I didn't understand why is the function commented out, and that is the blocker. 

I like how this refactor made the code more DRY.

However, I spotted a problem while reviewing: the modal window has an API call and does not handle the loading, and the error states. This causes the modal window to show for a second the previous modal window data on open. If you want, tackle this in this PR or in a follow-up PR, if not, I'll make a ticket and will do it later :)

src/Components/ModalContents.js
knightsean00: I left the function commented out because it seemed like it might be needed later if more information needs to be included in the modal. If we end up not using it I'll just delete it

src/Components/ModalContents.js
knightsean00: Resolved in most recent commit

brumik:
@knightsean00 can you please rebase to devel? The lint check should be fixed there now :)

brumik (approved):
LGTM
Followup - Modal API status handling, removing the commented code
Potential follow-up work: Upgrade and get the chart builder to create horizontal bar charts.

PR #498, opened 2021-06-14
Update JobExplorer to use Composable Table

knightsean00:
image


src/Components/JobExplorerList.js
```diff
@@ -2,217 +2,83 @@ import React, { useState, useEffect } from 'react';
...
+            cols.map((head, idx) => (
+              <Th key={`col-${idx}`}>{head}</Th>
+            ))
+          }
+        </Tr>
+      </Thead>
+      <Tbody>
+        {
+          jobs.map((job, idx) => {
+            rowIdx += 2;
```
jlmitch5, line 46: use the job's id for key purposes as well, I don't believe index of the map or the row is needed for uniquity, as the id will always be present and should always be unique.

jlmitch5 (changes requested):
This looks and functions fantastic...and you've been able to delete over a 100 LOC in favor of using patternfly standards!

I think conceptually it would be a bit simpler if id from the api was used instead of array or row indexes.

src/Components/JobExplorerList.js
```diff
@@ -2,217 +2,78 @@ import React, { useState, useEffect } from 'react';
...
+                      onToggle: handleExpansion
+                    }}
+                  />
+                  <Td href={job.id.tower_link}>{`${job.id.id} - ${job.id.template_name}`}</Td>
+                  <Td><JobStatus status={job.status} /></Td>
+                  <Td>{job.cluster_name}</Td>
+                  <Td>{job.org_name}</Td>
+                  <Td>{formatJobType(job.job_type)}</Td>
+                </Tr>
+                <Tr isExpanded={expanded[idx] === true}>
```
brumik, line 60: I don't know about this, but in the [examples](https://www.patternfly.org/v4/components/table) they seem to use the `ExpandableRowContent` component for this kind of functionality. But I never worked with the table component so @jlmitch5  probably has more insight into this.

brumik (commented):
Please run lint fix on this file (i am using vscode where you can do it from its commands, so I cannot supply you with an npm command which fixes one file, however, you can run `npm run lint:js:fix` to run on all files, commit this only file and discard changes on the other ones as workaround).

Also, the list seems to work fine :+1:

src/Components/JobExplorerList.js
knightsean00: Alright, my latest commit should have linting all set up, I forgot to create a .vscode config for my local directory.

I actually based my usage of the Expandable TableComposable from the PatternFly examples, so, as for the usage of `ExpandableRowContent`, it looks like they use it as a child of <Td> under <Tr>

image

brumik (approved):
I love how this refactor turned out, so satisfying :clap:

src/Components/JobExplorerList.js
```diff
@@ -1,218 +1,82 @@
...
+      <Thead>
+        <Tr>
+          <Th />
+          {cols.map((head, idx) => (
+            <Th key={`col-${idx}`}>{head}</Th>
+          ))}
+        </Tr>
+      </Thead>
+      <Tbody>
+        {jobs.map((job, idx) => (
```
jlmitch5, line 39: rather than using idx here, i would just use the job's id as the key to do everything off of.

jlmitch5 (changes requested):

src/Components/JobExplorerList.js
```diff
@@ -56,7 +61,7 @@ const JobExplorerList = ({ jobs }) => {
               <Td>{job.org_name}</Td>
               <Td>{formatJobType(job.job_type)}</Td>
             </Tr>
-            <Tr isExpanded={expanded[idx] === true}>
+            <Tr isExpanded={expanded[job.id.id] === true}>
```
brumik, line 64: you cna use instead of `expanded[job.id.id] === true` just `!!expanded[job.id.id]`

src/Components/JobExplorerList.js
```diff
@@ -221,9 +90,4 @@ JobExplorerList.propTypes = {
   windowWidth: PropTypes.number,
```
brumik, line 90: If jobs has no default value in the destructuring it should be required. And I think `windowWidth` is not used anymore.

src/Components/JobExplorerList.js
brumik: In situations like this I like to create a small dump component for each entry that manages its own `[expanded, setExpanded] =  useState(false)` state. I think that is the most readable way to go about it. What do you think @jlmitch5 ?

src/Components/JobExplorerList.js
```diff
@@ -16,20 +16,53 @@ import JobStatus from './JobStatus';
...
-      obj[job.id.id] = false;
-      return obj;
-    }, {})
-  );
-  const handleExpansion = (event, idx) => {
-    setExpanded({
-      ...expanded,
-      [idx]: !expanded[idx],
-    });
+const Row = ({ job }) => {
```
brumik, line 19: I think we are having separate files for each component. This could be in its own too (as John mentioned yesterday something like: `JobExplorerListRow`)

knightsean00: Oh, my bad, I thought that since it was a dump component I was just gonna put it in the same file like we have with other components.

jlmitch5:
Layout for the expanded row seems slightly awkward...I'd like to see three rows of the same width (and shrinking to two/one column for responsiveness

Screen Shot 2021-06-30 at 12 18 57 AM

I don't think we need to use TDs for the expanded content...rathere, I think we can just treat this expandable section as a pane and do arbitrary layout.  In this case, I think the DescriptionList pf components like we did for the savings planner detail screen would work well for this.

jlmitch5 (approved):
Awesome looks great!

PR #499, opened 2021-06-15
AA-544 add basic RBAC for add, edit, delete Savings Plan.

kialam:
related: https://issues.redhat.com/browse/AA-544

view only users see a list of savings plans (if any):
Screen Shot 2021-06-15 at 2 11 59 PM

editor or admin users see Add and Delete buttons as well as kebab menu on the cards for Edit and Link Template workflows:
Screen Shot 2021-06-15 at 2 11 24 PM

This PR implements a _basic_ RBAC UX workflow based on the permissions returned from the `plan_options` endpoint. There are no additional required API calls made to determine RBAC, so this is basically setting a boolean and showing/hiding/redirecting based on that boolean.

### DIFF
--- a/src/Containers/SavingsPlanner/Add/index.js
+++ b/src/Containers/SavingsPlanner/Add/index.js
@@ -24,7 +24,7 @@ const Add = () => {
   useEffect(() => {
     setOptions(readPlanOptions());
   }, []);
-  const canWrite = (options.isSuccess && (options.data.meta.rbac.perms.write === true || options.data.meta.rbac.perms.all === true));
+  const canWrite = (options.isSuccess && (options.data?.meta?.rbac?.perms?.write === true || options.data?.meta?.rbac?.perms?.all === true));
   const title = 'Create new plan';
 
   const showAdd = () => (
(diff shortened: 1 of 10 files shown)


src/Containers/SavingsPlanner/Add/index.js
```diff
@@ -16,16 +17,17 @@ import { readPlanOptions } from '../../../Api';
...
+import { Paths } from '../../../paths';
+
 const Add = () => {
   const [options, setOptions] = useApi({});
-
   useEffect(() => {
     setOptions(readPlanOptions());
   }, []);
-
+  const canWrite = (options.isSuccess && (options.data.meta.rbac.perms.write === true || options.data.meta.rbac.perms.all === true));
```
jlmitch5, line 27: can we expect options is returned in this format always?  If not, it may be better to use the ?. syntax so we don't run into thrown errors trying to access on undefined things.

jlmitch5 (commented):
From looking at the code, just a question of if the can* booleans need to be written with the ?. syntax or not.

Pulling down and taking a look at the app now

src/Containers/SavingsPlanner/Add/index.js
kialam: 👍

jlmitch5 (approved):
Looks good!

PR #503, opened 2021-06-16
[AA-533] Pass queryParams through Clusters page when redirecting to Job Explorer

knightsean00:
https://issues.redhat.com/browse/AA-533


jlmitch5 (approved):
works great and code looks great!

PR #504, opened 2021-06-23
Fixing linting

brumik:
This PR fixes linting, lots of problems with default values and prop type checking, and also moves the pagination on the savings planner page under the cards, so it is les confusing. Further work probably could make it better, but we were not so nitpicking this time.

### DIFF
--- a/src/Components/DeleteButton/DeleteButton.js
+++ b/src/Components/DeleteButton/DeleteButton.js
@@ -19,13 +19,13 @@ const Label = styled.span`
 
 const DeleteButton = ({
   onConfirm,
-  modalTitle,
-  name,
-  variant,
-  isDisabled,
+  modalTitle = 'Delete',
+  name = '',
+  variant = 'secondary',
+  isDisabled = true,
   ouiaId = null,
-  deleteMessage,
-  deleteDetailsRequests,
+  deleteMessage = 'Delete?',
+  deleteDetailsRequests = [],
   disabledTooltip,
   children,
 }) => {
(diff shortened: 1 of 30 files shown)


src/Components/Toolbar/ToolbarDeleteButton.js
```diff
@@ -264,20 +249,20 @@ function ToolbarDeleteButton({
...
+// TODO Harpreet Kataria I was not able to figure out the exact type of the
+// deleteDetailsRequests. Cna you please fill in?
+// Same for the delete message
+ToolbarDeleteButton.propTypes = {
+  itemsToDelete: PropTypes.arrayOf(ItemToDelete).isRequired,
+  pluralizedItemName: PropTypes.string,
+  errorMessage: PropTypes.string.isRequired,
+  onDelete: PropTypes.func.isRequired,
+  warningMessage: PropTypes.node,
+  deleteDetailsRequests: PropTypes.array.isRequired,
```
h-kataria, line 263: looks correct to me, it should be array of requests.

brumik: And can you describe what is in the request object? I would like to put here something like
```js
PropTypes.arrayOf(PropTypes.shape(...))
```

src/Components/DeleteButton/DeleteButton.js
```diff
@@ -144,14 +147,21 @@ function DeleteButton({
...
+  name: PropTypes.string.isRequired,
+  variant: PropTypes.object.isRequired,
+  isDisabled: PropTypes.bool.isRequired,
   ouiaId: PropTypes.string,
-};
-
-DeleteButton.defaultProps = {
-  ouiaId: null,
+  deleteMessage: PropTypes.string.isRequired,
+  deleteDetailsRequests: PropTypes.array.isRequired,
```
h-kataria, line 162: LGTM, it is expected to be an array of request objects

h-kataria (approved):

src/Components/DeleteButton/DeleteButton.js
```diff
@@ -10,31 +10,33 @@ import ErrorDetail from '../ErrorDetail';
...
+
+const DeleteButton = ({
   onConfirm,
-  modalTitle,
-  name,
-  variant,
+  modalTitle = 'Delete',
+  name = '',
+  variant = 'secondary',
+  isDisabled = true,
```
h-kataria, line 25: this should be dafaulted to `false` sorry...my bad

PR #506, opened 2021-06-25
Clear Modal state upon closing

knightsean00:
Effectively eliminates persistent job and stat display occuring after clicking on another modal popup.

### DIFF
--- a/src/Components/ModalContents.js
+++ b/src/Components/ModalContents.js
@@ -66,21 +66,6 @@ const formatTopFailedTask = (data) => {
   return `Unavailable`;
 };
 
-// const formatTopFailedStep = (data) => {
-//   if (!data) {
-//     return;
-//   }
-
-//   if (data && data[0]) {
-//     const percentage = Math.ceil(
-//       (data[0].failed_count / data[0].total_failed_count) * 100
-//     );
-//     return `${data[0].template_name} ${percentage}%`;
-//   }
-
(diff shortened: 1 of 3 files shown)


brumik:
I think this PR could incorporate API error handling and API loading screen too. Imagine somebody with slow internet connection, seeing an empty modal without loading screen popping up. They would probably think there is a bug instead of thinking it has no loading screen (same for accidental API error). If loading screen is set up correctly I don't think we need to cleanup the state by setting the data to empty array.

brumik (changes requested):
I think this PR could incorporate API error handling and API loading screen too. Imagine somebody with slow internet connection, seeing an empty modal without loading screen popping up. They would probably think there is a bug instead of thinking it has no loading screen (same for accidental API error). If loading screen is set up correctly I don't think we need to cleanup the state by setting the data to empty array.

For template how to do it you can look for the "page" components, like here: https://github.com/RedHatInsights/tower-analytics-frontend/blob/devel/src/Containers/Clusters/Clusters.js#L71

brumik (approved):
Sorry for pushing a commit @knightsean00 . After I checked out I realized that going with the Loading screen is "blinking" too much on reopens, so plated around a bit putting the isLoading to different places but I think the best solution is to merge your previous version with this, keeping the cleanup and the loading screen too.

PR #508, opened 2021-06-29
Converted custom Main component styling to grid and stack

brumik:
The Grid components by default works with 12 spans, where the closest looking division was 9:3 (which is also used on the Statistics tab at savings planner) so the right column is slightly smaller than was before.

The TopTemplates component was made to span until the end of the page so the user can see more than 1 template at once instead of lots of gray empty space. The left column made it match in height so it still looks consistent. 

Additional gains: easier to convert the page to a more responsive "mobile-friendly" version if we wish so (I did not looked into it but hoped it would work out of the box).

**Before:**
Screenshot from 2021-06-29 15-41-45

**After:**
Screenshot from 2021-06-29 15-52-37

### DIFF
--- a/src/Containers/AutomationCalculator/AutomationCalculator.js
+++ b/src/Containers/AutomationCalculator/AutomationCalculator.js
@@ -1,13 +1,19 @@
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
-import styled from 'styled-components';
 import Main from '@redhat-cloud-services/frontend-components/Main';
 import NotAuthorized from '@redhat-cloud-services/frontend-components/NotAuthorized';
 import {
   PageHeader,
   PageHeaderTitle,
 } from '@redhat-cloud-services/frontend-components/PageHeader';
-import { Card, CardBody } from '@patternfly/react-core';
+import {
+  Card,
+  CardBody,
+  Grid,
+  GridItem,
+  Stack,
+  StackItem,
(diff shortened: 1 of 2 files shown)


src/Containers/AutomationCalculator/AutomationCalculator.js
```diff
@@ -181,74 +168,73 @@ const AutomationCalculator = ({ history }) => {
...
+        </Stack>
+      </StackItem>
+    </Stack>
   );
 
   if (preflight?.error?.status === 403) {
     return <NotAuthorized {...notAuthorizedParams} />;
   }
 
+  const renderContents = () => {
```
brumik, line 214: Created this function to avoid putting `<Main>` component into each subcomponent. It does not accept clones so this is the only way it seems we can do this. I don't remember the case but I know @h-kataria you had somewhere a load of `Main` components.

jlmitch5:
Code looks good

@knightsean00 could you review this PR?  run locally and inspect to see if there are any changes in layout between devel and this.

knightsean00:
Everything looked good when I ran it locally. Unfortunately the mobile version doesn't work too well right now, but that can definitely be addressed now that we are using a responsive grid.

knightsean00 (approved):
Everything looked good when I ran it locally. Unfortunately the mobile version doesn't work too well right now, but that can definitely be addressed now that we are using a responsive grid.

PR #509, opened 2021-06-29
AA-570 AA-571 various fixes for savings planner ux audit

jlmitch5:
Some small tweaks that need to be made to the UI before release of savings planner:

- Plan cards should default to a page size of 10.  Hide pagination and bottom toolbar if empty list shows
- Revert pagination change and the overall plan cards section scrolls, even when the cards only take up some of the viewport.  Should scroll only we needed
- Cards should be tweaked to have a layout similar to hub (taller cards, more breathing room, more spacing between cards).
- All card detail labels should be bold, there should be a bit more padding between each detail, and no italics for date.
- Card titles should word-break if needed to prevent layout issues
- Tooltip on delete button should show why it’s disabled
- Red exclamation point icon in delete alert modal
- There should be a manage tasks item in kebab menu
- There should only be the red asterisk on name in the add/edit form (not type)
- Use options to get prettified label version of certain details in card/detail view
-  For Detail view
    - Tasks title gets bolded/smaller
    - No inset of actual tasks
    - Hr gets padding padding top 24px only;
    - Change date/time to not italics
- Small url bugs, there can sometimes be two slashes in url `savings-planner//add`, and the id sometimes is put as `:id19` instead of `19`.
- Add/Edit should redirect you to detail not list
- Resave after edit should regrab data for detail view
- Don't allow adding of empty tasks
- Disable save button based on if a name is passed or not.
- When the savings planner save fails (due to either a user error such as not including a plan name.  Or an api error such as duplicate plan name, display a modal that their was an error, as well as the response from the api)


src/Containers/SavingsPlanner/List/ListItem/index.js
```diff
@@ -53,7 +77,7 @@ const PlanCard = ({
     template_details,
   } = plan;
   const [isCardKebabOpen, setIsCardKebabOpen] = useState(false);
-
+  console.log(options);
```
h-kataria, line 80: @jlmitch5 looks like a leftover debug statement

src/Containers/SavingsPlanner/Shared/Form/index.js
```diff
@@ -0,0 +1,215 @@
...
+      ),
+      nextButtonText: 'Save',
+    },
+  ];
+
+  const CustomFooter = (
+    <WizardFooter>
+      <WizardContextConsumer>
+        {({ activeStep, onNext, onBack, onClose }) => {
+          console.log(data);
```
h-kataria, line 72: here too(extra debug statement)

h-kataria (approved):
Looks great, verified all the changes in UI. Great job @jlmitch5

PR #511, opened 2021-07-01
Minor ts fix

brumik:
* Changed `Number` to `number`: [Typescript do and dont's](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html)
* Moved the conversion outside the helper functions since it does not check for `NaN`
* Converted everywhere the params to the function with `+` for consistency


src/Containers/SavingsPlanner/Shared/currencyFormatter.ts
```diff
@@ -1,9 +1,9 @@
-const formatCurrency = (s: string): string => {
+const formatCurrency = (s: number): string => {
```
jlmitch5, line 1: super nitpick, but can we change the var name to n?

jlmitch5 (approved):
Very small nitpick change, but otherwise this is great.  It was a good call to make both of them accept number, the + is cleaner than parseInt.  Looks great

src/Containers/SavingsPlanner/Shared/currencyFormatter.ts
brumik: Ill fix, and the hourFormatter thing here too soon :)

PR #513, opened 2021-07-01
Clean CI

brumik:
After quick consult with @benthomasson :
* Travis is duplicate on PRs, now it runs only on the branches we need to deploy.
* Removing the docker initialization from the repo, it seems like nobody is using it. If somebody complains we can add it back :) (fun times)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,0 @@
-**/node_modules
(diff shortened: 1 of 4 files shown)


brumik:
Reopening to see without PR build the travis

PR #514, opened 2021-07-01
Ci clean 2

brumik:
After quick consult with @benthomasson :
* Travis is duplicate on PRs, now it runs only on the branches we need to deploy (devel, prod-stable)
* Removing the docker initialization from the repo, it seems like nobody is using it. If somebody complains we can add it back :) (fun times)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,0 @@
-**/node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,8 @@
+branches:
+  only:
+  - devel
+  - prod-stable
+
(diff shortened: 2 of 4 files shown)


brumik:
Travis errors out because I disabled it to run on PRs, but only after I closed the previous pr #513 which now Travis is unable to pull down, causing to error out with: 
```
The command "eval git fetch origin +refs/pull/513/merge: " failed. Retrying, 2 of 3.

fatal: couldn't find remote ref refs/pull/513/merge
```

Should be safe to merge

jlmitch5 (approved):

PR #516, opened 2021-07-01
Savings planner QE audit fixes

jlmitch5:
Just a couple of tiny fixes:
- sets a max height for the category dropdown
- pass 1 if no host number is specified, i.e. the default is deleted from the field
- hide details section of error modal if there are none passed by the API for us to display.
- make onKeyDown enter work for the name search.


h-kataria (approved):
LGTM, good catches

PR #517, opened 2021-07-02
Feature flags

brumik:
For running the FeatureFlags locally:
https://gitlab.cee.redhat.com/automation-analytics/unleash-proxy
* Currently: `launcher` needs tweaking
* `launcher` doesnt know HTTPS, need of a tool for creating HTTPS from HTTP (tunneling)

The context provider and the consumer works like charm otherwise :)


src/Api.js
```diff
@@ -43,177 +45,137 @@ function handleResponse(response) {
...
-  return authenticatedFetch(url, {
-    method: 'POST',
-    body: JSON.stringify(params),
-  }).then(handleResponse);
-};
+export const preflightRequest = () =>
+  authenticatedFetch(preflightEndpoint).then(handleResponse);
 
-export const readROI = ({ params = {} }) => {
+const readWithPagination = (endpoint, params) => {
```
jlmitch5, line 69: very nice abstraction!

src/Api.js
```diff
@@ -43,177 +45,137 @@ function handleResponse(response) {
...
 export const readPlan = ({ params = {} }) => {
-  let url = new URL(plansEndpoint, window.location.origin);
+  const url = new URL(plansEndpoint, window.location.origin);
   return authenticatedFetch(url, {
     method: 'POST',
     body: JSON.stringify(params),
   }).then(handleResponse);
 };
 
 export const readClustersOptions = ({ params = {} }) => {
```
jlmitch5, line 131: The options here all look identical too.  could you make a similar function for these?

src/FeatureFlags/FeatureFlag.tsx
```diff
@@ -0,0 +1,20 @@
...
+  <Consumer>
+    {features => 
+      (features[name] && features[name].isEnabled)
+        ? children
+        : defaultChildren
+    }
+  </Consumer>
+);
+
+export default FeatureFlag;
```
jlmitch5, line 22: export so that this is used

```
useFeatureFlags()
```

jlmitch5 (changes requested):
looks good just a couple things

src/Api.js
brumik: I was thinking about it but it is a little, maybe we can export the `endpoints` and have one function for reading the options and one for reading the data? :) that way a new endpoint would be just about defining and exporting the route

src/Api.js
```diff
@@ -23,7 +23,9 @@ const clustersOptionsEndpoint =
   '/api/tower-analytics/v1/dashboard_clusters_options/';
 const planOptionsEndpoint = '/api/tower-analytics/v1/plan_options/';
 
-function handleResponse(response) {
+const featuresEndpoint = '/api/featureflags/';
```
jameswnl, line 26: Currently, it is being served at `/api/featureflags/v0`

jlmitch5 (approved):

PR #518, opened 2021-07-02
Fixed prop icon undefined passed error on savings empty list

brumik:
The icon prop was giving prop validation errors. I could not get the icon to pass correctly but since the empty list is already pretty specific with the add button, I moved the icon there, and now all bad red errors are gone from the console.


jlmitch5 (approved):

jlmitch5:
looks good merging

PR #519, opened 2021-07-05
Grid components - what I missed before

brumik:


h-kataria:
@brum does the screen look different visually with these changes?

brumik:
@h-kataria as far as I know, there are no visual changes

knightsean00 (approved):
This looks great! We might want to consider making a different breakpoint for small devices (`sm={12}`), but I don't think this is super urgent because most users will not be viewing these pages on a mobile device anyway.

h-kataria (approved):
great work!

PR #520, opened 2021-07-07
Api usage unification

brumik:
It is the same. Well, more or less. It is in TS now. And it is shorter. Much shorter. However lot's of types...

What we could do:
* Export only the endpoints and the different fetching functions (get, post, delete, update)
* We probably would like to get a little unification on the backed side how the plan deleting/updating and returning works (@jlmitch5 I remember this from the session with you, that once the ID it was outside of the object and other time inside)
  * If we do this we could write a `delete` and a `update` function too like the `post`, `postWithPagination` and `get` functions now
* We could use get for all options endpoints OR we could use post for the plans_options endpoint too :)


src/Api.ts
```diff
@@ -0,0 +1,168 @@
...
+        status: response.status,
+        error: 'RBAC access denied',
+      });
+    } else {
+      return Promise.reject(json);
+    }
+  });
+};
+
+declare global {
```
jlmitch5, line 46: can we move this to the top (or right below import)

src/Api.ts
```diff
@@ -0,0 +1,168 @@
...
+export const get = (endpoint: string, params: Params = {}) => {
+  const url = new URL(endpoint, window.location.origin);
+  url.search = stringify(params);
+
+  return authenticatedFetch(url.toString(), {
+    method: 'GET'
+  }).then(handleResponse);
+}
+
+/* Reading Data */
```
jlmitch5, line 108: I think it makes more sense to group by resource ("Plans", "JobExplorer", "ROI", etc.) rather than type of call.  From a usage perspective, you'd want to what the available Plans api endpoints are, I think if you need to do something new with plans, rather than having to read through everything.

src/Api.ts
```diff
@@ -0,0 +1,168 @@
+import { stringify } from 'query-string';
```
jlmitch5, line 1: At this point, it's pretty clear this file is trying to do a few things.

1. provide a unified place for all functions used in the application to interact with the API (endpoint strings and exported endpoint functions)

2 a. provide a set of abstractions/utilities which are used to actually do the interaction with the api (handleResponse, authenticatedFetch
2 b. provide a set of functions which handle REST verbs (postWIthPagination, post, get ...update and delete haven't been abstracted yet but it makes sense they would be)

3. There's also the sort of related getFeatures and preflightRequest which are kind of unique cases that don't line up cleanly with the other stuff.

---

I propose we make a new API folder with 2 files.  This API file handles "1", and a new API utilities file handles "2a" (which are not exported), "2b" which are exported and used in "1", and "3" (my thought being so we can keep "2a" function unexported to improve clarity of what the public API is).  I'm not fully invested in those particulars, but it does seem to be a way we could make this stuff a little clearer.

src/Containers/SavingsPlanner/Details/Details.js
```diff
@@ -58,7 +58,7 @@ const Details = () => {
     });
     const fetchEndpoints = () => {
       setData(readPlan({ params: queryParams }));
-      setOptions(readPlanOptions());
+      setOptions(readPlanOptions({}));
```
jlmitch5, line 61: seems weird to require an empty object here.  I'd set the default to {} and allow null for the param type.

jlmitch5 (changes requested):
This is an interesting and very cool PR, and I think there's some really great stuff going on.  some suggestions for taking it further inlined.

I'm assuming all the feature flag stuff will go away.

src/Api.ts
brumik: I think all of these one-liners could go away in favor of exporting endpoints and the functions itself.

So for example instead of using:
```js
const [ data, setData ] = useApi();
....
setData(readJobExplorer(queryParams)
setData2(readJobExplorerOptions(queryParams)
```
we could do something like:
```js
....
setData(postWithPagination(jobExplorerEndpoint, queryParams);
setData2(post(jobExplorerOptionsEndpoint, queryParams);
```

This requires a little more typing in each components for sure, but also increases transparency and removes the need for the one-liner functions we have currently

brumik: and in adition we could use a
```js
export const exndpoints = {
  jobExplorer: {
    get: '...',
    options: '...',
  },
  savingsPlanner: {
    get: '...',
    options: '...',
    delete: '...',
    create: '...',
    ...
  }
  ...
}
```

src/Api/types.ts
```diff
@@ -0,0 +1,13 @@
...
+  limit?: string | number,
+  offset?: string | number,
+  sort_by?: string,
+  [x: string]: string | number | undefined
+};
+
+export type ReadParams = { params: Params };
+export type ReadParamsWithPagination = { params: ParamsWithPagination };
+export type DeleteParams = { id: string };
+export type UpdateParams = { id: string, params: Params };
```
jlmitch5, line 13: i'm not sure we are gating linting with these type script files..this should have a new line at the end of the file.  I've noticed lots of red squiggles in my editor for the ts files to.

jlmitch5 (approved):
Looking good, check ts file linting before merge.

src/Api/types.ts
brumik: There is no linting in place for TS I realized, I added linting in the other repo #526, Ill fix up there these too

PR #523, opened 2021-07-08
Add two new unit tests

knightsean00:


src/Containers/SavingsPlanner/Add/index.test.js
```diff
@@ -0,0 +1,43 @@
...
+    expect(screen.getByText('Link template')).toBeTruthy();
+  });
+
+  test('redirects upon 403', async () => {
+    api.preflightRequest.mockRejectedValue(mockResponses.preflightRequest403);
+    api.readPlanOptions.mockResolvedValue({
+      data: { meta: { rbac: { perms: { all: false } } } },
+      isSuccess: true,
+    });
+    var wrapper;
```
brumik, line 35: Using `var` was the main source of confusion and bugs before the `let` era. The difference is explained [here](https://stackoverflow.com/questions/762011/whats-the-difference-between-using-let-and-var) pretty well. It is nothing wrong to use it, but it is not considered a good practice anymore.

src/Utilities/tests/helpers.reactTestingLib.js
```diff
@@ -26,3 +26,16 @@ export const renderPage = (Component, { search } = defaultParams) =>
       </MemoryRouter>
     </Provider>
   );
+
+export const renderPageWithProps = (
```
brumik, line 30: When I wrote the `renderPage` I didn't considered props and other components that will need provider and router.

Can you (in this or a followup PR) unify these two functions like:
```js
export const renderComponentWithUtils = (Compoenent, { search } = defaultParams, props = {} ) => render(...);
```
If you have better name for it I'll take it, I could not come up with better rn :D

src/Containers/SavingsPlanner/Shared/Form/Steps/Templates/index.test.js
```diff
@@ -0,0 +1,91 @@
...
+      renderPageWithProps(Templates, defaultProps);
+    });
+
+    expect(screen.getByText('Link a template to this plan:')).toBeTruthy();
+
+    expect(screen.getByText('test_template_name_0')).toBeTruthy();
+    expect(screen.getByText('test_template_name_2')).toBeTruthy();
+
+    fireEvent.click(screen.getByTestId('radio-345'));
+    fireEvent.click(screen.getByTestId('radio-1'));
```
brumik, line 89: :+1: I like that you added a testId

Maybe you should test if the click did something or not (I would move that to another test case `it open the row when clicked on it`)

src/Containers/SavingsPlanner/Shared/Form/Steps/Templates/index.test.js
```diff
@@ -0,0 +1,91 @@
...
+
+const defaultProps = {
+  template_id: -2,
+  dispatch: mockDispatch,
+};
+
+describe('SavingsPlanner/Shared/Form/Templates', () => {
+  beforeEach(() => {
+    api.preflightRequest.mockResolvedValue(mockResponses.preflightRequest200);
+    api.readJobExplorer.mockResolvedValue({
```
brumik, line 20: I think the mock for this job explorer could go into the `__fixtures__` too with some descriptive name. When we created fixtures we planned to have responses there for most of our queries so we can reuse them in different suites if needed.

(This is also true for the `planOptions` tests with the `rbac false`)

brumik (changes requested):
I like the overall work :+1:  I have some suggestions how could the tests be further improved :)

src/Containers/SavingsPlanner/Add/index.test.js
```diff
@@ -0,0 +1,43 @@
+import '@testing-library/jest-dom';
```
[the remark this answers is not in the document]
knightsean00: I'm using a function (`toBeEmptyDOMElement()`). Also this package is just an extension of `testing-library` as a whole with quite a bit more functionality, so I thought it would be a useful addition as we begin to migrate from Enzyme.

src/Containers/SavingsPlanner/Add/index.test.js
```diff
@@ -0,0 +1,43 @@
...
+      data: { meta: { rbac: { perms: { all: false } } } },
+      isSuccess: true,
+    });
+    var wrapper;
+
+    await act(async () => {
+      wrapper = renderPage(Add);
+    });
+
+    expect(wrapper.container).toBeEmptyDOMElement();
```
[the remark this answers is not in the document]
knightsean00: I couldn't find a good way to determine if the page was empty based on just `screen` alone

src/Utilities/tests/helpers.reactTestingLib.js
knightsean00: Wait sorry, isn't that already done?

src/Containers/SavingsPlanner/Add/index.test.js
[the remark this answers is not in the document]
jlmitch5: I like this approach.  It's mentioned in the testing-library docs https://testing-library.com/docs/ecosystem-jest-dom/

I think this well help make things a bit easier to implement, as @knightsean00 mentions

src/Containers/SavingsPlanner/Add/index.test.js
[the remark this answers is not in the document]
jlmitch5: see other comment in this file

src/Containers/SavingsPlanner/Shared/Form/Steps/Templates/index.test.js
jlmitch5: I agree with @brumik moving this out to fixtures makes a lot of sense

src/Containers/SavingsPlanner/Shared/Form/Steps/Templates/index.test.js
jlmitch5: I'm fine either way here.

jlmitch5 (commented):
Really great tests.  I've piggy-backed my review off of @brumik's which had some good thoughts.

src/Containers/SavingsPlanner/Add/index.test.js
```diff
@@ -0,0 +1,44 @@
...
+    api.readPlanOptions.mockResolvedValue(
+      mockResponses.readPlansOptionsRBACFalse
+    );
+    let wrapper = renderPage(Add);
+
+    await waitFor(() => {
+      return expect(api.readPlanOptions).toHaveBeenCalledTimes(1);
+    });
+
+    expect(wrapper.container).toBeEmptyDOMElement();
```
brumik, line 42: I still don't like using the `.container` but it does its job.

brumik (approved):
Looking great 👏

PR #524, opened 2021-07-09
First set of reports

brumik:
Screenshot from 2021-07-13 13-32-21


src/Containers/OrganizationStatistics/OrganizationStatistics.js
```diff
@@ -220,10 +220,7 @@ const OrganizationStatistics = ({ history }) => {
                 <Tab eventKey={1} title={'Hosts'} />
               </Tabs>
               <CardBody>
-                {orgs.isLoading && <LoadingState />}
-                {orgs.error && <ApiErrorState message={orgs.error.error} />}
-                {orgs.isSuccess && orgs.data?.data.length <= 0 && <NoData />}
-                {orgs.isSuccess && orgs.data?.data.length > 0 && (
+                <ApiStatusWrapper api={orgs}>
```
brumik, line 223: Sorry for these here, I was just testing if it is working correctly

src/Containers/Reports/Details/Report.tsx
```diff
@@ -89,11 +89,19 @@ const Report: FunctionComponent<ReportGeneratorParams> = ({
   const { queryParams, setFromPagination, setFromToolbar } =
     useQueryParams(defaultParams);
 
-  const [attrPairs, setAttrPairs] = useState(extraAttributes);
+  const [attrPairs, setAttrPairs] = useState<Array<any>>([]);
```
brumik, line 92: Can you please use not `any`? I believe the the `extraAttributes` have the same type than this should be, so you can just copy or pull the type in.

For the array types please use `any[]` or `string[]`. They are equivalent but the standard is to use `Type[]` instead of `Array<Type>`

src/Containers/Reports/Details/Report.tsx
```diff
@@ -128,16 +131,13 @@ const Report: FunctionComponent<ReportGeneratorParams> = ({
     return {
       sort: {
         sortBy: {
-          index:
-            attrPairs.findIndex(
-              ({ key }) => key === queryParams.sort_options
-            ) || 0,
-          direction: queryParams.sort_order || 'none',
+          index: combinedAttrPairs.findIndex(({ key }) => key === queryParams.sort_options) || 0,
```
brumik, line 134: Is this and

src/Containers/Reports/Details/Report.tsx
```diff
@@ -171,23 +171,22 @@ const Report: FunctionComponent<ReportGeneratorParams> = ({
           >
             <Thead>
               <Tr>
-                {attrPairs.map(({ key, value }) => (
-                  <Th key={key} {...getSorParams(key)}>
-                    {value}
-                  </Th>
+                {combinedAttrPairs.map(({ key, value }) => (
```
brumik, line 174: this automatically formatted? Because then we need update the rules, the longer line got onelined and the short line was broken up into 5 lines, doesn't seem too consistent.

brumik (commented):
Thanks for updating this repo. I wasn't expecting the props to change after the initial render but it can definitely happen if we can solve partial page rerenders.

brumik:
I rebased to the QS and useRequest. It is definitely not working now. The api is getting querried like 100 times and then the page brokes with a stack trace error. I suspect the problem can be that I needed to call the `getQsConfig` inside the component, which always changes creating a pretty big look until eventually my chart builder breaks. Can somebody help with it @jlmitch5 @h-kataria ?

Also I realized that the useRequest wont work like this:
```ts
const { request: setData, ...api } = useRequest(
    useCallback(
      async () => await readData({ params: queryParams }),
      [queryParams]
    ),
    {} as ApiJson
  );
```

For some weird reason it works much better like this:
```ts
const { request: setData, ...api } = useRequest(
    useCallback(async () => {
      const result = await readData({ params: queryParams });
      return result;
    }, [queryParams]),
    {} as ApiJson
  );
```

And I have no idea why. In its core it is the same, just not saving the result in an intermediary variable. It would be nice if somebody could explain to me, it is bugging me. 

Another note: we should make the qs up with Typescript, I just had to add because of it (and the querry params) like 7 eslint disabling rules and it is still hacky.

h-kataria:
> I rebased to the QS and useRequest. It is definitely not working now. The api is getting querried like 100 times and then the page brokes with a stack trace error. I suspect the problem can be that I needed to call the `getQsConfig` inside the component, which always changes creating a pretty big look until eventually my chart builder breaks. Can somebody help with it @jlmitch5 @h-kataria ?
> 
> Also I realized that the useRequest wont work like this:
> 
> ```ts
> const { request: setData, ...api } = useRequest(
>     useCallback(
>       async () => await readData({ params: queryParams }),
>       [queryParams]
>     ),
>     {} as ApiJson
>   );
> ```
> 
> For some weird reason it works much better like this:
> 
> ```ts
> const { request: setData, ...api } = useRequest(
>     useCallback(async () => {
>       const result = await readData({ params: queryParams });
>       return result;
>     }, [queryParams]),
>     {} as ApiJson
>   );
> ```
> 
> And I have no idea why. In its core it is the same, just not saving the result in an intermediary variable. It would be nice if somebody could explain to me, it is bugging me.
> 
> Another note: we should make the qs up with Typescript, I just had to add because of it (and the querry params) like 7 eslint disabling rules and it is still hacky. This way typescript worth nothing.

I have implemented it this way in all the files
```
const { request: setData, ...api } = useRequest(
    useCallback(async () => {
      const result = await readData({ params: queryParams });
      return result;
    }, [queryParams]),
    {} as ApiJson
  );
```
had to return the result explicitly.
Also, you need to set QsConfig outside the component, otherwise, it will keep on looping thru that, why do you have to set that inside the component

brumik:
@h-kataria Yes, I seen it, but I do not understand why it is working like that. It would be nice to understand so in the future we can use it more effectively.

On the qs, I cannot set it outside the component while the default params are coming from a prop. Is there a solution to this?

knightsean00 (approved):

src/Api/types.ts
```diff
@@ -1,11 +1,14 @@
-export type Params = Record<string, string | number>;
-export type ApiJson = Record<string, string | Record<string, unknown>>;
+export type Params = Record<
+  string,
+  string | number | string[] | number[] | boolean
+>;
+export type ApiJson = Record<string, unknown>;
```
brumik, line 5: Is this accidental, or you needed to make the type more unknown than before?

src/Containers/Reports/Shared/schemas/affected_hosts_by_playbook.ts
```diff
@@ -0,0 +1,108 @@
...
+const schemaFnc = (label: string, y: string): ChartSchemaElement[] => [
+  {
+    id: 1,
+    kind: ChartKind.wrapper,
+    type: ChartTopLevelType.chart,
+    parent: null,
+    props: {
+      height: 400,
+      padding: {
+        top: 10,
```
brumik, line 46: You can instead of remove margin from the toolbar_content (Report:55) remove all the top padding on all of the charts, which I think is nicer, since we don't have to custom style something :)

src/FeatureFlags/Provider.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useState, useEffect, FunctionComponent } from 'react';
 import PropTypes from 'prop-types';
 import { Provider } from './Context';
-import { getFeatures } from '../Api';
+import { getFeatures } from '../Api/';
```
brumik, line 4: :+1:

brumik (commented):
Looks good, it is a great work you done here :clap: 

(It is my PR so I can only comment.)

**Note**: I did not pulled the reports to click trough

PR #527, opened 2021-07-12
Replaced use of useApi hook with useRequest

h-kataria:
- Also introduced qs hook, so users can now be able to save the url and load the page using the params in url, page reloads if user makes a change to url or change params from search/filters toolbars

Fixes:
https://issues.redhat.com/browse/AA-145
https://issues.redhat.com/browse/AA-147
https://issues.redhat.com/browse/AA-157


brumik (changes requested):
There are lots of linting problems which I believe can be fixed automatically (I am not sure if it fixes unused imports and variables).

I was able to go trough the first 10 files out of 27 as this PR is huuuuge :clap: :+1: (I bow to your endurance @h-kataria):
* On the automation page as I commented bellow the useRequest is now used to fetch the API even when we want to recalculate the data locally (I believe the useRequest has something for this use case too :)
* Automation calculator page has no chips, but the url shows the query params, which would be a little odd for me as an user
* The URL exposes the `attributes` param from the queryParams which should be hardcoded for most of our cases and this way the user can modify it, causing some pages or later reports to break (eg removing an attr which is displayed in a chart) 
* Modifying anything in the URL to incorrect value causes the API to fail (@jlmitch5 because of this exposition as a raw serialized string I am not sure the URL as the source of truth is the right approach, they can modify an object which is used for API calls directly from the URL to ANY value, they can add fields, or change any values).

As my feelings about this: I can understand the reason why we want to move towards useRequest since it is used in other places, but I feel it is just a little too complicated and more code as well.

As I mentioned above I have concerns about the URL as the single source of truth. However the code for synching could be improved too. I think we could isolate more the code pieces responsible for the queryParams -- url sync, since now it weaves trough all of the application (even touching the toolbar which should be a standalone, reusable component).

On that note: if the query params is in the URL bar it means it is global for the page so we could utilize a react context or redux (already in app too) to make the queryParams a true global state, and then just automagically sync in the background with the URL (one way or another)

src/Utilities/helpers.js
```diff
@@ -137,3 +138,21 @@ export function formatJobDetailsURL(baseURL, jobId) {
   const subDirectory2 = 'details';
   return `${baseURL}/#/${subDirectory1}/${jobId}/${subDirectory2}/`;
 }
+
+export const qsToObject = (qs) => {
```
jlmitch5, line 142: I would prefer these be in a qs utility rather than this catch all utilities file (which honestly, I'd eventually prefer to split apart move all this stuff to either more explicit helper files, or to shared utilities within a screen if they are only used in one place.

jlmitch5 (changes requested):
Thanks for taking this on!  Looks great generally!

My one higher level question is about queryParams.  I believe this is using the useQueryParams hook, and tower uses the qs service.  I'm curious how the approaches differ and if that will lead to any ux discrepancies across the platform.

Otherwise, just a small thing I've seen so far

jlmitch5:
@h-kataria @brumik I've filed a follow-up issue for the refactor work we discussed in the sync-up today

src/Components/Toolbar/Toolbar.js
```diff
@@ -38,11 +48,44 @@ const FilterableToolbar = ({
       return obj;
     }, {});
 
+  const handleSearch = (key, value) => {
+    let params = parseQueryString(qsConfig, history.location.search);
+    params = replaceParams(params, { [key]: value });
```
jlmitch5, line 53: again, I think you can get rid of this replaceParams line (see comment in helpers)

jlmitch5 (changes requested):
I looked over the more infrastructure-related code pretty thoroughly and then skimmed through all the implementations in the screens files.  I think this is great!  Just a few small clean up things and I think this is ready for merge.

src/Utilities/helpers.js
```diff
@@ -137,3 +139,32 @@ export function formatJobDetailsURL(baseURL, jobId) {
...
+//   params = mergeParams(params, { [key]: value });
+//   //params = replaceParams(params, { page: 1 });
+//   if (value === '' || value.length === 0)
+//     params = removeParams(qsConfig, params, {[key]: params[key]});
+//   pushHistoryState(params, qsConfig);
+// }
+
+export const handleSearch = (key, value, qsConfig, history) => {
+  let params = parseQueryString(qsConfig, history.location.search);
+  params = replaceParams(params, { [key]: value });
```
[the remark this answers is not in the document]
h-kataria: this line i believe is used, when selecting quick_date_range value, that is a single select.

jlmitch5 (approved):
Looks good!

brumik (approved):

brumik (changes requested):
Working on the ROI update bug

brumik (approved):
Everything is working

PR #528, opened 2021-07-13
Fix invalid hook call.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-15238

Causes chrome left navigation to fail.


brumik (approved):

jlmitch5 (commented):
thanks!

PR #535, opened 2021-07-22
[AA-177] Add column sorting to JobExplorer table

knightsean00:
https://issues.redhat.com/browse/AA-177


src/Components/JobExplorerList.js
```diff
@@ -1,20 +1,86 @@
...
 import PropTypes from 'prop-types';
 
 import { TableComposable, Thead, Tbody, Tr, Th } from '@patternfly/react-table';
 import JobExplorerListRow from './JobExplorerListRow';
 
 const cols = ['Id/Name', 'Status', 'Cluster', 'Organization', 'Type'];
 
-const JobExplorerList = ({ jobs }) => {
+const sortMap = {
+  'Id/Name': 'id',
```
brumik, line 10: Using a capitalized key in objects is not too good practice. If it is easy to change up the key - value pair or de-capitalize them in this object I think we should do it (if it is not easy we can leave it like this)

brumik (changes requested):
Would you consider moving the sort attribute names into a map like in the [Reports page](https://github.com/RedHatInsights/tower-analytics-frontend/pull/524/files#diff-41188b63e584a595b8a63e3a2a38a578a151cb62206f6468305d353a6474d22eR192) (it is not merged yet).

I wrote a small logic around it operating on the API format (array of `{ key: 'name', value: 'Id/Name'}`) so we can use it with the API options results too :) I think it  would be beneficial if we could later on extract this logic and just apply to all current and future tables :)

src/Components/JobExplorerList.js
jlmitch5: It does seem like I'd expect this to be inverted, as key: value, usually == key: label

brumik (changes requested):
The code looks good, but when pressing sorting, the table is not changed to loading component, but the loading is rendered above the table, causing the page to flicker on fast network and on slow networks it is confusing to see the loading component and the table too.

brumik (approved):

PR #536, opened 2021-07-22
[AA-178] Add Breakdown bars to JobExplorer Table

knightsean00:
https://issues.redhat.com/browse/AA-178

For some reason, I have not seen an example of what `most_failed_task` would look like in the production data so I have yet to make Breakdown bars for those yet. If someone (@brumik , @h-kataria , @jlmitch5 ) has any idea, please let me know so that I can complete this section.

image


src/Components/Breakdown.js
```diff
@@ -116,8 +123,9 @@ const Breakdown = ({ categoryCount = {}, categoryColor }) => {
 };
 
 Breakdown.propTypes = {
-  categoryColor: PropTypes.object,
-  categoryCount: PropTypes.object,
+  categoryColor: PropTypes.object.isRequired,
+  categoryCount: PropTypes.object.isRequired,
+  showPercent: PropTypes.bool,
```
brumik, line 128: If it is not required then please provide a default (`false`) default value to it. It will evaluate the same (as undefined and null is false in a condition too) but by providing a default value we know that it is a false boolean. It is a nice practice and also helps when converting to TS :)

src/Components/JobExplorerListRow.js
```diff
@@ -5,19 +5,125 @@ import { Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
...
 import { formatDateTime, formatJobType } from '../Utilities/helpers';
 import JobStatus from './JobStatus';
+import Breakdown from './Breakdown';
+import {
+  global_palette_green_300,
+  global_palette_black_850,
+  global_palette_gold_300,
+  global_palette_red_100,
+  global_palette_blue_300,
+} from '@patternfly/react-tokens';
```
brumik, line 15: This is great :+1: was it provided to you or you found out yourself? We should consider to extract these colors somewhere since we are using them in the icons too: ([`JobStatus.js:15`](https://github.com/RedHatInsights/tower-analytics-frontend/blob/devel/src/Components/JobStatus.js#L15))

src/Components/JobExplorerListRow.js
```diff
@@ -5,19 +5,125 @@ import { Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
...
 
+const renderFailedTaskBar = (failed_tasks) => {
+  const [more, setMore] = useState(false);
+  const categoryColor = {
+    passed: global_palette_green_300.value,
+    unreachable: global_palette_black_850.value,
+    failed: global_palette_red_100.value,
+  };
+
+  if (failed_tasks != null) {
```
brumik, line 36: Can we break up this multiple conditional return statements to more readable sub functions (or components if it looks like it could be extracted into its own component - basically if you can name it reasonably it can be probably a component).

src/Components/JobExplorerListRow.js
```diff
@@ -5,19 +5,125 @@ import { Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
...
+                  fullWidth={{ default: 'fullWidth' }}
+                >
+                  {more ? 'Show less' : 'Show more'}
+                </Button>
+              </FlexItem>
+            </Flex>
+            <br></br>
+          </>
+        ) : (
+          <React.Fragment />
```
brumik, line 92: Please use the `<>` shorthand for `<React.Fragment>`. If you want to return nothing from the component please use `null` or `''` (empty string) - TS prefers empty string :). Since the react fragment does create a DOM element (empty div pair) while the `''` or 'null` return do not create any DOM elements.

src/Components/JobExplorerListRow.js
```diff
@@ -5,19 +5,125 @@ import { Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
...
+
 const JobExplorerListRow = ({ job }) => {
   const [expanded, setExpanded] = useState(false);
   const handleExpansion = () => {
     setExpanded(!expanded);
   };
 
+  const categoryCount = job
+    ? {
+        ok: job?.ok_host_count ? job.ok_host_count : 0,
```
brumik, line 109: You can use the shorthand `??`. Basically:
```js
let job = { ok: 10 };
job?.ok ?? 0 // 10
job?.xx ?? 0 // 0
// BEWARE: 
job?.ok || 0 // 0
```

Basically `??` checks if it is returned null or undefined, while `||` converts it to boolean and check is then for true/false.

I think when the `job` is not defined we could return the object with all the fields set to `0` as in the check for the individual fields.

src/Components/JobExplorerListRow.js
```diff
@@ -5,19 +5,125 @@ import { Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
...
+        skipped: job?.skipped_host_count ? job.skipped_host_count : 0,
+        changed: job?.changed_host_count ? job.changed_host_count : 0,
+        failed: job?.failed_host_count ? job.failed_host_count : 0,
+        unreachable: job?.unreachable_host_count
+          ? job.unreachable_host_count
+          : 0,
+      }
+    : null;
+
+  const categoryColor = {
```
brumik, line 119: We already have a color map defined at the start of the file. Can we merge the two?

src/Components/JobExplorerListRow.js
```diff
@@ -31,7 +137,12 @@ const JobExplorerListRow = ({ job }) => {
       label: 'Finished',
       value: job.created ? formatDateTime(job.finished) : 'Unavailable',
     },
+    {
+      label: 'Tasks',
+      value: job.host_task_count ? job.host_task_count : 0,
```
brumik, line 142: Same as before with the `??`

brumik (changes requested):
It is looking good :) 

We are doing so much checking for the `job` having certain values. We could just provide a default value to it and forget about checking maybe?

src/Components/JobExplorerListRow.js
jlmitch5: I wonder if there's a "prefer <> format for fragments" linting rule we could add?

src/Components/JobExplorerListRow.js
knightsean00: Not gonna lie I just kind of guessed those colors lol. It would be nice if Taufique could give some input on these color choices.

src/Components/Breakdown.js
```diff
@@ -116,8 +123,13 @@ const Breakdown = ({ categoryCount = {}, categoryColor }) => {
...
+};
+
+Breakdown.defaultProps = {
+  showPercent: false,
```
brumik, line 132: Once Ben sent me a link why `defaultProps` is not preferred over defining the default value while de-structuring the params. I cannot find the link anymore, but it broke down to performance issues. While default props is running trough the react framework, the de structured default value is created at compile time. So since then I am using the destructuring for defining default value. 

Can we move it there (params destructuring) too?

src/Components/JobExplorerListRow.js
brumik: Yeah :D maybe we can ask Taufique him to provide some guidelines or colors in the mockup or alongside of it

src/Components/JobExplorerListRow.js
```diff
@@ -5,19 +5,121 @@ import { Tr, Td, ExpandableRowContent } from '@patternfly/react-table';
...
+            onClick={() => {
+              setShowMore(!showMore);
+            }}
+            fullWidth={{ default: 'fullWidth' }}
+          >
+            {showMore ? 'Show less' : 'Show more'}
+          </Button>
+        </FlexItem>
+      </Flex>
+      <br></br>
```
brumik, line 53: Is this by accident or you want to render two empty lines? (<br> is a simple html tag without closing pair so </br> also renders a empty line). If you want space maybe you can use css margin to make it happen instead? Then we could remove the br and the react fragments.

brumik (approved):
This is looking good already. Please update the default params and it is good to go ^^.

PR #537, opened 2021-07-22
[AA-636] Fix SavingsPlanner Validation Errors

knightsean00:
https://issues.redhat.com/browse/AA-636

So far, only the issue with negative numbers in the run_on_hosts field has been fixed. I am still talking with @jlmitch5 about what should be done with duplicate tasks.


jlmitch5 (approved):
Looks good

PR #538, opened 2021-07-23
- split useRequest requests into individual requests

h-kataria:
- And fixed to load correct contents in modal when template/wokflow is selected

@knightsean00 this addresses the bug you found. @jlmitch5 please review.

### DIFF
--- a/src/Components/ModalContents.js
+++ b/src/Components/ModalContents.js
@@ -75,32 +75,28 @@ const formatTotalTime = (elapsed) =>
 
 const ModalContents = ({ selectedId, isOpen, handleModal, qp, jobType }) => {
   const {
-    result: { relatedJobs, stats },
-    error,
-    isLoading,
-    isSuccess,
-    request: fetchEndpoints,
+    result: stats,
+    request: fetchStats,
+    ...statsApi
   } = useRequest(
-    useCallback(async () => {
-      const [stats, relatedJobs] = await Promise.all([
-        readJobExplorer({ params: agreggateTemplateParams }),
-        readJobExplorer({ params: relatedTemplateJobsParams }),
-      ]);
-      return {
-        relatedJobs: relatedJobs.items,
-        stats: stats.items[0],
-      };
-    }, []),
-    {
-      stats: {},
-      relatedJobs: {},
-    }
+    useCallback(
+      () => readJobExplorer({ params: agreggateTemplateParams }),
+      [selectedId]
+    ),
+    {}
   );
 
-  useEffect(() => {
-    fetchEndpoints();
-  }, [selectedId, fetchEndpoints]);
-
+  const {
+    result: relatedJobs,
+    request: fetchJobs,
+    ...jobsApi
+  } = useRequest(
+    useCallback(
+      () => readJobExplorer({ params: relatedTemplateJobsParams }),
+      [selectedId]
+    ),
+    {}
+  );
   let history = useHistory();
 
   const redirectToJobExplorer = () => {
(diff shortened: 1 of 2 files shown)


src/Components/ModalContents.js
```diff
@@ -75,32 +75,32 @@ const formatTotalTime = (elapsed) =>
...
-      };
-    }, []),
-    {
-      stats: {},
-      relatedJobs: {},
-    }
+      const relatedJobs = await readJobExplorer({ params: relatedTemplateJobsParams })
+      return { relatedJobs: relatedJobs.items };
+    }, [selectedId]),
+    { relatedJobs: {} }
```
brumik, line 102: Can we return just the items we need and not wrap everything in another object. I know that you used this pattern everywhere but it is just adding an unnecessary wrapper object what we de-structure nearly everywhere instantly.

src/Components/ModalContents.js
```diff
@@ -75,32 +75,32 @@ const formatTotalTime = (elapsed) =>
...
+    error: statsError,
+    isLoading: statsIsLoading,
+    isSuccess: statsIsSuccess,
+    request: fetchStats,
+  } = useRequest(
+    useCallback(async () => {
+      const stats = await readJobExplorer({ params: agreggateTemplateParams })
+      return { stats: stats.items[0] };
+    }, [selectedId]),
+    { stats: {}, statsError, statsIsLoading, statsIsSuccess }
```
brumik, line 88: This default value I believe is for the `result` variable. You are also using `variables` for defining it which is not bad, but yo uare using variables which are returned to you from this function so they are "used before defined". In JS you can do this, but the value of the variable will be `undefined` and it is considered a bad practice as it can cause confusion.

src/Components/ModalContents.js
```diff
@@ -75,32 +75,32 @@ const formatTotalTime = (elapsed) =>
...
-    result: { relatedJobs, stats },
+    result: stats,
+    error: statsError,
+    isLoading: statsIsLoading,
+    isSuccess: statsIsSuccess,
+    request: fetchStats,
+  } = useRequest(
+    useCallback(async () => {
+      const stats = await readJobExplorer({ params: agreggateTemplateParams })
+      return { stats: stats.items[0] };
```
brumik, line 86: I am not sure, but you may want the aggregate values:

Which you can get more accurately from the `apiData.meta.legend` as it is now returned in every call. Please consider looking into what is needed and if you can get it from the meta. The `items` array elements may not have always the same items (and the others is not calculated correctly).

brumik (changes requested):
Please fix linting, and tests.

Also it does not seem to work, anytime I open a template modal I get the `No template name` and empty modal body.

h-kataria:
@brumik please re-review

src/Components/ModalContents.js
```diff
@@ -201,22 +202,26 @@ const ModalContents = ({ selectedId, isOpen, handleModal, qp, jobType }) => {
     waiting: global_palette_light_green_200.value,
   };
 
-  const descriptionStats = stats && [
+  const descriptionStats = stats.items && [
     {
       label: 'Number of runs',
       id: 'total-runs',
-      value: stats.total_count ? stats.total_count : 'Unavailable',
+      value: stats.items[0].total_count
```
brumik, line 209: Here we could use the `??` operator (`stats.items[0]?.total_count ?? 'Unavaiable`). The `??` chceks for null or undefined and returns the value if it defined otherwise it returns the param on the right side (Unavaiable here)

src/Components/ModalContents.js
```diff
@@ -248,66 +256,81 @@ const ModalContents = ({ selectedId, isOpen, handleModal, qp, jobType }) => {
     <Modal
       aria-label="modal"
       variant={ModalVariant.medium}
-      title={stats.name ? stats.name : 'No template name'}
+      title={
+        jobsApi.isSuccess &&
+        statsApi.isSuccess &&
+        stats?.items?.length > 0 &&
```
brumik, line 262: It is working with: `stats?.items[0]?.name ?? `No template name` too :)

brumik (changes requested):
Looks good now, I have a few typing saver tricks here :). The only issue I found is the missing bar from the top of the template:

**Cloud/beta (before)**:
Screenshot from 2021-07-26 13-26-08

**After:**
Screenshot from 2021-07-26 13-26-02

Is this intentional?

h-kataria:
> Looks good now, I have a few typing saver tricks here :). The only issue I found is the missing bar from the top of the template:
> 
> **Cloud/beta (before)**:
> Screenshot from 2021-07-26 13-26-08
> 
> **After:**
> Screenshot from 2021-07-26 13-26-02
> 
> Is this intentional?

@brumik resolved this issue, good catch. took care of other comments as well

brumik (approved):
All looking great :+1:

PR #541, opened 2021-07-27
Added namespace prefix on params in url.

h-kataria:
params need to be prefixed by namespace when building a link to redirect to job explorer from various screens.

@jlmitch5 please review, this should fix the issue you saw with the link to template from Savings-planner list/details view


jlmitch5 (approved):
awesome!  this looks good.  In the future we may want to look at passing the QS_CONFIG to get/set query params in a way that auto appends/strips these types of prefixes (and keeps them from potentially drifting if we do something like change the prefix.

jlmitch5 (approved):
Awesome, this looks great!

PR #542, opened 2021-07-28
Updated use request to ts with small optimizations

brumik:
* Updated the useRequest to typescript
* Updated the isMounted to typescript (little unsafe types but I think it will do fora while)
* Corrected some used before defined errors in the used useRequest on the Job explorer page
* Removed the unused variables from the useDeleteItems and moved the refetch of the items outside. I believe the delete items function was handling more and we stripped down to our needs and this is why some unused params were passed


h-kataria:
@brumik tested changes in UI, everything else works fine but Delete button on Savings Planner list view does not refresh the list after plan is deleted. Could not test Delete button from Plan details screen due to unrelated RBAC infinite loop issue on that screen.

src/Utilities/useRequest.ts
```diff
@@ -1,56 +1,78 @@
...
  * The hook also accepts an optional second parameter which is a default
  * value to set as result before the first time the request is made.
  */
-export const useRequest = (makeRequest, initialValue) => {
-  const [result, setResult] = useState(initialValue);
-  const [error, setError] = useState(null);
-  const [isLoading, setIsLoading] = useState(false);
-  const [isSuccess, setIsSuccess] = useState(false);
+interface UseRequestVariables<T> {
+  result: T;
```
jlmitch5, line 20: Not sure about this being called "T", usually single letter variables are harder to understand than more descriptively named ones.  Maybe it is a convention of typescript I don't know about yet though.  What are your thoughts/are there any docs on this?

src/Utilities/useRequest.ts
```diff
@@ -1,56 +1,78 @@
...
+  error: ErrorType;
+  isLoading: boolean;
+  isSuccess: boolean;
+}
+
+interface UseRequestReturn<T> extends UseRequestVariables<T> {
```
jlmitch5, line 26: These names UseRequestReturn and UseRequestVariables feel pretty imperative.  The extends here also seems like there is an inheritance structure that I don't really understand, and I think may be more confusing than it's worth.

I do think I understand the value here...code completion/knowing the params of this function when called in other places.

Maybe it would help to extract the typing out to a separate file?  Honestly, I'm not sure.

jlmitch5 (commented):
The changes to JobExplorer seem very nice!

The typescript syntax here is a bit harder for me to follow than the past files with the naming, T's and nested `<<>>`, but I don't have the knowledge off hand to know what would be better.  Let's sync-up on this, tomorrow.  And if you happen to know anyone, it could be good to share with someone on the CRC side with a lot of TS experience.

src/Utilities/useRequest.ts
brumik: I could name it something different but in TS generics are mostly named by a single letter (T or if there is multiple one for example in a map then <K, T> which stands Key(type) and Type).

You can read about generics here: https://www.typescriptlang.org/docs/handbook/2/generics.html

src/Utilities/useRequest.ts
brumik: I can copy paste the type from the above, but `extend` can do it for us, so why copy paste and make the code WET?
In TS extending types, union types and discriminatory types are common and in some cases unavoidable without typing out same things again and again.

It is up to us where we put the types, but I find easier to read single modules when the types are just right about the function, so you can have them in the view while editing the function. If the types get crazy and complicated I like to extract them in a separate `types.ts` file (which is quite standard I think) since these complicated cases should be already in multiple files.

src/Containers/SavingsPlanner/List/List.js
```diff
@@ -126,16 +125,12 @@ const List = () => {
...
-      allItemsSelected: isAllSelected,
-      fetchItems: fetchEndpoints,
-    }
+    }, [selected])
   );
 
   const handleDelete = async () => {
-    await deleteItems();
+    deleteItems();
+    fetchEndpoints();
```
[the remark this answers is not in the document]
brumik: The function (`deleteItems()` and `fetchEndpoints()`) are not async. The delete items is using the useRequest which is returning in the `request` field a `useCallback` function (which contains an async function) but itself the `useCallback` is not async.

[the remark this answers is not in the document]
brumik: Okey, discovered useCallback can return promise too. I am little confused still from all these useCallbacks nested in each other.

src/Utilities/useRequest.ts
jlmitch5: Thanks for the explanation, makes sense!

jlmitch5 (approved):
This looks good!  Thanks for explaining how all this fits together to me @brumik

PR #543, opened 2021-07-29
useRequest usage unification

brumik:
* Make the `readApi` params into a simple object instead of nesting, it makes TS much easier in the future
* Fixed the job explorer bookmarking (it was using duplicate bookmarking which worked weird)
* Unified the useRequest usage, removed unnecessary wrapping where was not needed.
* Corrected the default variables passed to the useRequest
* Unified the `preflight` usage
* Added the `Not authorized` to the `EmptyState` (which is handling all possible preflight errors, probably we should rename)
* **Fixed** the infinite rbac call on the details page

### DIFF
--- a/src/Containers/SavingsPlanner/Shared/Form/index.js
+++ b/src/Containers/SavingsPlanner/Shared/Form/index.js
@@ -46,7 +46,7 @@ const Form = ({ title, options, data = {} }) => {
         });
       }
 
-      // TODO this is very shitty should be a promise
+      // TODO this should be a promise
       return {};
     }, []),
     data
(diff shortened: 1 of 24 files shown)


src/Containers/SavingsPlanner/Shared/Form/index.js
```diff
@@ -29,30 +29,27 @@ const Form = ({ title, options, data = {} }) => {
...
+        return createPlan({
+          params: requestPayload,
+        });
       }
-      return {
-        data,
-      };
+
+      // TODO this should be a promise
+      return {};
```
jlmitch5, line 50: this could also throw an error if the payload is required (and it looks like it is)

jlmitch5 (approved):
This looks great and simplifies the way we call API functions, great job!

src/Containers/SavingsPlanner/Shared/Form/index.js
brumik: This could be in a followup PR where we check the error and the loading states. Right now I am not sure which cases the user could get to this return, so I figured we could leave it like this and later update it. As I marked in TS this won't work anyway, since the function should return Promise.

PR #551, opened 2021-08-05
Change charts to line, add domainPadding to y-axis, and tooltips to all lines

knightsean00:


brumik (approved):
LGTM - and surprised that the lib can deal so well with this setup

Btw: there is a problem with the reports but it is not part of this PR:
https://issues.redhat.com/browse/AA-678 for future reference

PR #554, opened 2021-08-06
Quality of code fixes

brumik:
Planned to be a bigger PR but creating at least with the fixes.


src/Containers/JobExplorer/JobExplorer.js
```diff
@@ -60,10 +60,6 @@ const JobExplorer = () => {
     { items: [], meta: {} }
   );
 
-  useEffect(() => {
```
h-kataria, line 63: @brumik is this piece of code here and in other modules redundant?

h-kataria (commented):
other than one question on deleted code, this looks good to me

knightsean00 (changes requested):
We should probably fix up the linting errors that GitHub raised (using variables before being defined) just so we don't have these errors in the future

src/Containers/JobExplorer/JobExplorer.js
brumik: Yes, since the navigation update when doing history.push to other pages the crc router updates, so we don't need to hack it.

brumik:
@knightsean00 I would like to have it for error in the future. As you can see I set ts to `error` and js to `warn`. This is because last times I discovered we have a lot of first use then define cases. I would fix it as we touch things. The warn is tho show where are the cases. Or I can make a followup PR which refactors all of use before define problems, but I don't think this belongs here.

h-kataria (approved):

knightsean00 (approved):

PR #555, opened 2021-08-06
Converting toolbar to TS

brumik:

PR #557, opened 2021-08-09
Added granularity to the report charts

brumik:
### Problems:
* Chart x axis labels are not too helpful when setting to year (showing 01/01 for all ticks)
  * possible solutions: display full date always or change the date depending on the granularity (hard mode)
* Quick date range from API does not change with the granularity which is then weird
* Granularity could go next to the date selector instead to the category drop down. Waiting for the #555 PR to be merged, so we don't cause accidental bugs.

(Done) TODO: 
- Granularity to the left of the date selector
- Change the chart axis depending on the granularity

### DIFF
--- a/src/Utilities/useQueryParams.js
+++ b/src/Utilities/useQueryParams.js
@@ -87,7 +87,7 @@ export const useQueryParams = (initial) => {
         return { ...state, ...value };
       case 'SET_QUICK_DATE_RANGE':
         return value !== 'custom'
-          ? { ...state, start_date: null, end_date: null }
+          ? { ...state, ...value, start_date: null, end_date: null }
           : { ...state, ...value };
       case 'SET_START_DATE':
       case 'SET_END_DATE': {
--- a/src/Utilities/useQueryParams.js
+++ b/src/Utilities/useQueryParams.js
@@ -43,6 +43,7 @@ export const useQueryParams = (initial) => {
         return isNaN(value)
           ? { ...state, offset: 0 } // Defaults back to 0
           : { ...state, offset: parseInt(value) };
+      case 'SET_GRANULARITY':
       case 'SET_ATTRIBUTES':
       case 'SET_JOB_TYPE':
       case 'SET_STATUS':
@@ -107,6 +108,7 @@ export const useQueryParams = (initial) => {
     name: 'SET_NAME',
     only_root_workflows_and_standalone_jobs: 'SET_ROOT_WORKFLOWS_AND_JOBS',
     inventory_id: 'SET_INVENTORY',
+    granularity: 'SET_GRANULARITY',
   };
 
   return {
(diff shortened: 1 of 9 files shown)


h-kataria:
LGTM, tested in UI. @jlmitch5 i defer it to you for approval

brumik:
This is a Draft PR. I need feedback on the description. Do not merge yet.

src/Utilities/useQueryParams.js
```diff
@@ -43,6 +43,33 @@ export const useQueryParams = (initial) => {
...
+            return {
+              ...state,
+              ...value,
+              quick_date_range: 'last_6_months',
+            };
+          case 'yearly':
+            return {
+              ...state,
+              ...value,
+              quick_date_range: 'last_6_years',
```
jlmitch5, line 64: @trahman73 what would you say a good default for the yearly granular setting would be? Are options are Past 1, 3, 6 and 10 years

src/Utilities/useQueryParams.js
```diff
@@ -58,16 +85,10 @@ export const useQueryParams = (initial) => {
...
-        if (value !== 'custom') {
-          newState = { ...newState, start_date: null, end_date: null };
-        }
-
-        newState = { ...newState, ...value };
-        return newState;
-      }
-
+      case 'SET_QUICK_DATE_RANGE':
+        return value !== 'custom'
```
jlmitch5, line 89: is this just a bug fix?  or is it related to the granularity?

jlmitch5 (commented):
I pulled this down and it looks great!

My only question is if we should do a different default for yearly granularity.  I'm curious what @trahman73 thinks.

Also I wasn't quite sure about the useQueryParams change for the quick date range setting...was that just fixing a bug?

src/Utilities/useQueryParams.js
trahman73: I would say Past 1 as the default.

brumik: @trahman73 I don't think that would be wise, that is one tick which means no lines in the chart just dots :) so it is pretty useless for the user.

src/Utilities/useQueryParams.js
brumik: I had some weird bug there and I don't know why it got fixed by removing the intermediate `let` variable. Slightly worrying that I don't know the reason but tested it and works so fine by me.

src/Utilities/useQueryParams.js
trahman73: @brumik Ah, I didn't realize that would result in only a single tick. Then I would say 6 or even 10 if it fits comfortably in the view.

brumik: I set 6 as a good middle ground (I tried to set the second biggest for each granularity for consistency)

h-kataria (approved):

PR #559, opened 2021-08-09
Cleanup and convert Utilities to TS

brumik:
* Changed authorization endpoint to `v1` - (I tried to check all HTTP errors but could not test the authorization endpoint fully)
* Unified error messages from the endpoints,
* Moved stuff from Utilities which are specific to pages.
* Converted remaining utilities to TS except the ones I touched in another PR to not cause discrepancy in functionality (renaming makes impossible to merge changes)

**Each change is in its own commit**


h-kataria:
@brumik when i change something in the date drop down on any of the pages in toolbar, it no longer adds that to the url querystring. in the screenshot below, it shows other applied filters in url but not the date filter for past 62 days.
Screen Shot 2021-08-10 at 10 51 33 AM

brumik:
@h-kataria this issue is not created in this PR as it is the same on the cloud beta. I think it may have been introduced with the fix of sort_by discrepancy. I will create a Jira issue and fix it in another PR. This PR is good to if that is the only issue with it as it is not caused it.

h-kataria (approved):

PR #564, opened 2021-08-12
Update URL-Parse package

knightsean00:


jlmitch5 (approved):

brumik:
@jlmitch5 Was this a hotfix? If not we should rather update the package which uses the URL parse package than adding it to the `dependencies`. Or if we cannot do it, just make the update for the time being in the `package-lock` file. This solution requires us to update more packages in the future, which I think we don't want :)

PR #565, opened 2021-08-12
Add tracking tooltip for all points on Reports page

knightsean00:


brumik (approved):
Screenshot from 2021-08-16 09-57-11

Works, looks good.

PR #566, opened 2021-08-12
Add case for empty organization name for Organization Statistics chart

knightsean00:


brumik (approved):
I don't know how this bug comes up time to time :D. It feels like I fixed it 3-4 times, but there it is.... still coming back :D 

Thanks ^^

PR #567, opened 2021-08-13
Update Savings Planner to handle new API outputs

knightsean00:
Old Savings Planner looked for `options.data.{property_here}` but the new output is instead `options.{property_here}`. Also updated default params for Template Selection stage of Savings Planner.


brumik (approved):
LGTM

PR #568, opened 2021-08-16
Fix Login Flow for cypress tests

one-t:
This PR adds the intermediate "Next" click after entering the username to enable password entry.


brumik (changes requested):
I am not sure why we would need this. We are running cypress on standalone with keycloak so both the username and password fields are showing right from the start.

PR #571, opened 2021-08-19
AA-145 - Store Query Parameters in url

one-t:
This PR adds some basic tests for query parameter storage in the URL. There is one outstanding issue at the moment, which is whether or not jobtype should be retained there since it currently is not.


brumik:
@one-t Every filter should be in the URL. Is there an issue created for it in Jira? If not can you create and in every case assign to me? I am already investigating some weird behavior in the URL bookmarks.

brumik (approved):

PR #579, opened 2021-09-01
Converted Notifications to TSX

h-kataria:
Converted Notifications to use useRequest and cleaned up code that was not being used

https://issues.redhat.com/browse/AA-706

Thanks a lot to @brumik for big help with this conversion.


brumik:
Can you replace the `Notification.js` to `Notification` in the tests and make them to pass :). Hopefully it is less painfull :D

PR #581, opened 2021-09-02
Updated some packages minor versions

brumik:
Also removed the `url-parse` package from the package.json. We are not dependent on it, it is just a sub dependency of our dev packages. It got there a few weeks ago because of a vulnerability, but now it should be fine. If it persists we should lock its version in the package-lock file instead or update the parent package.


jlmitch5 (commented):
llet's hold this until after the prod merge

PR #583, opened 2021-09-07
[AA-660] Api calls error and loading handling - step 1/4

brumik:
@h-kataria @jlmitch5 @skochay330 

I would be happy if you would check if you like this format of API handling and to see if you can break it somehow or if it introduces some bugs. I think this is the first critical part, since I tried to make a component which can handle errors.

Followup if this is good
* Add in a followup PR the handling of RBAC errors too for the savings planner
* Review the PR from the previous step thoroughly - like this PR.
* Update in one big swoop all the components to this and to the RBAC template
* Sleep calmly knowing that we are handling loading and errors from API correctly trough generalized components :)

Hoping that with splitting up the refactor (can I call this a refactor?) into small pieces we can spot bugs faster and make reviews faster resulting in better product quality.

### DIFF
--- a/src/Components/ApiStatus/ApiErrorState.tsx
+++ b/src/Components/ApiStatus/ApiErrorState.tsx
@@ -10,7 +10,7 @@ import { ExclamationCircleIcon } from '@patternfly/react-icons';
 import { global_danger_color_200 as globalDangerColor200 } from '@patternfly/react-tokens';
 
 interface Props {
-  message: string | Record<string, any>;
+  message: string;
 }
 
 const ApiErrorState: FunctionComponent<Props> = ({ message }) => (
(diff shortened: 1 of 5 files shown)


jlmitch5:
Let's close this for now.  I think we've done a bit too much on refactoring things related to the api requests, and we need to let the code base settle with the query param changes that are going in and we can revisit.

brumik:
This is the smallest "refactor" we had for a long time. It is rather a bugfix. I just added two components which can handle all the api errors and put a wrapper around components that uses them. 

If I remove the Job explorer page changes, can we get at least those two files in? It took a long time and Ladas to figure out what API error can we get and how to handle it for data and options endpoints.

jlmitch5:
👍  that sounds good

brumik:
@jlmitch5 @h-kataria I opened this PR in September to introduce a new more consistent and fail proof format of API handling on pages. In 9 month I rebased it once and I remember explaining and talking about it with John (who approved the motiv). I think at this point we could potentially throw away this initiative. The product is moving into less pages - more reports and what is in the codebase is still working just right, even if it is not the most consistent. 

If there is still interest, I am happy to spend some time with the team/on it too.

h-kataria:
Assigned Jira task to @ZitaNemeckova for now, we will take a look at this at a later time. Closing this PR for now.

PR #586, opened 2021-09-08
404 page for root urls

brumik:
> Currently we are redirecting to the clusters page if a top level link is not found in our routes. We should display the 404 page instead as this will be confusing to users especially if we remove a page (for example the orgs page)

Screenshot from 2021-09-08 16-56-33

### DIFF
--- a/src/Routes.js
+++ b/src/Routes.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import asyncComponent from './Utilities/asyncComponent';
 import { Paths } from './paths';
+import Error404 from './Components/Error404';
 
 const components = {
   clusters: asyncComponent(() => import('./Containers/Clusters/Clusters')),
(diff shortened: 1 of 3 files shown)


src/Routes.js
```diff
@@ -59,9 +60,9 @@ export const Routes = () => {
           rootClass={key}
         />
       ))}
-      {/* Finally, catch all unmatched routes and redirect to Clusters page */}
+      {/* Finally, catch all unmatched routes and render 404 */}
       <Route>
-        <Redirect to={Paths.clusters} />
+        <Error404 body="Sorry, we could not find what you were looking for. The page you requested may had been changed or moved." />
```
h-kataria, line 65: message should be `"Sorry, we could not find what you were looking for. The page you requested may have been changed or moved." ` changed `had` to `have`

brumik:
Change title to: 404: Page does not exist.

PR #591, opened 2021-09-14
Added a new report fo Templates Explorer view

h-kataria:
https://issues.redhat.com/browse/AA-196

Backend changes are in [PR](https://gitlab.cee.redhat.com/automation-analytics/automation-analytics-backend/-/merge_requests/240)

Screen Shot 2021-09-14 at 4 14 32 PM

Screen Shot 2021-09-14 at 4 15 05 PM

Screen Shot 2021-09-20 at 1 27 46 PM


src/Containers/SavingsPlanner/Details/StatisticsTab/index.tsx
```diff
@@ -26,7 +26,7 @@ import ChartRenderer, {
   ChartTopLevelType,
   ChartSchema,
   ApiType,
-  NonGroupedApi,
+  NonGroupedApi, ChartLegendOrientation, ChartLegendPosition,
```
brumik, line 29: This seems like an accident

src/Containers/Reports/Shared/schemas/templates_explorer.ts
```diff
@@ -0,0 +1,145 @@
...
+import { AttributesType, ReportPageParams } from '../types';
+
+const name = 'Templates Explorer';
+
+const expandRows = true;
+
+const description =
+  'The number of times a job template has ran in a specified time window.\n\nYou can use this report to be able to tell which playbooks are running most frequently, allowing you to see which groups in your organization are running Ansible the most.';
+const categories = [CATEGORIES.executive];
+const list_attributes = ['failed_count', 'successful_count', 'total_count']
```
brumik, line 23: Can we use camelCases here instead of snake_case? JS is by default camelCased and we are just using snake_case for the API parameters

src/Containers/Reports/Shared/schemas/index.ts
```diff
@@ -2,11 +2,13 @@ import { ReportPageParams } from '../types';
 import affected_hosts_by_playbook from './affected_hosts_by_playbook';
 import changes_made from './changes_made';
 import playbook_run_rate from './playbook_run_rate';
+import templates_explorer from './templates_explorer';
 
 const reports = [
   affected_hosts_by_playbook,
   changes_made,
   playbook_run_rate,
+  templates_explorer,
```
brumik, line 11: This has changed, probably need to rebase, it should be an easy merge :)

src/Containers/Reports/Shared/schemas/affected_hosts_by_playbook.ts
```diff
@@ -12,7 +12,6 @@ import { CATEGORIES } from '../constants';
 import { AttributesType, ReportPageParams } from '../types';
 
 const name = 'Hosts changed by job template';
-
```
brumik, line 15: Accident? Probably creating conflicts :)

src/Containers/Reports/Details/Report.tsx
```diff
@@ -285,4 +471,9 @@ Report.propTypes = {
   schemaFnc: PropTypes.func.isRequired,
 };
 
+// Specifies the default values for props:
+Report.defaultProps = {
```
brumik, line 475: Please don't use defaultProps, instead define the value at de-structuring it. It is much more effective and readable.

src/Containers/Reports/Details/Report.tsx
```diff
@@ -274,6 +459,7 @@ const Report: FunctionComponent<ReportGeneratorParams> = ({
 
 Report.propTypes = {
```
brumik, line 460: We don't really need `propTypes` anymore either. Could you please remove it in this PR since you are already editing this file. I changed the rule for TS files to not to require it a while ago

src/Containers/Reports/Details/Report.tsx
```diff
@@ -212,7 +391,7 @@ const Report: FunctionComponent<ReportGeneratorParams> = ({
     const whitelistKeys = options?.sort_options?.map(
       ({ key }: { key: string }) => key
     );
-    if (!whitelistKeys.includes(currKey)) return {};
+    if (!whitelistKeys?.includes(currKey)) return {};
```
brumik, line 394: :+1:

src/Containers/Reports/Details/Report.tsx
```diff
@@ -182,7 +358,10 @@ const Report: FunctionComponent<ReportGeneratorParams> = ({
 
   const [attrPairs, setAttrPairs] = useState<AttributesType>([]);
   useEffect(() => {
-    if (options.sort_options) {
+    if (defaultParams.list_attributes && options.sort_options) {
+      const attrsList = options.sort_options.map((el) => { if(defaultParams.list_attributes.some(a => el.key.indexOf(a) >= 0 )) return el})
+      setAttrPairs([...extraAttributes, ...attrsList.filter(Boolean)]);
```
brumik, line 363: Sorry, I could not figure out these two rows, could you explain what are they doing?

Using the `filter` function with a plain JS type (`Boolean`) seems pretty unusual for me and never seen before, what is the intent?

src/Containers/Reports/Shared/schemas/templates_explorer.ts
```diff
@@ -0,0 +1,145 @@
...
+    'ok_host_count',
+    'skipped_host_count',
+    'total_count',
+  ]
+
+const defaultParams: Params = {
+  limit: 6,
+  offset: 0,
+  attributes: [...list_attributes, ...expanded_attributes],
+  list_attributes: list_attributes,
```
brumik, line 46: The new endpoint supports `list_attributes` and `expanded_attributes`?

Also you can just shorthand here the `list_` and `expanded_attributes`.

src/Containers/Reports/Details/Report.tsx
```diff
@@ -117,10 +135,117 @@ const getDateFormatByGranularity = (granularity: string): string => {
   if (granularity === 'daily') return 'formatDateAsDayMonth';
 };
 
-const renderData = (dataApi, chartSchema, attrPairs, getSorParams) => {
+const categoryColor = {
```
brumik, line 138: We were talking about this before I think, and we could extract this to some `color constants` file in the utilities if it is not already there. We need it in few places already (2 or 3 I think).

brumik (dismissed):
The progress looks good, I left some code quality comments bellow.

What I think could be important to make the Reports readable, because we are starting to create a monster here (already hitting the 500 lines length while the recommended is up to 200 loc/component)

I would suggest to break off the `Table` as a whole into its own `ReportTable` component (folder where index is the component). Inside the table I would break off 3 smaller components: `TableHeader`, `TableRow`, `TableExpandedRow`. They should be fairly easy components to make them in TS too.

I would start the next changes with a rebase to devel :)

src/Containers/Reports/Shared/schemas/templates_explorer.ts
h-kataria: These are not used for the endpoint, I am defining these separately as I need to only use a few of them in the Reports list view and rest of the attributes will be used in the expanded view section of the report. Currently reports.tsx displays all the columns that are in the `attributes` array. in order to get all the data in single API call, I am dividing the attributes into variables that can be used to show data is different section of the report

src/Containers/Reports/Details/Report.tsx
h-kataria: filter function here is eliminating the entries that are null or undefined from the `attrList` array
These lines are checking if list_attributes is defined then just show those attributes in the list view instead of showing all of the attributes because other attributes are being used in the expanded section.

src/Containers/Reports/Shared/schemas/templates_explorer.ts
brumik: Then we could pass them maybe as a prop? Since the place you defined them is sent to the API request as it is. Right now it should not pose any issue but in the future it may, and also makes it a little confusing.

src/Containers/Reports/Details/Report.tsx
brumik: So if I understand it correctly you are doing something like this?:
```js
const attrList = options.sort_options.filter(({ key }) => defaultParams.list_attributes.includes(key));
```

src/Containers/Reports/Shared/types.ts
```diff
@@ -10,7 +10,9 @@ export type SchemaFnc = (
 
 export interface ReportGeneratorParams {
   defaultParams: Params;
+  expandRows?: boolean;
   extraAttributes: AttributesType;
+  listAttributes?: Array<any>,
```
brumik, line 15: `listAttributes: string[]`

src/Containers/Reports/Shared/schemas/index.ts
```diff
@@ -14,7 +15,7 @@ const defaultReport: ReportPageParams = {
   categories: [] as string[],
 };
 
-const flaggedReports = [hostsByOrganization, jobsTasksByOrganization];
+const flaggedReports = [hostsByOrganization, jobsTasksByOrganization, templates_explorer];
```
brumik, line 18: This puts the template explorer report in one bucked under the `orgRepors` flag. This is what we want or should we create another flag for this report?

jlmitch5 (approved):

PR #593, opened 2021-09-15
Make Filterable Toolbar Responsive

skochay330:
Made the filterable toolbar responsive. The filter icon now only appears when window is minimized to a specific width (previously, it served no function). 

Before:
Screen Shot 2021-09-15 at 2 35 45 PM

After:
Screen Shot 2021-09-15 at 2 36 00 PM

After (when filter icon is clicked):
Screen Shot 2021-09-15 at 2 36 13 PM

### DIFF
--- a/src/Components/Toolbar/Toolbar.tsx
+++ b/src/Components/Toolbar/Toolbar.tsx
@@ -7,6 +7,7 @@ import {
   ToolbarItem,
   ToolbarItemVariant,
   ButtonVariant,
+  ToolbarToggleGroup,
 } from '@patternfly/react-core';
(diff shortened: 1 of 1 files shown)


brumik (changes requested):
It is looking good, however there are some strange behaviours which hopefully could be fixed:

Screenshot from 2021-09-16 12-04-17
You can see here the select box got on two lines (second line is the badge) and the paginatioun is out of the `Card`.

Screenshot from 2021-09-16 12-06-01
I would like to see this as a button (like the sort direction button looks like)

skochay330:
New commit "Filterable Toolbar Edits" addresses changes requested by @brumik.

brumik (approved):

PR #596, opened 2021-09-21
Package version updates

skochay330:
Updated all package versions.

This excludes: 
- Typescript, webpack, and jest packages 
- "d3", "@patternfly/react-core", and "node-fetch" (which require major updates)


brumik:
@skochay330 can you please rebase to devel and regenerate the package-lock file with `npm i` so we are getting the latest packages there.

* I don't see that `@patternfly/react-core` has a major version. It has just `4.157.3` as latest - and since we are depending on it a lot, it would be beneficial for us to have the latest version. 
* `node-fetch` is used in testing: if you can update the major version without breaking the tests, I would take it :) otherwise not so important, afaik it has no vulnerabilities.
* `"url-parse": "^1.5.3"` is not needed - could you please remove it as I asked? It was added in to combat a vulnerability, but the package is just a subpackage of another packages. If the vulnerability is still there we will need to edit the lock file in a followup PR.

P.S.: I am sorry for using major/minor/patch versions. I know officially now they are using the less confusing and more mouthful breaking change version, feature version and bug-fix version.

brumik (approved):

PR #597, opened 2021-09-23
Fixed date picker not showing in automation calculator

skochay330:
The date selectors were not displaying when "Custom" was selected in Automation Calculator.

Before: 
image

After:
Screen Shot 2021-09-23 at 2 16 31 PM

### DIFF
--- a/src/Components/Toolbar/Groups/QuickDateGroup.tsx
+++ b/src/Components/Toolbar/Groups/QuickDateGroup.tsx
@@ -53,7 +53,7 @@ const QuickDateGroup: FunctionComponent<Props> = ({
         selectOptions={values.quick_date_range}
         setValue={(value) => setFilters('quick_date_range', value)}
       />
-      {filters.quick_date_range === 'custom' && (
+      {filters.quick_date_range.toString().includes('custom') && (
         <Split hasGutter>
           <SplitItem>
             <ToolbarInput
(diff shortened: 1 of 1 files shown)


src/Components/Toolbar/Groups/QuickDateGroup.tsx
```diff
@@ -53,7 +53,7 @@ const QuickDateGroup: FunctionComponent<Props> = ({
         selectOptions={values.quick_date_range}
         setValue={(value) => setFilters('quick_date_range', value)}
       />
-      {filters.quick_date_range === 'custom' && (
+      {filters.quick_date_range.toString().includes('custom') && (
```
brumik, line 56: Currently the `roi` endpoint returns a `roi_custom` string instead of `custom` as in the other endpoints. I believe we talked about this changing in the future since the two endpoints are not compatible with each other because of this issue (`roi_` prefix on quick date ranges). @jlmitch5 is the thing about changing it to relative strings still on what Ladas was pushing forward? I don't see any Jira issues anymore regarding this.

@skochay330 I think it would be much more future-proof to have it like this:
`['custom', 'roi_custom'].includes(filters.quick_date_range as string)` as it would explicitly list the custom names. Then when having a quick date range with custom in it, it would not break anything. Furthermore it would be more obvious what options trigger the custom date range in the toolbar. (In addition then you can tell the TS that it is a string, and you don't need to call an extra function to do nothing :)

skochay330:
Closing this PR in favor of https://github.com/RedHatInsights/tower-analytics-frontend/pull/598

PR #603, opened 2021-10-01
Pdf download button

skochay330:
Created PDF download button in the toolbar of Reports.


src/Api/api.ts
```diff
@@ -106,3 +108,6 @@ export const readClustersOptions = (params: Params): Promise<ApiJson> =>
 
 export const readNotifications = (params: Params): Promise<ApiJson> =>
   get(notificationsEndpoint, params);
+
+export const generatePdf = (params: Params): Promise<void> =>
```
jlmitch5, line 112: Could this be ApiJson, or another more specific type...maybe it needs to be this because it is the intermediary pdf file stream of data?  @brumik @h-kataria

src/Containers/Reports/Details/Report.tsx
```diff
@@ -156,6 +159,16 @@ const Report: FunctionComponent<ReportGeneratorParams> = ({
               isCompact
             />
           }
+          additionalControls={[
+            <DownloadPdfButton
```
jlmitch5, line 163: I believe we want to wrap this in a feature flag

jlmitch5 (approved):
I think this is great, awesome job!  @brumik can you show @skochay330 how to set up the feature flag she added in the code here next week?

PR #607, opened 2021-10-04
Fixed tooltip activation space

skochay330:
Fixed the small activation space on the download PDF button.

Before (only activated tooltip in the middle of the icon):
before activation space

After (now activates in the entire button area):
activation space

### DIFF
--- a/src/Components/Toolbar/DownloadPdfButton.js
+++ b/src/Components/Toolbar/DownloadPdfButton.js
@@ -30,23 +30,21 @@ const DownloadPdfButton = ({ slug, data, y, label, xTickFormat }) => {
   const getPdfButtonText = (error) => error || 'Download PDF version of report';
 
   return (
-    <>
-      <Tooltip
-        position={TooltipPosition.top}
-        content={<div>{getPdfButtonText(error)}</div>}
+    <Tooltip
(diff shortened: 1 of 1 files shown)


src/Components/Toolbar/DownloadPdfButton.js
```diff
@@ -30,21 +30,23 @@ const DownloadPdfButton = ({ slug, data, y, label, xTickFormat }) => {
   const getPdfButtonText = (error) => error || 'Download PDF version of report';
 
   return (
-    <Button
-      variant={error ? ButtonVariant.link : ButtonVariant.plain}
-      aria-label={getPdfButtonText(error)}
-      onClick={() => request(data)}
-      isDanger={error}
-    >
+    <>
```
brumik, line 33: You don't need to add wrapper element. You use it only when you would have more than one element in a root of a component. React does not likes that :)

brumik (changes requested):
One minor change, otherwise looks good :)

src/Components/Toolbar/DownloadPdfButton.js
skochay330: Good catch! Sorry, forgot to remove that.

brumik (approved):

PR #617, opened 2021-10-06
Dev update: option to use stage proxy instead of standalone

brumik:
* For usage look at the updated readme.


brumik:
@jlmitch5 The proxy writes out the URL you should use. It can change from env to env.

brumik:
> Also, we should allow proxying against a specific URL (such as an ephemeral deployment). This will be needed to make the cypress stuff work. So passing what url to proxy as a param to npm start, as well as documentation on how to pass it should be in this PR.

Yes, it is WIP, this is the prerequisite PR for it, to see working against something more "common". The eph will work just like this, when I am able to make it run.

PR #641, opened 2021-10-26
[AA-827] Converted ROI to the new design except the chart

brumik:
All changes hidden behind a feature flag `newAutomationCalculator`. Added the feature flag to stage and prod.

The Component is ready for: 
1. Buttons for changing money/time if API allows
2. Changing to charting lib when gets support

**Before**
Screenshot from 2021-10-26 14-35-34

**After:**
Screenshot from 2021-10-26 14-11-45
Screenshot from 2021-10-26 14-11-36
Screenshot from 2021-10-26 14-11-26

**Changing the chart is not working yet, the chart would look like this which is not good enough:**
Screenshot from 2021-10-26 14-10-09


jlmitch5 (approved):
Looks great!

PR #646, opened 2021-10-29
Changed styled component padding to chart padding

brumik:
This is cleaner. This is the 3rd time to trying to get it in. I just leave this here for future selves to know that it is possible to not to override the toolbar.


jlmitch5 (commented):
I don't think we need this, can we close it?

brumik:
Yup, it was just a showcase for a cleaner padding solution.

PR #656, opened 2021-11-09
[AA-847] ROI to a report

brumik:
* Waiting for description
* Tafique approved the UX

**Current**
Screenshot from 2021-11-11 15-17-25
Screenshot from 2021-11-11 15-17-18
Screenshot from 2021-11-11 15-17-11
Screenshot from 2021-11-11 15-16-54


src/Api/types.ts
```diff
@@ -1,6 +1,6 @@
 export type Params = Record<
   string,
-  string | number | string[] | number[] | boolean
+  string | number | string[] | number[] | boolean | undefined
```
h-kataria, line 3: why do we need undefined here

brumik: end_date can be undefined, we need the value in the queryParams but we are not passing to the API when it is a custom date, this is where you can use `undefined`. It is better than using `null` as `null` is non typed (can be any object and JS has only objects)

h-kataria (approved):

h-kataria:
@brumik this PR is good to be merged, this addresses the issue found in https://issues.redhat.com/browse/AA-881
We can update the description in the follow-up PR as this is hidden report is hidden behind the feature flag

PR #657, opened 2021-11-09
PDF download modal UI

skochay330:
Created a modal to display when PDF download button is clicked, which allows the user to choose if they would like to download a report which contains only the current page, all items (if less than 100), or top 100 (if more than 100). 'Current page' is selected by default, and only one option can be selected at a time.

Jira Issue: https://issues.redhat.com/browse/AA-867

Screen Shot 2021-11-09 at 11 11 54 AM


src/Components/Toolbar/DownloadPdfButton.js
```diff
@@ -54,14 +65,60 @@ const DownloadPdfButton = ({
...
+        variant={ModalVariant.small}
+        title="Export report"
+        isOpen={isExportModalOpen}
+        onClose={() => {
+          setIsExportModalOpen(false);
+        }}
+        actions={[
+          <Button
+            key="export"
+            variant="primary"
```
brumik, line 86: For all buttons please prefer to use the variant constant: `ButtonVariant.xy`

src/Components/Toolbar/DownloadPdfButton.js
```diff
@@ -54,14 +65,60 @@ const DownloadPdfButton = ({
...
+            onClick={() => request()}
+            isDisabled={!isCurrent && !isTotalCount}
+          >
+            Export
+          </Button>,
+          <Button
+            key="cancel"
+            variant="link"
```
brumik, line 94: Same as ^

brumik (changes requested):
This looks good, but I think it could be better:
1. (This comment depends on the backend) There will be more options in the future, such as export as csv. Therefore I would recomment to add instead of `optionSelected` an `options` field which would be an array where we can pass both `csv/pdf` and `limit:10/limit:100` value. I would also recommend using a limit passed to the backend with `:` as a delimiter.
2. Please move the modal window into its own component. Then the button would not get too big, and logically I think it makes sense to have the logic separate. The modal could return only one field `options` which I described above (even if it is not in the API like this)
3. Since the user cannot export it without selecting an option, why would we not mark the `limit:10` option as default. This would save a click in most of the use cases for the users, which is always welcome :)

jlmitch5 (approved):
I disagree with @brumik's proposed changes.  In terms of options, this is a good level of abstraction for the time being, and I don't think things are at the level where we get much value out of moving this to a separate file.  We can make changes to handle these things in the future, once we have more defined requirements around further options.

src/Components/Toolbar/DownloadPdfButton.js
```diff
@@ -21,8 +26,13 @@ const DownloadPdfButton = ({
   y,
   label,
   xTickFormat,
+  totalCount,
+  optionSelected,
```
brumik, line 30: This is not passed to this component. Did you wanted to do something with it?

src/Components/Toolbar/DownloadPdfButton.js
```diff
@@ -33,6 +43,7 @@ const DownloadPdfButton = ({
         y,
         label,
         x_tick_format: xTickFormat,
+        optionSelected,
```
brumik, line 46: I think this line would be something else, depending what the api waits along the lines:
```js
x_tick_format: xTickFormat,
optionSelected: isTotalCount ? 'totalCount' : 'current' // I don't know what the api wants
```

src/Components/Toolbar/DownloadPdfButton.js
```diff
@@ -21,8 +26,13 @@ const DownloadPdfButton = ({
   y,
   label,
   xTickFormat,
+  totalCount,
+  optionSelected,
 }) => {
   const [isModalOpen, setIsModalOpen] = useState(false);
+  const [isExportModalOpen, setIsExportModalOpen] = useState(false);
+  const [isCurrent, setIsCurrent] = useState(false);
+  const [isTotalCount, setIsTotalCount] = useState(false);
```
brumik, line 35: One state is enough as I mentioned in the review above

src/Components/Toolbar/DownloadPdfButton.js
```diff
@@ -54,14 +65,60 @@ const DownloadPdfButton = ({
...
+            onClick={() => setIsExportModalOpen(false)}
+          >
+            Cancel
+          </Button>,
+        ]}
+      >
+        <Grid md={4}>
+          <GridItem>
+            <Radio
+              onChange={() => setIsCurrent(true)}
```
brumik, line 104: Here you need to add `isChecked` as mentioned above.

(Example from PF docs https://codesandbox.io/s/3p7uc)

brumik (changes requested):
If @jlmitch5 approves then OK for the code changes. However what John did not see (and me neither on the first run) that there is a bug:

You are using the PF component mixed (half managed half non managed state), so the radio in the ui shows the switching correctly, but both of the states `isCurrent` and `isTotal` will be true if you click on both of them and then export. 

You can fix this by managing the state completely by adding the `isChecked` property to the radio button. Then I would unify the two states into one (so `isTotalCount` => true then exporting 100 if false then just the current). this way we have less state to manage. 

I would like to see this PR tested with the API working before approving and merging. Is there a backend work on this?

h-kataria:
PDF generator MR: 
PDF Generator MR: https://gitlab.cee.redhat.com/automation-analytics/pdf-generator/-/merge_requests/26

Backend MR: https://gitlab.cee.redhat.com/automation-analytics/automation-analytics-backend/-/merge_requests/287

src/Components/Toolbar/DownloadPdfButton.js
```diff
@@ -33,6 +41,7 @@ const DownloadPdfButton = ({
         y,
         label,
         x_tick_format: xTickFormat,
+        showExtraRows: isCurrent ? false : true,
```
brumik, line 44: Is current is a boolean, you don't need to use a ternary operator (and when you want to convert something else to boolean you can use `!!someStringVar` )

src/Components/Toolbar/DownloadPdfButton.js
```diff
@@ -54,14 +63,61 @@ const DownloadPdfButton = ({
...
+        title="Export report"
+        isOpen={isExportModalOpen}
+        onClose={() => {
+          setIsExportModalOpen(false);
+        }}
+        actions={[
+          <Button
+            key="export"
+            variant={ButtonVariant.primary}
+            onClick={() => request() && setIsExportModalOpen(false)}
```
brumik, line 85: I don't think the `request()` is returning anything, but if it returns false the model will not close, and as we are using another modal for error reporting this could lead to bugs. Consider using a function with these two expressions inside.

brumik (approved):
Those two comments above are small improvements. Otherwise this PR looks good for me :+1:

h-kataria (approved):

PR #658, opened 2021-11-10
[aa-882] Job Explorer Custom Date Range Bugfix

jeanetteobr:
Expose error state when custom date range value is beyond 62 days.

Current:
Screen Shot 2021-11-10 at 10 45 03 AM

Fix:
Screen Shot 2021-11-10 at 10 45 32 AM


h-kataria (approved):

h-kataria:
@jeanetteobr did you try this on other screens, like savings planner etc.

PR #663, opened 2021-11-17
Display tags with descriptions on all reports

skochay330:
Added tags to show on the card view and details view of all reports, with a tooltip containing the description for the tag. 

Jira issue: https://issues.redhat.com/browse/AA-888 

Before: 
Screen Shot 2021-11-17 at 2 22 27 PM
Screen Shot 2021-11-17 at 2 22 38 PM

After:
Screen Shot 2021-11-17 at 2 23 11 PM 
Screen Shot 2021-11-17 at 2 25 43 PM

### DIFF
--- a/src/Containers/Reports/List/ListItem/index.tsx
+++ b/src/Containers/Reports/List/ListItem/index.tsx
@@ -10,10 +10,13 @@ import {
   CardBody,
   CardFooter,
   Label as PFLabel,
+  Tooltip,
+  TooltipPosition,
 } from '@patternfly/react-core';
 
 import paths from '../../paths';
 import { ReportPageParams } from '../../Shared/types';
+import { TAGS } from '../../Shared/constants';
 
 const CardTitle = styled(PFCardTitle)`
   word-break: break-word;
@@ -28,6 +31,7 @@ const Small = styled.small`
 
 const Label = styled(PFLabel)`
   margin-right: 10px;
+  margin-bottom: 10px;
 `;
 
 interface Props {
@@ -35,7 +39,7 @@ interface Props {
 }
 
 const ListItem: FunctionComponent<Props> = ({
-  report: { slug, description, name, categories },
+  report: { slug, description, name, tags },
 }) => (
   <Card data-testid={slug}>
     <CardHeader>
@@ -47,9 +51,20 @@ const ListItem: FunctionComponent<Props> = ({
     </CardHeader>
     <CardBody>{description ? <Small>{description}</Small> : null}</CardBody>
     <CardFooter>
-      {categories.map((category, idx) => (
-        <Label key={idx}>{category}</Label>
-      ))}
+      {tags.map((tagKey, idx) => {
+        const tag = TAGS.find((t) => t.key === tagKey);
+        if (tag) {
+          return (
+            <Tooltip
+              key={`tooltip_${idx}`}
+              position={TooltipPosition.top}
+              content={tag.description}
+            >
+              <Label key={idx}>{tag.name}</Label>
+            </Tooltip>
+          );
+        }
+      })}
     </CardFooter>
   </Card>
 );
--- a/src/Containers/Reports/Shared/constants.ts
+++ b/src/Containers/Reports/Shared/constants.ts
@@ -1,7 +1,121 @@
-export const CATEGORIES = {
-  operations: 'Operations',
-  executive: 'Executive',
-};
+import { Tag } from './types';
+
+export enum TagName {
+  operations = 'Operations',
+  executive = 'Executive',
+  subscription = 'Subscription',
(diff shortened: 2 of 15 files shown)


src/Containers/Reports/Shared/constants.ts
```diff
@@ -1,6 +1,90 @@
...
-  operations: 'Operations',
-  executive: 'Executive',
+export interface OptionsForTags {
+  [key: string]: {
+    name: string;
+    description: string;
+  };
+}
+
+export const TAGS: OptionsForTags = {
```
brumik, line 8: This constants seems to me like it could be stored in the backend in the future, so we could follow the `options` endpoints convention and make it an array instead of an object, along the lines:
```
TAGS = [{
  key: operations,
  name: 'Operations',
  description: 'BlahBlahBlah'
}]
```

cc @jlmitch5

src/Containers/Reports/Shared/constants.ts
```diff
@@ -1,6 +1,90 @@
-export const CATEGORIES = {
-  operations: 'Operations',
-  executive: 'Executive',
+export interface OptionsForTags {
```
brumik, line 1: This should be moved to the `types.ts` file.

src/Containers/Reports/Shared/types.ts
```diff
@@ -27,7 +27,7 @@ export interface ReportPageParams {
   slug: string;
   name: string;
   description: string;
-  categories: string[];
+  tags: Params[];
```
brumik, line 30: Tags are not matching the params type. You probable want to create an interface `Tag`
```
interface Tag {  string; name: string; desc: string }
```
and then on this line you can do `Tag[]`

src/Containers/Reports/Shared/constants.ts
```diff
@@ -1,6 +1,90 @@
...
+  [key: string]: {
+    name: string;
+    description: string;
+  };
+}
+
+export const TAGS: OptionsForTags = {
+  operations: {
```
brumik, line 9: If we decide to do this as an array, from my previous comment this constant could have the type `Tag[]` too.

brumik (changes requested):
I think in overall looking good, I have some comments about TS types. Also this will have a lot of conflicts with #661. 

If it is possible I would wait with this pr for #661 just because it is easier to rebase 2 commits with small changes. cc @jlmitch5

src/Containers/Reports/Details/Details.tsx
```diff
@@ -54,15 +55,23 @@ const Details: FunctionComponent<Record<string, never>> = () => {
...
-                key={`tooltip_${idx}`}
-                position={TooltipPosition.bottom}
-                content={tag.description}
-              >
-                <Label key={idx}>{tag.name}</Label>
-              </Tooltip>
-            ))}
+            {tags.map((tagKey, idx) => {
+              const tag = TAGS.find((t) => t.key === tagKey);
+              if (tag === undefined) {
```
brumik, line 60: In this case I would maybe prefer not showing the tag, but not breaking the page. Throwing an error will show the "Page is broken, please reload screen" which is breaking the user experience. Maybe we can do 

if tag then return tooltip otherwise do nothing.

src/Containers/Reports/List/ListItem/index.tsx
```diff
@@ -50,15 +51,21 @@ const ListItem: FunctionComponent<Props> = ({
...
-          key={`tooltip_${idx}`}
-          position={TooltipPosition.top}
-          content={tag.description}
-        >
-          <Label key={idx}>{tag.name}</Label>
-        </Tooltip>
-      ))}
+      {tags.map((tagKey, idx) => {
+        const tag = TAGS.find((t) => t.key === tagKey);
+        if (tag === undefined) {
```
brumik, line 56: Same ^

src/Containers/Reports/Shared/constants.ts
```diff
@@ -1,91 +1,121 @@
-export interface OptionsForTags {
-  [key: string]: {
-    name: string;
-    description: string;
-  };
+import { Tag } from './types';
+
+export enum TagName {
```
brumik, line 3: This is cool right? ^^ 

I think though that before migrating this into db we should use the string part of the enum as a key, so in the format of `automationHub = 'automationHub` or `automation_hub` (the later is probably looking more readable in url if filtering by it)

Since this is an enum now, we can change it any time easily so I am fine with this as it is now, and we can change it up later when there are some requirements for it.

brumik (changes requested):
This is close to perfect, the only thing that we should not render the tag when you cannot find it instead of breaking the page (see the bellow comment)

brumik (approved):
LGTM

PR #665, opened 2021-11-23
Download notifications

jlmitch5:
Jira issue: https://issues.redhat.com/browse/AA-900 

Screenshot from 2021-12-08 10-32-40
Screenshot from 2021-12-08 10-33-03
Screenshot from 2021-12-08 10-29-29


src/store/pdfDownloadButton/reducer.ts
```diff
@@ -0,0 +1,25 @@
...
+  switch (action.type) {
+    case ReducerTypes.pending:
+      return {
+        ...state,
+        [action.meta.slug]: DownloadState.pending,
+      };
+    case ReducerTypes.fulfilled:
+      return {
+        ...state,
+        [action.meta.slug]: DownloadState.fulfilled,
```
brumik, line 13: @jlmitch5 @skochay330 I know you removed it, but we need it because otherwise the download button stuck in loading state forever

src/store/pdfDownloadButton/types.ts
```diff
@@ -0,0 +1,49 @@
+export enum DownloadState {
+  pending,
+  fulfilled,
+  rejected,
+}
+
+export type State = Record<string, DownloadState>;
+
+export enum ReducerTypes {
+  pending = `PDF_DOWNLOAD_PENDING`,
```
brumik, line 10: I could not find a better name for this reducer (including the folder name), but this is probably a bit better fitting than "Toast"

brumik:
@jlmitch5 This is your PR, so you cannot review it, but you so you can do unofficial review with comments.

src/Api/methods.ts
```diff
@@ -39,24 +49,37 @@ export const authenticatedFetch = (
...
-            .then((error: ApiJson) =>
-              Promise.reject({ status: response.status, error })
-            );
+            .then((error: { detail: { name: string[] } }) => {
+              // Add error reporting notification if we errored out.
+              dispatch(
+                addNotification(notif.rejected(notif.id, error?.detail.name[0]))
+              );
+
+              return Promise.reject();
```
jlmitch5, line 81: Let's add the status/error to the rejection, like on line 58.

src/frontend-components-notifications.d.ts
```diff
@@ -0,0 +1,35 @@
+declare module '@redhat-cloud-services/frontend-components-notifications/redux' {
```
jlmitch5, line 1: let's make a comment that this will be deprecated and we should remove once the lib is typed

src/store/pdfDownloadButton/actions.ts
```diff
@@ -0,0 +1,36 @@
...
+  autoDismiss: false,
+  id,
+});
+
+export const downloadPdf = (
+  params: PDFParams,
+  dispatch: DispatchType,
+  slug: string
+): ActionTypes => ({
+  type: ReducerTypes.act,
```
jlmitch5, line 28: This is reimplementing the Download type you have in the types file.  Use it instead.

jlmitch5 (commented):
Looks good to me, great job on this!

skochay330 (approved):

PR #666, opened 2021-11-24
AA 2.1 Migration Report

jeanetteobr:


cypress/integration/Reports.spec.js
```diff
@@ -221,3 +223,22 @@ describe('Report: Module Usage By Task Smoketests', () => {
...
+    cy.screenshot('report_aa21m_bar.png');
+    cy.get('#line').click();
+    cy.screenshot('report_aa21m_line.png');
+  });
+
+  it('Can change lookback', () => {
+    cy.get('[data-cy="quick_date_range"]').click();
+    cy.get('.pf-c-select__menu-item').contains('Past 62 days').click();
+  });
+});
```
jlmitch5, line 244: awesome!

src/Containers/Reports/Layouts/Standard/ReportCard.tsx
```diff
@@ -55,28 +55,63 @@ const ReportCard: FunctionComponent<ReportGeneratorParams> = ({
...
     fetchOptions();
   }, [queryParams]);
 
+  const updateFilter = () => {
+    if (
+      queryParams.task_action_name &&
+      queryParams.task_action_id.length === 0 &&
+      options?.task_action_id
+    ) {
+      if (Array.isArray(queryParams.task_action_name)) {
```
jlmitch5, line 88: Just realized when seeing it in this format.  Since this if is the only execution path inside here, I think we can actually hoist this as a 4th clause of the parent if, i.e.:

      queryParams.task_action_name &&
      queryParams.task_action_id.length === 0 &&
      options?.task_action_id &&
      Array.isArray(queryParams.task_action_name)

src/Containers/Reports/Layouts/Standard/ReportCard.tsx
```diff
@@ -55,28 +55,63 @@ const ReportCard: FunctionComponent<ReportGeneratorParams> = ({
...
+        //   });
+        // }
+      }
+    }
+  };
+
+  useEffect(() => {
+    updateFilter();
+    setData();
+  }, [options?.task_action_id]);
```
jlmitch5, line 112: Nice, I like how this useEffect is specifically called on just this very particular case, it should cause unnecessary execution of that block to be pretty minimal.

jlmitch5 (commented):
This is fantastic, and is gonna be a really great help to people upgrading to 2.1!  Just one small code tweak, and then I wanted to talk to you about adding a feature flag

src/Containers/Reports/Layouts/Standard/ReportCard.tsx
jeanetteobr: Makes sense. Thanks!

jlmitch5 (approved):
Awesome stuff!

PR #670, opened 2021-11-30
[AA-910] Updated the roi report table displayed values

brumik:
Note the 2 decimal places in the Template productivity column and the `s` in the expanded row elasped data field.
Screenshot from 2021-11-30 10-49-33


src/Containers/Reports/Layouts/AutomationCalculator/TemplatesTable/ExplandedRowContents.tsx
```diff
@@ -23,7 +23,7 @@ const ExpandedRowContents: FunctionComponent<Props> = ({ template }) => (
         <DescriptionListGroup>
           <DescriptionListTerm>Elapsed</DescriptionListTerm>
           <DescriptionListDescription>
-            {template.elapsed}
+            {template.elapsed}s
```
jlmitch5, line 26: can this be in the format "300 seconds", so with a space and the word written out.

jlmitch5 (commented):
looks good, thanks for doing this. one small change.

PR #676, opened 2021-12-09
Removed radio options in PDF download modal

skochay330:
Removed all radio options in pdf download modal when the total number of items is less than or equal to currently displayed items. Instead, text will be displayed reflecting that this report would show all available data.

Jira issue: https://issues.redhat.com/browse/AA-913

Screen Shot 2021-12-07 at 5 37 36 PM


jlmitch5 (approved):
Great job!

PR #678, opened 2021-12-14
Update charting lib

brumik:
This makes no changes how the UI looks


brumik:
Yes. It bumped the version of the charting library to the latest which changes how the data is passed down. Every other change outside the package.json is addressing that change. @h-kataria

h-kataria (approved):

PR #679, opened 2021-12-14
Change the logic how the FF works in devel env

brumik:
And add the ephemeral environment as a devel environment.

See the code comments how the new logic works. This should enable more precise testing, while not disturbing or changing any of the current workflows.


dmzoneill (approved):
/lgtm

PR #681, opened 2021-12-15
[AA-926] Make report tables reorder-able

brumik:
*  Fix the types and mutation in the reports card 
*  Updated the schema to allow reorder-able columns in the table 

No UI/behavioural changes

### DIFF
--- a/src/Containers/Reports/Layouts/Standard/ReportCard.tsx
+++ b/src/Containers/Reports/Layouts/Standard/ReportCard.tsx
@@ -71,26 +71,35 @@ const ReportCard: FunctionComponent<ReportGeneratorParams> = ({
     fetchOptions(queryParams);
   }, [queryParams]);
 
-  const updateFilter = () => {
+  /**
+   * The function is used because the API is not returning the value.
+   * When API gets support, this function should be removed.
+   */
+  const addTaskActionId = (
+    qp: Record<string, string | string[]>
+  ): Record<string, string | string[]> => {
     if (
-      queryParams.task_action_name &&
-      queryParams.task_action_id.length === 0 &&
+      qp.task_action_name &&
+      qp.task_action_id.length === 0 &&
       options?.task_action_id &&
-      Array.isArray(queryParams.task_action_name)
+      Array.isArray(qp.task_action_name)
     ) {
-      const task_action_name = queryParams.task_action_name as Array<T>;
+      const task_action_name = qp.task_action_name;
       const modules = options.task_action_id.filter((obj) =>
         task_action_name.includes(obj.value)
       );
-      queryParams.task_action_id = modules.map((module) =>
-        module.key?.toString()
-      );
+      qp.task_action_id = modules.map((module) => module.key?.toString());
     }
+
+    return qp;
   };
 
+  /**
+   * When API gets support, this effect should be removed.
+   */
   useEffect(() => {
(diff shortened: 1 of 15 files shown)


src/Containers/Reports/Layouts/Standard/ReportCard.tsx
```diff
@@ -56,48 +55,47 @@ const ReportCard: FunctionComponent<ReportGeneratorParams> = ({
...
-  };
-
+  /**
+   * The function is used because the API is not returning the value.
+   * When API gets support, this function should be removed.
+   */
   useEffect(() => {
-    updateFilter();
-    setData(queryParams);
+    const updateFilter = (
```
jlmitch5, line 72: let's keep the function definition out of  the useEffect, but otherwise, looks great

src/Containers/Reports/Shared/schemas/changesMade.ts
```diff
@@ -25,24 +25,27 @@ const tags = [
...
-
-const tableAttributes = [
-  'host_count',
-  'changed_host_count',
-  'host_task_count',
-  'host_task_changed_count',
+  { key: 'host_count', value: 'Host count' },
+  { key: 'changed_host_count', value: 'Changed host count' },
+  { key: 'host_task_changed_count', value: 'Changed task count' },
+  { key: 'host_task_count', value: 'Task count' },
```
jlmitch5, line 34: Task should be above changed task count here (the 3rd col)

src/Containers/Reports/Shared/schemas/templatesExplorer.ts
```diff
@@ -66,7 +67,7 @@ const expandedAttributes = [
 const defaultParams: Params = {
   limit: 6,
   offset: 0,
-  attributes: [...tableAttributes, ...expandedAttributes],
+  attributes: ['failed_count', 'successful_count', 'total_count', ...expandedAttributes],
```
jlmitch5, line 70: these should be reordered to match the headers arr

jlmitch5 (commented):
Looks great, some small changes

src/Containers/Reports/Shared/schemas/templatesExplorer.ts
brumik: Does it matter somewhere? I think has no particular effect on anything. (but I can do it if you wish)

src/Containers/Reports/Layouts/Standard/ReportCard.tsx
brumik: Okey (it has sense to have it in one block)

jlmitch5: I think it helps to see what the useEffect is instantiating when it is just the function calls, not the definition.  It gets longer and the like code-flow becomes a bit harder to parse when it is defined within the effect.

src/Containers/Reports/Shared/schemas/templatesExplorer.ts
jlmitch5: Just figured it would be good to be consistent with the move above, in case we want to add new items in the future, etc.

h-kataria:
other than addressing comments from @jlmitch5, this looks good and working in UI.

jlmitch5 (approved):

PR #683, opened 2021-12-16
[AA-927] Extended row mapper

brumik:
Waiting for #681 

This PR makes the expanded table row swappable in the future, while keeping the schema DB friendly.


jlmitch5 (approved):
This is a nice update to the schemas!

PR #686, opened 2021-12-16
Add job/stage to travis pipeline, dummy workflow for PR

dmzoneill:


jlmitch5 (approved):
I looked at the updates you made on your fork of the bootstrap file, they all make sense.

I'm okay if we pull from a fork of this for now.  If it's possible, I'd like to contribute back to the RedHadInsights repo with conditionals of what we need for the PR check, so that we don't have issues where they change bootstrap and it breaks us because we're no longer up to date with something we need.  Could we file a Jira ticket about that?

jameswnl:
> I looked at the updates you made on your fork of the bootstrap file, they all make sense.
> 
> I'm okay if we pull from a fork of this for now. If it's possible, I'd like to contribute back to the RedHadInsights repo with conditionals of what we need for the PR check, so that we don't have issues where they change bootstrap and it breaks us because we're no longer up to date with something we need. Could we file a Jira ticket about that?

John, that's the plan. ATM, this is just an experiment to see how far we can go and what might be missing.

jlmitch5:
Ah great, carry on @jameswnl :)

PR #692, opened 2021-12-17
Transcende legend

brumik:
* Make all charts toggles (hide show in legend) to be present in the URL in the `settings` namespace by default, allowing to override it with a prop.
* To make it work I had to:
  * make the useQueryParam hook to work with mutliple components for the same namespace. 
  * that needed and update to the QueryParam context, saving the initial values (and removing them) in the global state so all the different hooks can add and remove them
  * While doing it I applied to the changed files a minimal TS update.
  
@jlmitch5 @h-kataria 
I consider this PR done, I won't have time to do the PDF generator part of the transcending legend before Christmas, but it will be relatively fast and easy after this, to do in January.

### DIFF
--- a/src/QueryParams/useQueryParams.js
+++ b/src/QueryParams/useQueryParams.js
@@ -145,7 +145,7 @@ const useQueryParams = (initial, namespace = DEFAULT_NAMESPACE) => {
 
   const { push: dispatch } = useAsyncActionQueue({
     executeAction,
-    waitFor: params,
+    waitFor: [params],
   });
 
   return {
--- a/src/QueryParams/useQueryParams.js
+++ b/src/QueryParams/useQueryParams.js
@@ -1,4 +1,4 @@
-import { useContext } from 'react';
+import { useContext, useEffect } from 'react';
 import moment from 'moment';
 import { QueryParamsContext } from './Context';
 import useAsyncActionQueue from '../Utilities/useAsyncActionQueue';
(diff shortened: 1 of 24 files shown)


src/QueryParams/useQueryParams.js
```diff
@@ -31,6 +29,14 @@ const paramsReducer = (state, { type, value }) => {
       return { ...state, ...value };
 
     /* v1 api reducers */
+    /* Settings reducers START */
+    /* TODO: Rip these two types of reducers apart from the others */
+    /* It is doable for example by creating a different useParams hook */
+    case 'SET_CHART_TYPE':
```
brumik, line 35: @h-kataria Also this could be maybe put into a Jira task. This format is not upkeepable if we continue to add stuff into the URL. It should be a relatively easy task and small refactor to do. If we decide to put it into Jira I can describe maybe in more detail what are the possibilities from my perspective.

h-kataria:
@brumik @jlmitch5 noticed one thing with this PR, not sure if this is the expected behavior. When I hide/disable one of the legends and go to the next page, it automatically hides the element at the same position as the one I had disabled on the first page. Is this expected behavior?

h-kataria:
@brumik, @jlmitch5, and my thoughts are that the "Others" item in the legend should persist the toggle state across pages/sorts/filters/etc. I think for other legend items it should reset legend visibility when changing paging/sorting/filtering etc. What are your thoughts on this?

jlmitch5:
> @brumik, @jlmitch5, and my thoughts are that the "Others" item in the legend should persist the toggle state across pages/sorts/filters/etc. I think for other legend items it should reset legend visibility when changing paging/sorting/filtering etc. What are your thoughts on this?

I think that sounds like a good way to do it

h-kataria (approved):

PR #694, opened 2021-12-20
Make use request fully typechecked

brumik:
Added the request params type to the use request. Now it can derive types from the functions like `readData` api functions and check them.


src/Containers/Reports/Layouts/Standard/ReportCard.tsx
```diff
@@ -58,8 +58,10 @@ const ReportCard: FunctionComponent<ReportGeneratorParams> = ({
     'settings'
   );
 
-  const { result: options, request: fetchOptions } =
-    useRequest<OptionsReturnType>(readOptions, { sort_options: [] });
+  const { result: options, request: fetchOptions } = useRequest<
+    any,
+    OptionsReturnType
+  >(readOptions, { sort_options: [] });
```
brumik, line 64: These `<any, SomeType>` can go away as soon as the function passed to the hook is correctly typed (in this case the `readOptions` function). The hook is capable to get the type from this function both for the arguments and the return type.

brumik:
@h-kataria Do we want to merge or close this (and the other's too). This is probably the smallest.

src/Containers/Notifications/Notifications.tsx
```diff
@@ -126,7 +126,7 @@ const Notifications: FC<Record<string, never>> = () => {
     isLoading,
     isSuccess,
     request: fetchNotifications,
-  } = useRequest<NotificationDataType>(
+  } = useRequest<any, NotificationDataType>(
```
h-kataria, line 129: @brumik linter does not like `any` here and at other places in this PR.

h-kataria: @jeanetteobr try replacing `any` with `unknown` changes look good to me, but I want to test this out in UI and PR needs rebasing.

brumik: It is possible to update the `fetchNotifications` function with types when passing to the `request`, then TS would not need the <any, NotificationDataType> here. If you still have `any` as warning it may be a better choice above unknown since we do know the type, we just did not implemented yet, so the warning is kinda helpful telling it.

h-kataria (approved):

h-kataria:
need to address linter errors

brumik:
@h-kataria The linter errors are coming from different files than this PR touches: `/src/Components/Toolbar/DownloadButton/Steps/EmailDetails/index.tsx`. This PR only adds to the `no explicity any` warnings but causes no errors.

brumik:
Do you still need this PR (it's open for 7 months)? If not I can close it.

brumik:
Conflicts + stale = closing, if somebody wants this can repopen and rebase it.

PR #706, opened 2021-12-23
removing duplicated lint and test in deploy stage

jameswnl:
Currently, the travis ci already have lint and test stages. The third stage `Deploy` is running again the lint and test. That is duplicating the work and slowing down the PR ci turn around time.

https://github.com/RedHatInsights/tower-analytics-frontend/blob/main/.travis.yml#L18-L23

```  
include:
  - stage: Lint
    script: npm run lint
  - stage: Test
    script: npm run test
  - stage: Deploy
    script: npm run deploy && curl -sSL https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master/src/bootstrap.sh | bash -s
```


brumik (approved):
It's good to go,

However you can also replace npm-run-all with npm run as there is no more concurrent tasks needed to run.

PR #714, opened 2022-01-05
Added an option to define the default selected filter

brumik:
And make the reports to use it when some of them has a default
module instead of status selected

Screenshot from 2022-01-05 11-13-36


jlmitch5 (approved):
code looks good

h-kataria (approved):

PR #715, opened 2022-01-05
Added UI to be able to send an email request to backend server

h-kataria:
https://issues.redhat.com/browse/AA-825

API MR: https://gitlab.cee.redhat.com/automation-analytics/automation-analytics-backend/-/merge_requests/307

Modal with Email radio button and email info fields
Screen Shot 2022-01-05 at 11 30 13 AM

Email
Screen Shot 2022-01-05 at 10 54 05 AM

Notification email was sent successfully
Screen Shot 2022-01-05 at 10 53 45 AM

Notification while email is being processed
Screen Shot 2022-01-05 at 10 53 01 AM

Notification with Error sending an email
Screen Shot 2022-01-05 at 10 51 12 AM

### DIFF
--- a/src/Api/methods.ts
+++ b/src/Api/methods.ts
@@ -95,7 +95,28 @@ export const postWithFileReturn = async (
       if (response.body) return saveStream(response.body, fileStream);
     });
 };
+export const postWithEmail = async (
+  endpoint: string,
+  params: Params,
+  { dispatch, ...notif }: NotificationParams
+): Promise<void> => {
+  const url = new URL(endpoint, window.location.origin);
 
+  // Dispatch notification when starts the download.
+  dispatch(addNotification(notif.pending(notif.id, 'Processing Email')));
+  console.log('Email params: ', params);
+  return authenticatedFetch(url.toString(), {
+    method: 'POST',
+    body: JSON.stringify(params),
+  }).then((response) => {
+    // Delete pending notification when we have results.
+    dispatch(removeNotification(notif.id));
+
+    if (response.ok) dispatch(addNotification(notif.success(notif.id)));
+    else dispatch(addNotification(notif.failure(notif.id)));
+    return;
+  });
+};
 export const get = (
   endpoint: string,
   params: Params = {}
--- a/src/Api/methods.ts
+++ b/src/Api/methods.ts
@@ -111,8 +111,16 @@ export const postWithEmail = async (
     // Delete pending notification when we have results.
     dispatch(removeNotification(notif.id));
 
-    if (response.ok) dispatch(addNotification(notif.success(notif.id)));
-    else dispatch(addNotification(notif.failure(notif.id)));
+    if (response.ok)
+      dispatch(
+        addNotification(notif.success(notif.id, 'Email sent successfully'))
+      );
+    else
+      dispatch(
+        addNotification(
+          notif.rejected(notif.id, 'Error sending email, please try again.')
+        )
+      );
     return;
   });
 };
--- a/src/Api/methods.ts
+++ b/src/Api/methods.ts
@@ -2,10 +2,11 @@ import { stringify } from 'query-string';
 import { saveStream } from './streamSaver';
 import {
   ApiJson,
+  NotificationParams,
   Params,
   ParamsWithPagination,
   PDFParams,
-  NotificationParams,
+  saveROIParams,
 } from './types';
 import { createWriteStream } from 'streamsaver';
 import {
@@ -139,6 +140,17 @@ export const get = (
 export const post = (
   endpoint: string,
   params: Params = {}
+): Promise<ApiJson> => {
+  const url = new URL(endpoint, window.location.origin);
+  return authenticatedFetch(url.toString(), {
+    method: 'POST',
+    body: JSON.stringify(params),
+  }).then(handleResponse);
+};
+
+export const saveROIData = (
+  endpoint: string,
+  params: saveROIParams
 ): Promise<ApiJson> => {
   const url = new URL(endpoint, window.location.origin);
   return authenticatedFetch(url.toString(), {
--- a/src/Api/methods.ts
+++ b/src/Api/methods.ts
@@ -104,7 +104,6 @@ export const postWithEmail = async (
 
   // Dispatch notification when starts the download.
   dispatch(addNotification(notif.pending(notif.id, 'Processing Email')));
-  console.log('Email params: ', params);
   return authenticatedFetch(url.toString(), {
     method: 'POST',
     body: JSON.stringify(params),
--- a/src/Api/types.ts
+++ b/src/Api/types.ts
@@ -34,6 +34,7 @@ export interface PDFParams {
 
 export type NotificationAsyncFunction = (
   id: string,
+  title?: string,
   message?: string
 ) => NotificationOptions;
 
@@ -41,7 +42,6 @@ export interface NotificationParams {
   pending: NotificationAsyncFunction;
   rejected: NotificationAsyncFunction;
   success: NotificationAsyncFunction;
-  failure: NotificationAsyncFunction;
   dispatch: DispatchType;
   id: string;
 }
--- a/src/Api/types.ts
+++ b/src/Api/types.ts
@@ -32,6 +32,17 @@ export interface PDFParams {
   };
 }
 
+export interface saveROIParams {
+  currency: string;
+  hourly_manual_labor_cost: number;
+  hourly_automation_cost: number;
+  templates_manual_equivalent: {
+    template_id: number;
+    effort_minutes: number;
+    template_weigh_in: boolean;
+  }[];
+}
+
 export type NotificationAsyncFunction = (
   id: string,
   title?: string,
--- a/src/Api/types.ts
+++ b/src/Api/types.ts
@@ -40,6 +40,8 @@ export type NotificationAsyncFunction = (
 export interface NotificationParams {
   pending: NotificationAsyncFunction;
   rejected: NotificationAsyncFunction;
+  success: NotificationAsyncFunction;
+  failure: NotificationAsyncFunction;
   dispatch: DispatchType;
   id: string;
 }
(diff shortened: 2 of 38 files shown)


src/Api/types.ts
```diff
@@ -38,6 +38,8 @@ export type NotificationAsyncFunction = (
 export interface NotificationParams {
   pending: NotificationAsyncFunction;
   rejected: NotificationAsyncFunction;
+  success: NotificationAsyncFunction;
+  failure: NotificationAsyncFunction;
```
brumik, line 42: success === resolved
failure === rejected

src/Components/Toolbar/DownloadPdfButton.tsx
```diff
@@ -108,13 +138,25 @@ const DownloadPdfButton: FC<Props> = ({
...
               variant={ButtonVariant.primary}
+              isDisabled={downloadType === 'email'}
               onClick={() => {
                 setIsExportModalOpen(false);
                 downloadPdf();
               }}
             >
               Export
             </Button>
+            <Button
```
brumik, line 149: To not to mix two types of download logic I would create:
`DownloadPDFButton({ closeModalFnc, pdfParams });`
`EmailPDFButton({ closeModalFnc, pdfParams });`
Where each would contain it's own logic. The modal could stay as their parent component.

src/Components/Toolbar/DownloadPdfButton.tsx
```diff
@@ -154,9 +196,26 @@ const DownloadPdfButton: FC<Props> = ({
...
+                  aria-label="email-radio"
+                />
+              </GridItem>
             </>
           )}
         </Grid>
+        {downloadType === 'email' && (
+          <Grid>
+            <GridItem>
+              <EmailSend emailInfo={emailInfo} onChange={setEmailInfo} />
```
brumik, line 215: This is cool :+1:

src/Components/Toolbar/DownloadPdfButton.tsx
```diff
@@ -154,9 +196,26 @@ const DownloadPdfButton: FC<Props> = ({
                   aria-label="total-count-radio"
                 />
               </GridItem>
+              <GridItem>
+                <Radio
+                  onChange={() => setDownloadType('email')}
+                  isChecked={downloadType === 'email'}
```
brumik, line 202: Disclaimer: I was not at any UX designs for this feature.

I think if we would make a checkbox next to the button that would say: 
* [ ] Send Email
Then we could show always only one button instead of disabling them and the `current` / `all rows` could be still used to change how many data should be in the report.

src/store/pdfDownloadButton/actions.ts
```diff
@@ -20,6 +24,21 @@ const rejected = (id: string, message?: string) => ({
   id,
 });
 
+const success = (id: string) => ({
```
brumik, line 27: same as before, success is resolved

src/store/pdfDownloadButton/actions.ts
```diff
@@ -20,6 +24,21 @@ const rejected = (id: string, message?: string) => ({
...
+  variant: NotificationType.info,
+  title: 'Email sent successfully.',
+  id,
+  autoDismiss: false,
+});
+
+const failure = (id: string, message?: string) => ({
```
brumik, line 34: failure === rejected

src/store/pdfDownloadButton/types.ts
```diff
@@ -2,13 +2,17 @@ export enum DownloadState {
...
+  failure,
 }
 
 export type State = Record<string, DownloadState>;
 
 export enum ReducerTypes {
   pending = `PDF_DOWNLOAD_PENDING`,
   fulfilled = `PDF_DOWNLOAD_FULFILLED`,
+  success = `EMAIL_SUCCESS`,
+  failure = `EMAIL_FAILURE`,
```
brumik, line 15: these two won't be needed I believe

However if we need to keep it let's namespace it with prepending it with the name of the reducer. This way you can make sure that if in different reducers you have an action with the same name (`EMAIL_SUCCESS` for example) then the dispatch won't trigger the actions in both reducers.

brumik (commented):
This is not a review. 

I think this feature will be great and even in WIP is looking pretty good. I looked into the code and made some suggestions. 

Also it needs rebase and linting :)

src/Components/Toolbar/DownloadPdfButton.tsx
h-kataria: Thanks for the feedback, we can discuss this on the UX call next week. We haven't had any UX on this feature yet, this is the first cut of the Email feature and was designed by me to show how it works.

src/Api/methods.ts
```diff
@@ -95,7 +95,27 @@ export const postWithFileReturn = async (
...
+  dispatch(addNotification(notif.pending(notif.id, 'Processing Email')));
+  return authenticatedFetch(url.toString(), {
+    method: 'POST',
+    body: JSON.stringify(params),
+  }).then((response) => {
+    // Delete pending notification when we have results.
+    dispatch(removeNotification(notif.id));
+
+    if (response.ok) dispatch(addNotification(notif.success(notif.id)));
+    else dispatch(addNotification(notif.failure(notif.id)));
```
jlmitch5, line 115: lets use notif.rejected instead of creating the new notif.failure?

src/store/pdfDownloadButton/actions.ts
```diff
@@ -20,6 +23,21 @@ const rejected = (id: string, message?: string) => ({
   id,
 });
 
+const success = (id: string) => ({
+  variant: NotificationType.success,
+  title: 'Email sent successfully.',
+  id,
+  autoDismiss: false,
```
jlmitch5, line 30: auto-dismiss should be true, or rather, i think we can remove this custom prop and just go w default behavior for this.

src/Components/Toolbar/EmailSend.tsx
```diff
@@ -0,0 +1,132 @@
...
+          </FormHelperText>
+        )}
+      </FormGroup>
+      {emailInfo.users.length > 0 && (
+        <FormGroup label="User emails" fieldId="emails-field">
+          {emailInfo.users.map(({ name, emails }, i) => {
+            return (
+              <p key={i}>
+                {/* eslint-disable-next-line @typescript-eslint/no-unsafe-call */}
+                <b>{name}</b>: {emails.join(', ')}
```
jlmitch5, line 92: looks like the linter wants us to type emails...I believe explicitly calling it an array of strings would fix it.  or typing emailInfo.users as an array of objects with name and emails props

src/Components/Toolbar/EmailSend.tsx
```diff
@@ -0,0 +1,132 @@
...
+    }
+    onChange({ ...emailInfo });
+  };
+
+  const [showError, setShowError] = useState(false);
+  const [isExpanded, setIsExpanded] = useState(false);
+  return (
+    <Form>
+      <FormGroup label="Recipient" isRequired fieldId="recipient-field">
+        <Select
```
jlmitch5, line 57: looks like all aria stuff is done correctly here, nice!  would you mind testing to make sure the form can be navigated completely with keyboard (tab/etc.)

src/Components/Toolbar/EmailSend.tsx
```diff
@@ -0,0 +1,132 @@
...
+  emailInfo,
+  onChange = () => null,
+  rbacGroups,
+}) => {
+  const { body, recipient, reportUrl, subject } = emailInfo;
+  const onInputChange = (field: string, val: string) => {
+    onChange({ ...emailInfo, [field]: val });
+  };
+
+  const onSelectionChange = (field: string, val: string) => {
```
jlmitch5, line 36: very nice, this form is pretty easy to understand, parse and cleanly written

src/Components/Toolbar/EmailSend.tsx
```diff
@@ -0,0 +1,132 @@
...
+} from '@patternfly/react-core';
+import { ExclamationCircleIcon } from '@patternfly/react-icons';
+
+interface Props {
+  emailInfo: {
+    recipient: string[];
+    subject: string;
+    body: string;
+    reportUrl: string;
+    users: Record<string, string | string[] | any>[];
```
jlmitch5, line 20: type here seems a bit strange...why not just string[]?

src/Components/Toolbar/DownloadPdfButton.tsx
```diff
@@ -164,9 +310,30 @@ const DownloadPdfButton: FC<Props> = ({
...
+                  aria-label="email-radio"
+                />
+              </GridItem>
             </>
           )}
         </Grid>
+        {downloadType === 'email' && (
+          <Grid style={{ paddingTop: '15px' }}>
+            <GridItem>
+              <EmailSend
```
jlmitch5, line 329: Let's call this EmailDetailsForm

src/Components/Toolbar/DownloadPdfButton.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FC, useState } from 'react';
+import React, { FC, useEffect, useState } from 'react';
```
jlmitch5, line 1: This is still readable, but I do feel like it is getting a bit long.  At this point, it probably would benefit from an abstraction of pdf download button and the modal that opens into two separate components.

jlmitch5 (changes requested):
Very good! just a few details here and there

src/Components/Toolbar/types.ts
```diff
@@ -22,3 +22,32 @@ export type ApiOptionsType = Record<
...
+  platform_default: boolean;
+  roleCount: number;
+  created: string;
+  modified: string;
+  system: boolean;
+}
+
+export interface RbacPrincipalFromApi {
+  username: string;
+  //email: string;
```
jlmitch5, line 52: remove comment?

jlmitch5 (approved):
Very nice, code looks good. remove that comment, see which casts can be successfully removed

src/Components/Toolbar/types.ts
h-kataria: this line needs to stay, we need to uncomment and adjust code to read email of RBAC group principal instead of username in real environment

PR #718, opened 2022-01-06
[WIP] QP

brumik:
// From the Components/Chart/index.txt

    // This dispatch triggers a queryParams change
    // the queryParams change triggers a data change
    // the data change triggers the use effect, meaning
    // the hidden is reset on every second update after
    // the URL HAS THE API QUERY PARAMS.

    // Solution:
    // Query param value change should not trigger change in other values!!!!
    // Acchievable: Redux
    // Not acchievable: Like this (QueryParamsProvider)


src/QueryParams/types.ts
```diff
@@ -4,12 +4,14 @@ export type QueryParams = Record<
...
+type UpdateInsideFunction = (curr: QueryParams) => QueryParams;
+
 export type UpdateFunction = ({
-  newQueryParams,
   namespace,
+  fnc,
 }: {
-  newQueryParams: QueryParams;
   namespace: string;
+  fnc: UpdateInsideFunction;
```
brumik, line 14: This change is not really needed. But even with this it is not working for different reasons soooo

brumik:
@h-kataria I think this can be closed as you used other methods to fix this problem.

PR #720, opened 2022-01-07
ROI report table fixes

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AA-935

Before:

After:
Screenshot 2022-01-07 at 14 56 05
Screenshot 2022-01-07 at 14 56 25
Screenshot 2022-01-07 at 14 56 37

### DIFF
--- a/src/Containers/Reports/Layouts/AutomationCalculator/TemplatesTable/Row.tsx
+++ b/src/Containers/Reports/Layouts/AutomationCalculator/TemplatesTable/Row.tsx
@@ -24,6 +24,27 @@ interface Props {
   redirectToJobExplorer: (id: number) => void;
 }
 
+const setLabeledValue = (key: string, value: number) => {
+  let label;
+  switch(key) {
+    case 'elapsed':
+      label = 'seconds';
+      break;
+    case 'template_automation_percentage':
+      label = '%';
+      break;
+    case 'successful_hosts_savings':
+    case 'failed_hosts_costs':
+    case 'monetary_gain':
+      label = '$';
+      break;
+    default:
+      label = '';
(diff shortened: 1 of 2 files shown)


src/Containers/Reports/Layouts/AutomationCalculator/TemplatesTable/Row.tsx
```diff
@@ -24,6 +24,27 @@ interface Props {
...
+      break;
+    case 'successful_hosts_savings':
+    case 'failed_hosts_costs':
+    case 'monetary_gain':
+      label = '$';
+      break;
+    default:
+      label = '';
+  }
+  let convertedValue = label.includes('count') ? value : label === '$' ? currencyFormatter(value) : +value.toFixed(2);
```
brumik, line 44: I think a double ternary operator is really hard to read. I would say this is not the last time we want to format by seconds and percentages, so maybe we can create a helper function in the utilities (if makes sense localized one - for example in elapsed seconds the floating point is dot in half of the world and comma in the other half) and then here we can just apply them depending on the key

Something along the lines of:
```ts
case 'elapsed': return formatSeconds(value);
case '...something_percentage': return formatPercentage(value);
case 'somethin_money': return formatCurrency(value);
...
default: return value;
```

This change would make this code in my opinion much more readable and using helper functions would make it much more prepared for localization (which is in the plans somewhere down the line) and would make it also more easier to stay consistent trough the app.

ZitaNemeckova:
Closes in favor https://github.com/RedHatInsights/tower-analytics-frontend/pull/724 that's created correctly from branch and not fork.

PR #726, opened 2022-01-17
 ROI report has persistent inputs

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AA-850

- [x] show all/show button in table persists

- [x] manual cost of automation  persists

- [x] automated process cost persists

- [x] manual time input persists

### DIFF
--- a/src/QueryParams/useQueryParams.js
+++ b/src/QueryParams/useQueryParams.js
@@ -86,6 +86,7 @@ const paramsReducer = (state, { type, value }) => {
     case 'SET_ROOT_WORKFLOWS_AND_JOBS':
     case 'SET_INVENTORY':
     case 'SET_SORT_OPTIONS':
+    case 'SET_CALCULATOR':
     case 'SET_SORT_ORDER':
       return { ...state, ...value };
     case 'SET_QUICK_DATE_RANGE':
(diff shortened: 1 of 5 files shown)


src/QueryParams/useQueryParams.js
```diff
@@ -193,6 +218,12 @@ const useQueryParams = (initial, namespace = DEFAULT_NAMESPACE) => {
         dispatch({ type: 'SET_LIMIT', value: limit });
       }
     },
+    setFromTable: (varName, value) => {
```
brumik, line 221: I know this file is ripe for refactor, this is just my 2cents (take it or leave it):

These two functions are basically identical to using `dispatch({ type: 'ACTION', value: 'xxx' });` with an extra possibility of actionMapper not having a key. I think moving towards proper `redux styled actions` would be here more sensible:

```ts
// Extra typechecking and making sure the `type` always fits :)
const setManualCost = (value: number) => ({ type: 'SET_CALCULATOR_MANUAL', value });
...
// then in the calculator
dispatch(setManualCost(value));
```

Or just use dispatch with same complexity and more flexibility:
```ts
// calculator code:
dispatch({ type: 'SET_CALCULATOR_MANUAL' value });
```

h-kataria:
@ZitaNemeckova can you please resolve conflicts on the PR

src/QueryParams/useQueryParams.js
h-kataria: @ZitaNemeckova I suggest going with the latter suggestion to be consistent. But we can  do that in a follow-up PR

h-kataria:
@ZitaNemeckova I found an issue similar to https://issues.redhat.com/browse/AA-952 with Show/Hide toggle in this PR. You can fix that also in a follow-up PR or along with a fix for AA-952
Another issue that I found is that it keeps refreshing the screen as I am typing in input boxes to persist the changes, perhaps we can update the code to only re-render specific parts of the screen while the user is typing, like the chart/Total Savings component and Savings in the table. All this can be done in a separate PR, this PR is looking good as it is, so I am going go to go ahead and merge this one. Feel free to ping me if you need help or have questions.

h-kataria (approved):

PR #740, opened 2022-02-04
Preserve ROI report params in DB

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AA-1007
Fixes https://issues.redhat.com/browse/AA-987 as well

- Values from calculation and table are no longer saved in params but to DB.

- Not using `useRequest` for saving the values to DB allows me not to reload components. But it goes against consistency of the code so it's up to discussion :)

- rbac check - a readonly user cannot edit inputs. See screenshot at the end

- moving hide/show logic from API (based on index) to DB (based on id) solves https://issues.redhat.com/browse/AA-987

TODO: 
- [x] rbac check
- [x] changes saved in DB
- [x] values read from api not default
- [x] removw extra code
- [x] Add alert when save fails
- [x] clean the code
- [x] refresh of the page not working

### RBAC for an user with only read rights
Before:
Screenshot 2022-02-04 at 13 34 53

After:
Screenshot 2022-02-03 at 13 19 28

Error:
Screenshot 2022-02-17 at 12 19 34

### DIFF
--- a/src/Api/types.ts
+++ b/src/Api/types.ts
@@ -32,6 +32,13 @@ export interface PDFParams {
   };
 }
 
+export interface saveROIParams {
+  currency?: string;
+  hourly_manual_labor_cost?: number;
+  hourly_automation_cost?: number;
+  templates_manual_equivalent?: {template_is: number, effort_minutes: number, template_weigh_in: boolean}[];
+}
+
 export type NotificationAsyncFunction = (
   id: string,
   message?: string
--- a/src/Api/types.ts
+++ b/src/Api/types.ts
@@ -33,11 +33,11 @@ export interface PDFParams {
 }
 
 export interface saveROIParams {
-  currency?: string;
-  hourly_manual_labor_cost?: number;
-  hourly_automation_cost?: number;
-  templates_manual_equivalent?: {
-    template_is: number;
+  currency: string;
+  hourly_manual_labor_cost: number;
+  hourly_automation_cost: number;
+  templates_manual_equivalent: {
+    template_id: number;
     effort_minutes: number;
     template_weigh_in: boolean;
   }[];
--- a/src/Api/types.ts
+++ b/src/Api/types.ts
@@ -36,7 +36,11 @@ export interface saveROIParams {
   currency?: string;
   hourly_manual_labor_cost?: number;
   hourly_automation_cost?: number;
-  templates_manual_equivalent?: {template_is: number, effort_minutes: number, template_weigh_in: boolean}[];
+  templates_manual_equivalent?: {
+    template_is: number;
+    effort_minutes: number;
+    template_weigh_in: boolean;
+  }[];
 }
 
 export type NotificationAsyncFunction = (
--- a/src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
+++ b/src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
@@ -173,25 +173,23 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
     };
   };
 
-  const updateCalculationValues = (varName: string, value: number) => {
+  const updateCalculationValues = async (varName: string, value: number) => {
     const hourly_automation_cost =
       varName === 'automation_cost' ? value : costAutomation;
     const hourly_manual_labor_cost =
       varName === 'manual_cost' ? value : costManual;
-    const prom = saveROI(
-      getROISaveData(
-        api.result.items,
-        hourly_manual_labor_cost,
-        hourly_automation_cost
-      )
-    );
-    prom
-      .then(() =>
-        varName === 'manual_cost'
-          ? setCostManual(value)
-          : setCostAutomation(value)
-      )
-      .catch((err) => console.log(err));
+    try {
+      await saveROI(
+        getROISaveData(
+          api.result.items,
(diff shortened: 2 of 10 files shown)


ZitaNemeckova:
@h-kataria can you please review. It's not perfect but I need a new set of eyes on it and a pointer how to add alerts when saving fails.

src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
```diff
@@ -172,6 +155,44 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
...
+      templates_manual_equivalent: updatedDataApi,
+    };
+  };
+
+  const updateCalculationValues = (varName: string, value: number) => {
+    const hourly_automation_cost =
+      varName === 'automation_cost' ? value : costAutomation;
+    const hourly_manual_labor_cost =
+      varName === 'manual_cost' ? value : costManual;
+    const prom = saveROI(
```
jlmitch5, line 181: can we use async/await here?

if you hadn't seen that syntax before, just a few things:
- mark the function as async
- rewrite the promise call as.

```
try {
  await getROISaveData
} catch (err) {
  err handling (which should probably be surfaced someway, not just a console.log)
}
// stuff in .then
```

src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
```diff
@@ -192,12 +213,8 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
         return el;
       }
     });
-
-    setValue(updatedData);
-    setFromTable(
-      'time_per_item',
-      updatedData.map((item) => item.avgRunTime)
-    );
+    const prom = saveROI(getROISaveData(updatedData));
```
jlmitch5, line 216: same here with the async/await

src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
```diff
@@ -206,11 +223,8 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
       : api.result.items.map((el) =>
           el.id === id ? { ...el, enabled: value } : el
         );
-    setValue(updatedData);
-    setFromTable(
-      'enabled_per_item',
-      updatedData.map((item) => item.enabled)
-    );
+    const prom = saveROI(getROISaveData(updatedData));
```
jlmitch5, line 226: same here async/await

src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
```diff
@@ -297,6 +320,10 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
     return tooltip;
   };
 
+  const isReadOnly = (api) => {
+    return !!api.result.rbac?.perms?.read;
```
jlmitch5, line 324: nice and clean!

src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
```diff
@@ -227,6 +241,15 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
       },
     };
   };
+  /**
+   * Set cost from API on load. Don't reload it.
```
jlmitch5, line 245: nice comment here, it wouldn't have been clear to me that this wasn't setting it every time they change without it.

jlmitch5 (changes requested):
looking good, just a couple things

src/Api/types.ts
```diff
@@ -32,6 +32,17 @@ export interface PDFParams {
   };
 }
 
+export interface saveROIParams {
+  currency?: string;
+  hourly_manual_labor_cost?: number;
+  hourly_automation_cost?: number;
+  templates_manual_equivalent?: {
```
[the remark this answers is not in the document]
ZitaNemeckova: @jlmitch5 It's according to what names API expects. Except  for`template_is`. That  is a typo (and fixed).

h-kataria:
@ZitaNemeckova I found lost $10

the issue is caused by the code that calculates `costManual` and `costAutomation` default values in the report vs the old ROI screen, compare these two blocks of code, in the older version, costAutomation value defaults to 20, but in the new version it is being returned as 25 from API, so either that's a bug or need to adjust the tests appropriately.

Old ROI screen code:
https://github.com/RedHatInsights/tower-analytics-frontend/blob/devel/src/Containers/AutomationCalculator/AutomationCalculator.js#L77-L78

New code from changes in this PR for ROI report:
https://github.com/RedHatInsights/tower-analytics-frontend/pull/740/files#diff-3bfe0ae51512c2bb1ca6456a2679e8328a32f5035d3d70593504a1bf63399228R253-R258

h-kataria:
> @h-kataria can you please review. It's not perfect but I need a new set of eyes on it and a pointer how to add alerts when saving fails and failing tests.

Can we add a toast notification when save fails with error info, that will not be auto-dismissed, what do you think @jlmitch5

ZitaNemeckova:
@h-kataria Alert resolved. It's completely ready for review :)

jlmitch5 (approved):
Looks great!

PR #741, opened 2022-02-04
Debugging reports 404 error and attempt to fix

skochay330:
Added more console.logs to investigate temporary 404 error on stage reports and attempted to solve the issue- will need to merge to stage to see if it works (issue cannot be recreated locally).

Jira issue:  https://issues.redhat.com/browse/AA-988

### DIFF
--- a/src/Containers/Reports/Details/index.tsx
+++ b/src/Containers/Reports/Details/index.tsx
@@ -1,10 +1,8 @@
-import React, { FunctionComponent, useEffect } from 'react';
+import React, { FunctionComponent } from 'react';
 import { useParams } from 'react-router-dom';
 import styled from 'styled-components';
 import Main from '@redhat-cloud-services/frontend-components/Main';
 import Error404 from '../../../Components/Error404';
-import LoadingState from '../../../Components/ApiStatus/LoadingState';
-import { ReportSchema } from '../Layouts/types';
 
 import {
   PageHeader,
@@ -22,6 +20,8 @@ import getComponent from '../Layouts';
 import { getReport } from '../Shared/schemas';
 import paths from '../paths';
 import { TAGS } from '../Shared/constants';
+import { ReportSchema } from '../Layouts/types';
+import LoadingState from '../../../Components/ApiStatus/LoadingState';
 
 const Description = styled.p`
   max-width: 70em;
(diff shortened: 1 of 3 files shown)


src/FeatureFlags/useFeatureFlag.ts
```diff
@@ -26,8 +26,12 @@ const isEnabledDevel = (feature?: FeatureFlagType) => {
  * If defined and false => disabled
  */
 const isEnabled = (feature?: FeatureFlagType) => {
+  console.log(feature);
+  console.log(!!feature);
+  console.log(feature?.enabled);
   return !!feature && feature?.enabled;
 };
+console.log(isEnabled);
```
ZitaNemeckova, line 34: This will return a function. Is it what you want?

ZitaNemeckova (commented):
Looks good! I would add more `console.logs` around beta logic given it's breaking on beta. WDYT?

src/FeatureFlags/useFeatureFlag.ts
skochay330: Good catch, thanks!

src/Containers/Reports/Details/index.tsx
```diff
@@ -76,7 +79,10 @@ const Details: FunctionComponent<Record<string, never>> = () => {
       );
   };
 
-  return render();
+  useEffect(() => {
+    render();
+  }, [report]);
+  return report ? render() : <LoadingState />;
```
ZitaNemeckova, line 85: Unfrtunately this won't work for non-existing reports and an user will get `LoadingState` forever instead of 404 page. I think it needs to be fixe in `getReport ` to make sure all correct reports are returned at first try.

h-kataria: @ZitaNemeckova won't the useEffect re-render when `reports` is updated and returns `undefined`

ZitaNemeckova: @h-kataria  `undefined` => `false` so it goes to the loading state. And the problem is that `getReport` returns `undefined` at first and then it returns a report. And you need to consider following use cases:

- When an user  goes to a report that doesn't exist like `/ansible/insights/reports/non-existing` `getReport` function will return `undefined` and the user will see a loading screen instead of 404.  
- When an user is looking at a report that is only available in beta and clicks to `Stop using the beta release`. 

Those two are the cases when `undefined` will not change and 404 is needed.

h-kataria (approved):

PR #745, opened 2022-02-09
added more debugging statements

h-kataria:
@skochay330


skochay330 (approved):

src/FeatureFlags/useFeatureFlag.ts
```diff
@@ -44,7 +44,12 @@ const useFeatureFlag = (flag: ValidFeatureFlags): boolean => {
```
ZitaNemeckova, line 44: @h-kataria I'd add a console.log to line 37 and see what's in the `features`.

ZitaNemeckova (approved):

src/FeatureFlags/useFeatureFlag.ts
h-kataria: @skochay330 @ZitaNemeckova it looks like first time when page is loaded, `features` is an empty array, need to look into that a bit more to see why that's happening

PR #774, opened 2022-03-22
[AA-1075] Report preview UX fixes

skochay330:
Condensed and standardized the height of the preview pane for all reports. 

Jira issue: https://issues.redhat.com/browse/AA-1075

Before: 
Screen Shot 2022-03-22 at 5 18 37 PM

After: 
Screen Shot 2022-03-23 at 3 01 56 PM

### DIFF
--- a/src/Components/ApiStatus/NoData.tsx
+++ b/src/Components/ApiStatus/NoData.tsx
@@ -8,7 +8,7 @@ import {
 import { CubesIcon } from '@patternfly/react-icons';
 
 const NoData: FunctionComponent<Record<string, never>> = () => (
-  <EmptyState variant={EmptyStateVariant.full}>
+  <EmptyState variant={EmptyStateVariant.full} style={{ minHeight: '400px' }}>
     <EmptyStateIcon icon={CubesIcon} />
     <Title headingLevel="h5" size="lg">
       No Data
--- a/src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
+++ b/src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
@@ -437,15 +437,15 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
     </Stack>
   );
 
-  const renderContents = () => (
-    <Card>
-      <CardBody>
-        <FilterableToolbar
-          categories={options}
-          filters={queryParams}
-          setFilters={setFromToolbar}
-          pagination={
-            fullCard && (
+  const renderContents = () =>
+    fullCard ? (
+      <Card>
+        <CardBody>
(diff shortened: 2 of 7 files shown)


src/Components/ApiStatus/NoData.tsx
```diff
@@ -8,7 +8,7 @@ import {
 import { CubesIcon } from '@patternfly/react-icons';
 
 const NoData: FunctionComponent<Record<string, never>> = () => (
-  <EmptyState variant={EmptyStateVariant.full}>
+  <EmptyState variant={EmptyStateVariant.full} style={{ minHeight: '400px' }}>
```
h-kataria, line 11: @skochay330 do we need this change in this PR?

skochay330: I believe we do, because the goal is to make sure the pane is the same height at all times. In a case where there is no data to display, we want to make sure the pane remains the same size rather than shrinking.

h-kataria:
@ZitaNemeckova please review

h-kataria:
LGTM

h-kataria (approved):

src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
```diff
@@ -482,15 +476,12 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
                 startDate={queryParams.start_date}
                 endDate={queryParams.end_date}
                 dateRange={queryParams.quick_date_range}
-                inputs={{ costManual, costAutomation }}
```
ZitaNemeckova, line 485: This line needs to stay otherwise user's inputs won't be part of a generated pdf.

ZitaNemeckova:
@skochay330 This may be out of scope of this PR but it seems to be a bit broken when I change a filter (and attributes are saved to the URL) and then change the previewed report. Sometime it has wrong values in the filter othertimes it crashes. I'd remove all params from the URL. when previewed report changes.

ZitaNemeckova (changes requested):
`inputs={{ costManual, costAutomation }}` shouldn't be removed.

ZitaNemeckova (approved):
LGTM 👍

src/Containers/Reports/Shared/schemas/hostsByOrganizations.ts
```diff
@@ -42,7 +42,7 @@ const defaultParams = {
   quick_date_range: 'last_30_days',
   status: [],
   org_id: [],
-  job_type: ['workflowjob', 'job'],
```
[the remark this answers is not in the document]
skochay330: Yes, we want to remove the pre-set filters from both schemas

PR #778, opened 2022-04-06
 In standard reports use ids for visibility instead of index

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AA-952


h-kataria (approved):

ZitaNemeckova:
@h-kataria I wanted to merge it with pdf change

PR #779, opened 2022-04-06
load standalone env using stage-beta branch

h-kataria:
This fixes errors trying to run/access UI locally when running in standalone mode.

@ZitaNemeckova can you verify if we need this change in our local environments to run UI successfully. I cannot run locally without this fix.


h-kataria:
this changes was merged as part of https://github.com/RedHatInsights/tower-analytics-frontend/pull/785

PR #788, opened 2022-04-25
Set chartType to default if undefined

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AA-1121

**Steps to reproduce:**

- go to any standard report

- click to any item in the legend

- click download button and try to download PDF

**Current behaviour :**

Empty PDF is generated

**Expected behaviour:**

A PDF with chart and table is generated


src/Containers/Reports/Layouts/Standard/ReportCard.tsx
```diff
@@ -168,6 +168,10 @@ const ReportCard: FunctionComponent<StandardProps> = ({
       },
     };
   };
+  // if visibility sets chartType to undefined set it back to default value
```
h-kataria, line 171: it's being set to default above then, why do we need to set it again? Shouldn't we fix the visibility code to not touch other params?

ZitaNemeckova: @h-kataria When a chart changes params there's no reload so the initial value is lost because it's not in the URL. I could give the chart the intial value and make it save it to URL but it seemed complicated. Or I could enfoce the default value to be in the URL from the begging.

h-kataria: let's change  https://github.com/RedHatInsights/tower-analytics-frontend/blob/ec24b1e344c7c2ccc9b0d4ef2c11785216757be1/src/Containers/Reports/Layouts/Standard/ReportCard.tsx#L184 and https://github.com/RedHatInsights/tower-analytics-frontend/blob/ec24b1e344c7c2ccc9b0d4ef2c11785216757be1/src/Containers/Reports/Layouts/Standard/ReportCard.tsx#L205 to `chartParams.chartType` that should fix the issue

ZitaNemeckova: @h-kataria It helped. Thanks :) It's ready for review.

skochay330 (approved):
Tested in UI, and the code LGTM!

ZitaNemeckova:
@h-kataria Fixed. Thanks for review :)

h-kataria (approved):

PR #801, opened 2022-05-24
Make sure to only call fetchReport when report is selected

h-kataria:
This eliminates an extra API call to get report

https://issues.redhat.com/browse/AA-1052

@skochay330 please review/verify


skochay330 (approved):
LGTM 👍🏼

PR #806, opened 2022-06-06
Made changes to hide expiry date field

h-kataria:
- Made changes to hide expiry date field when external users are not provided
- Added a note below expiry date field to specify that the expiry only applies to external users only

https://issues.redhat.com/browse/AA-1145
Screen Shot 2022-06-06 at 12 23 46 PM
5

/cc @trahman73 @appuk


skochay330 (approved):
Verified in UI, and code looks good!

PR #809, opened 2022-06-14
Add current page and total Savings to ROI

ZitaNemeckova:
Fixes https://issues.redhat.com/browse/AA-1014

Fixes :
- The Total Savings figure in big green font reflects savings only for that page. Total Savings is always supposed to show the savings across all templates, as in, all pages. Lets have 2 displays here in order to make things very clear for the user to understand and to comply with the older version of the Automation Calculator screen where we ALWAYS displayed the overall total: 
Grand Total : This will display the entire total of the Savings across ALL pages 
Page Total : This will be a summation of all the Savings on the templates on the given page
 

- Individual page totals are not showing accurate numbers.  Note the page total number on page1. Now navigate to page2 and come back to page1. You will see a different total than what was displayed before. Hitting refresh on page1 fixes the total. Needs to be resolved ASAP since it is creating confusion and chaos on the screen

Related but independent of https://gitlab.cee.redhat.com/automation-analytics/pdf-generator/-/merge_requests/108

Before:
Screenshot 2022-06-14 at 13 06 14

After:
Screenshot 2022-06-14 at 13 04 36

### DIFF
--- a/src/Containers/Reports/Layouts/AutomationCalculator/TemplatesTable/Row.tsx
+++ b/src/Containers/Reports/Layouts/AutomationCalculator/TemplatesTable/Row.tsx
@@ -63,7 +63,6 @@ const Row: FunctionComponent<Props> = ({
     window.localStorage.setItem(id.toString(), value ? 'true' : 'false');
     setIsExpanded(value);
   };
-  console.log(template);
 
   return (
     <>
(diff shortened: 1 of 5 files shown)


src/Containers/Reports/Layouts/AutomationCalculator/TemplatesTable/Row.tsx
```diff
@@ -63,6 +63,7 @@ const Row: FunctionComponent<Props> = ({
     window.localStorage.setItem(id.toString(), value ? 'true' : 'false');
     setIsExpanded(value);
   };
+  console.log(template);
```
h-kataria, line 66: @ZitaNemeckova this can be removed

h-kataria (approved):

PR #811, opened 2022-06-21
Replace data-testid with data-cy

ZitaNemeckova:
cc @fullsushidev

### DIFF
--- a/src/Charts/BarChart.js
+++ b/src/Charts/BarChart.js
@@ -208,7 +208,7 @@ class BarChart extends Component {
   }
 
   render() {
-    return <div id={this.props.id} data-testid="barchart" />;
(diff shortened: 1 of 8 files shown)


fullsushidev:
thank you @ZitaNemeckova

fullsushidev:
@ZitaNemeckova Could we maybe use only one id for tests in general? either jest or cypress?

ZitaNemeckova:
@fullsushidev I'll have a look into removing `data-testid`.

ZitaNemeckova:
@fullsushidev I replaced `data-testid`. Harpreet agreed that `data-cy` is better.

fullsushidev:
@ZitaNemeckova @h-kataria Could we merge this PR?

PR #815, opened 2022-06-22
Add data-cy to toolbar items

ZitaNemeckova:
All items should have an unique `data-cy` so Cypress can click on all of them using `daata-cy`. 

Only problem is pagination. It's a PF component and I'm only able to add `data-cy` to a whole thing not the buttons.

cc @fullsushidev


fullsushidev:
Thank you @ZitaNemeckova.
About the PF components, I think some data-cy in the parent div or other parent component would be enough.

h-kataria (approved):

PR #820, opened 2022-06-30
Update cypress tests to work on stage

fullsushidev:
- Bump cypress to v9.7 (since v10 has breaking changes)
- Bump and installs plugins

Previously, all tests were failing and now it's fixed with:
- Update login page ids
- Include command to accept the cookies

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -23,6 +23,7 @@
 //
 // -- This will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+import 'cypress-wait-until';
 
 Cypress.Commands.add('getBaseUrl', () => Cypress.env('baseUrl'));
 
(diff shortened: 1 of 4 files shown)


fullsushidev:
cc @ZitaNemeckova @h-kataria @skochay330 @appuk

cypress/support/commands.js
```diff
@@ -66,5 +94,7 @@ Cypress.Commands.add('loginFlow', () => {
     );
   }
 
+  cy.acceptCookiesDialog();
+  cy.wait(5000)
```
ZitaNemeckova, line 98: We should wait on a request or an element instead.
Something like
`cy.get('[data-cy=correct_elem]', { timeout: 5000 }).should('be.visible');`

cypress/support/commands.js
```diff
@@ -44,14 +45,41 @@ Cypress.Commands.add('clearFeatureDialogs', () => {
...
+      .find('a.call')
+      .should('be.visible')
+      .click(true)
+  }
+
+  cy.waitUntil(() => acceptCookies());
+
+});
+
+// cy.waitUntil(() => cy.getCookie('token').then(cookie => cookie.value === '<EXPECTED_VALUE>'));
```
ZitaNemeckova, line 74: Please remove commented code :)

fullsushidev: oops, done!

cypress/support/commands.js
fullsushidev: I totally agree with you, but sadly I couldn't make it work anywhere yet.
I tried to include this in different places like:

```
  const getIframeDocument = () => {
    return cy
      .get('iframe', { timeout: 5000 })
      .its('0.contentDocument').should('exist')
  }
```
or

```
  const getIframeBody = () => {
    return getIframeDocument()
      .its('body', { timeout: 5000 }).should('not.be.undefined')
      .then(cy.wrap)
  }
```
and so on, and nothing prevented the test to run faster than the iframe.
I also tried to use `.then`, and the cypress plugin `waitUntil` but still with no success. :(
  
Another possibility is to use aliases and do something like `cy.wait(@cookieButton)`
  
Either way, it's taking me too much time to find a better solution.
  
What do you think if we move this way and remove this code once we have a better and final solution (maybe following your suggestion with the code from hub team).
I'm adding a comment and to-do tag in the code.

ZitaNemeckova: Ok, sounds good to address it in follow-up PR :)

PR #822, opened 2022-07-06
Fix: Add a redirect for the root route to the Routes

brumik:
The root route used the default route which is a not found
404 error screen. To make the user experience pleasant lets
redirect this route to the clusters page.


brumik:
I dont have the backend setup, so please try it out before merging with backend too. (And lets see if the tests are passing)

dmzoneill (approved):

PR #823, opened 2022-07-06
Fix: Add a redirect for the root route to the Routes

dmzoneill:
Moved redirect after the actual routes

### DIFF
--- a/src/Routes.tsx
+++ b/src/Routes.tsx
@@ -57,8 +57,6 @@ export const Routes: FunctionComponent<Record<string, never>> = () => {
     <Switch>
       {/* Catch urls with the trailing slash and remove it */}
       <Redirect from="/:url*(/+)" to={pathname.slice(0, -1)} />
-      {/* Redirect the root path to the clusters so it does not give 404. */}
-      <Redirect from="/" to={Paths.clusters} />
       {/* Render the valid routes */}
       {Object.keys(components).map((key) => (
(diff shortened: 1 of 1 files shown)


h-kataria (approved):

h-kataria (changes requested):
can you make following changes, to fix the issue

```
diff --git a/src/Routes.tsx b/src/Routes.tsx

index 9859ee26..bbe3e672 100644

--- a/src/Routes.tsx

+++ b/src/Routes.tsx

@@ -57,12 +57,12 @@ export const Routes: FunctionComponent<Record<string, never>> = () => {

     <Switch>

       {/* Catch urls with the trailing slash and remove it */}

       <Redirect from="/:url*(/+)" to={pathname.slice(0, -1)} />

-      {/* Redirect the root path to the clusters so it does not give 404. */}

-      <Redirect from="/" to={Paths.clusters} />

       {/* Render the valid routes */}

       {Object.keys(components).map((key) => (

         <InsightsRoute key={key} path={key} component={components[key]} />

       ))}

+      {/* Redirect the root path to the clusters so it does not give 404. */}

+      <Redirect from="/" to={Paths.clusters} />

       {/* Finally, catch all unmatched routes and render 404 */}

       <Route>

         <Error404 body="Sorry, we could not find what you were looking for. The page you requested may have been changed or moved." />
```

PR #825, opened 2022-07-07
[cypress] Update login URL assertion

fullsushidev:
- also comment out iframe code
- will be used locally only if necessary


fullsushidev:
@appuk I'm still working on my other PR, so I just opened this one to unblock the team.

dmzoneill (approved):

PR #845, opened 2022-08-10
Cypress pagination tests

fullsushidev:
**The current implementation is a simpler version of PR #849 plan** in a slightly generic way.

On this PR, the Job explorer will be used as a model to develop the way we will test pagination for all the reports who basically needs to:
- [x] Test if the user can navigate through all the pages;
- [x] Test if the user can change the number of items shown on the list;

-- 

Depends on #848

### DIFF
--- a/cypress/support/e2e.js
+++ b/cypress/support/e2e.js
@@ -15,8 +15,8 @@
 
 // Import commands.js using ES2015 syntax:
 import './commands'
-import './pagination'
 import './login'
+import './pagination'
 
 // Returning false here prevents Cypress from failing the test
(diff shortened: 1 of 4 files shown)


cypress/support/e2e.js
```diff
@@ -15,6 +15,9 @@
 
 // Import commands.js using ES2015 syntax:
 import './commands'
+import './pagination'
+import './login'
+import './pagination'
```
h-kataria, line 20: duplicate import statement

h-kataria (approved):
LGTM

PR #849, opened 2022-08-22
Extend cypress pagination tests

fullsushidev:
This PR updates the pagination tests (#845) to work for most of reports pages.

--

TLDR: Since the `data-cy` wasn't possible to be added to every single component to be tested, a bigger solution is needed, and to make the test more clean, several helper commands were created to be reused among the reports pages and any other page that uses the same pagination component.

--

The final idea is to develop the following plan:
- Get the amount of pages available;
    - Iterate and compare the amount of pages and how many times the pagination arrow can be clicked;
- Return the page's dropdown;
  - Validate all existing values on the dropdown;
  - Test clicking on each option and assert if the table is correctly updated;
  - A next step would be to validate pages when the total rows are lower than the pagination options;

Maybe it would be nice to extend this tests and make it work for only a page per test run, where the page is chosen randomly.

-- 

From the original plan, the following were implemented (for top and bottom pagination components):

- [x] Navigate trough pages using pagination arrows,
- [x] Asserting if the arrows are enabled/disabled,
- [x] Toggle the items per page and assert it has all the options,
- [x] Assert the amount of rows with the items per page selected (default value and maximum value)

-- 

depends on #845

### DIFF
--- a/cypress/e2e/Global.spec.js
+++ b/cypress/e2e/Global.spec.js
@@ -65,7 +65,7 @@ describe('Insights smoketests', () => {
     cy.get('[data-cy="page_component"]').should('exist');
   };
 
-// TODO: include assetion maybe with snapshots
+  // TODO: include assetion maybe with snapshots
   it('can open each page without breaking the UI', () => {
     cy.visit(orgsUrl);
     checkPageHasNoErrors();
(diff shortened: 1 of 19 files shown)


fullsushidev:
Closing this one to try to fix the PR_check

PR #859, opened 2022-09-01
Extend cypress pagination tests

fullsushidev:
This PR updates the pagination tests (https://github.com/RedHatInsights/tower-analytics-frontend/pull/845) to work for all pages.

--

TLDR: Since the data-cy wasn't possible to be added to every single component to be tested, a bigger solution is needed, and to make the test more clean, several helper commands were created to be reused among the reports pages and any other page that uses the same pagination component and tables.

--

The final idea is to develop the following plan:

- Get the amount of pages available;
- Iterate and compare the amount of pages and how many times the pagination arrow can be clicked;
- Return the page's dropdown;
- Validate all existing values on the dropdown;
- Test clicking on each option and assert if the table is correctly updated;
- A next step would be to validate pages when the total rows are lower than the pagination options;
- Maybe it would be nice to extend this tests and make it work for only a page per test run, where the page is chosen randomly.

--

From the original plan, the following were implemented (for top and bottom pagination components):

- [x]  Navigate trough pages using pagination arrows,
- [x]  Asserting if the arrows are enabled/disabled,
- [x]  Toggle the items per page and assert it has all the options,
- [x]  Assert the amount of rows with the items per page selected (default value and maximum value)

--

depends on https://github.com/RedHatInsights/tower-analytics-frontend/pull/845

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Automation Analytics provides data analytics for Ansible Tower that provides vis
 1. `npm ci` - install dependencies from the lockfile
 2. get the backend running: [automation analytics backend](https://gitlab.cee.redhat.com/automation-analytics/automation-analytics-backend)
 3. `npm start` - starts standalone: webpack serves the files alongside with insights, rbac and keycloak.
-4. Go to `http://localhost:1337/beta/ansible/insights` and use the admin/admin credentials to login.
+4. Go to `http://localhost:1337/ansible/automation-analytics` and use the admin/admin credentials to login.
 
 #### Developing against a deployed backend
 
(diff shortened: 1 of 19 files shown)


h-kataria (approved):

README.md
```diff
@@ -11,7 +11,7 @@ Automation Analytics provides data analytics for Ansible Tower that provides vis
 1. `npm ci` - install dependencies from the lockfile
 2. get the backend running: [automation analytics backend](https://gitlab.cee.redhat.com/automation-analytics/automation-analytics-backend)
 3. `npm start` - starts standalone: webpack serves the files alongside with insights, rbac and keycloak.
-4. Go to `http://localhost:1337/beta/ansible/insights` and use the admin/admin credentials to login.
+4. Go to `http://localhost:1337/ansible/automation-analytics` and use the admin/admin credentials to login.
```
ZitaNemeckova, line 14: Do we really want to remove `beta`?

fullsushidev: Actually no, just fixed it

PR #873, opened 2022-10-25
[AA-1412] Display multiple job statuses

skochay330:
This PR addresses a bug in the savings planner that caused the page to break when there were multiple job statuses reported for one template. The UI can now handle a scenario where the API could send multiple statuses and as a result, a separate tag will appear for each status. 

A tooltip was also added to the "Last job status" label in order to display the date of the last reported status. 

Jira issue: https://issues.redhat.com/browse/AA-1412 

Screen Shot 2022-10-25 at 2 43 26 PM


h-kataria:
changes look good to me, just a quick question for @AparnaKarve does the order of status matter?

h-kataria (approved):

PR #880, opened 2022-11-09
[AA-1323] Carry all filters to host anomaly scatterplot

skochay330:
When a specific bar is clicked in the host anomaly bar chart, ALL filters are carried to the host anomaly scatterplot report. Previously, only template and date range were being carried. 

Jira issue: https://issues.redhat.com/browse/AA-1323 

Before: 
Screenshot 2022-11-09 at 3 14 54 PM

After:
Screenshot 2022-11-09 at 3 13 39 PM


ZitaNemeckova (approved):
Works! LGTM 👍 Great job!

PR #894, opened 2022-12-06
Reorganize and update cypress

fullsushidev:
- [x] Bump cypress version
- [x] Include stage and ephemeral fixtures
- [x] move tests to new folders for each env
- [x] fix PR check

PS: I'll leave the remaining tests to be fixed in another PR.


fullsushidev:
I'll leave the remaining tests to be fixed in another PR.

ZitaNemeckova (approved):

PR #900, opened 2022-12-14
Fix reports preview not changing reports

ZitaNemeckova:
**PR content:**
- removes all logic that changes URL params directly
- allows update function to take `removeDefault` params that says that all params in namespace default should be removed

**Problem:**

Before we allowed unlimited number of API calls so we could make changes to URL multiple times per one user action. Now we have limited the number of API calls so if we do multiple changes to URL per one user action only one of those changes will be processed and rest will be discarded. For the Reports page when a report is selected for preview it updates `selected_report` param and removed all default params. Only the second one was done. This PR does those two changes at once.

**Steps to reproduce:**
- Go to Reports page
- Try to select different report

**Before:**

First report is displayed 

**After:**

Selected report is displayed

_Note: This PR needs a lot of testing around selecting a report to be previewed in Reports page. Also testing that filtering in the toolbars is still working would be good._

### DIFF
--- a/src/Containers/Reports/List/List.tsx
+++ b/src/Containers/Reports/List/List.tsx
@@ -1,7 +1,7 @@
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-nocheck
 import React, { FunctionComponent, useEffect, useState } from 'react';
-import { Link, useHistory } from 'react-router-dom';
+import { Link } from 'react-router-dom';
 import Main from '@redhat-cloud-services/frontend-components/Main';
 import {
   PageHeader,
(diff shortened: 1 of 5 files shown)


h-kataria:
code changes look good, glad we removed the ugly hack.

fullsushidev:
@ZitaNemeckova You can try to run cypress again as soon as this PR #899 is merged!

ZitaNemeckova:
@h-kataria @skochay330 it should be fixed and ready to be merged.

Use case:
- change anything in preview
- change any filter in the page header so the first report is different

Before: 
Error page

After:
It works

skochay330 (approved):
Works great!

PR #910, opened 2023-01-10
Daoneill fix cypress cookie accept

dmzoneill:


fullsushidev (approved):
LGTM

(the cookies code doesn't work in cypress >10 so this will be updated in another PR)

ZitaNemeckova (approved):

PR #915, opened 2023-01-26
Show first 500 entries in drop down

h-kataria:
Show filter and show entries from all the entries that were received from backend API. We're continuing to show 500 entries only to avoid performance issues while rendering drop-down with 5000 entries.

https://issues.redhat.com/browse/AA-1554

Backend PR https://gitlab.cee.redhat.com/automation-analytics/automation-analytics-backend/-/merge_requests/678

### DIFF
--- a/src/Components/Toolbar/Groups/ToolbarInput/Select.tsx
+++ b/src/Components/Toolbar/Groups/ToolbarInput/Select.tsx
@@ -51,7 +51,7 @@ const Select: FunctionComponent<Props> = ({
 }) => {
   const [expanded, setExpanded] = useState(false);
   const options = optionsForCategories[categoryKey];
-  const selectOptions = nonTypedSelectOptions.map(
+  let selectOptions = nonTypedSelectOptions.map(
     ({ key, value, description }) => ({
       key: key?.toString(),
       value: value?.toString(),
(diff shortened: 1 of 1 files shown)


src/Components/Toolbar/Groups/ToolbarInput/Select.tsx
```diff
@@ -76,9 +80,11 @@ const Select: FunctionComponent<Props> = ({
...
     if (textInput === '') return renderValues(selectOptions);
     return renderValues(
-      selectOptions.filter(({ value }) =>
-        value.toString().toLowerCase().includes(textInput.toLowerCase())
-      )
+      selectOptionsMasterCopy
+        .filter(({ value }) =>
+          value.toString().toLowerCase().includes(textInput.toLowerCase())
+        )
+        .slice(0, 50)
```
h-kataria, line 87: @skochay330 does this need to be 500?

AparnaKarve: @h-kataria When we filter for a specific string, we do not want to return results more than 50.

If the result is not in the first 50 items, then thats the cue for the user to refine their search further.

AparnaKarve: Usually with real data (where templates aren't named like in our test data, for e.g. - template_1, template_2, template_abc), there is a high likelihood that the filter results would be available in the first 50 items.

skochay330: @AparnaKarve ^^

skochay330 (approved):

PR #934, opened 2023-04-06
[AA-1649] custom date range selection fix

daphnemaeve:
Jira issue: https://issues.redhat.com/browse/AA-1649

Issue with custom date range error seems to be caused by formatDate only taking Date objects, while some .js files pass strings to the function. This fix changes the function to take in a Date object or a string, and expands the test to also check that Date objects are formatted correctly.


h-kataria (commented):
The issue may have been caused by the version of the moment package, I see that it was recently updated, we can merge this PR and test it on stage to see if the issue gets resolved. Another way we can try is to revert the moment package version locally and try to recreate the issue on localhost.

h-kataria (approved):

PR #945, opened 2023-04-25
Use preview as well as beta in application's URL

ZitaNemeckova:
- beta/preview both works

Test with `stage`. 

Beta toggle sets it to beta so change it in URL to preview and see that all links in nav bar and application have preview in the URL and work.


skochay330 (approved):
Tested against stage, LGTM

PR #953, opened 2023-05-30
Add missing data-cy

ZitaNemeckova:


fullsushidev (approved):
Looks good to me.
The remaining failures you can leave as is. I believe it will be better to use a different strategy, otherwise it will fail constantly.

PR #958, opened 2023-06-08
[AA-1732] Templates by org clickable linking fix

skochay330:
When navigating from templates by organization to templates explorer by clicking on the numbers in the table, the templates explorer report will now be filtered by the organization id rather than by a list of templates. Any additional filters that were applied in the templates by organization report will also be carried over to templates explorer, if the filter option exists there. 

Jira issue: https://issues.redhat.com/browse/AA-1732

Before: 
Screenshot 2023-06-08 at 5 02 00 PM

After:
Screenshot 2023-06-08 at 5 04 41 PM


daphnemaeve (approved):
Looks good to me!

fullsushidev (approved):

fullsushidev:
yes it is @skochay330

PR #975, opened 2023-08-16
Fix build out of memory issues

Hyperkid123:
### Changes
- update dependencies to the latest compatible minor versions
- exclude large modules from module federation dependency sharing

**note** The dependency changes will likely cause some regressions. Someone who understands the codebase should go through the app and include fixes if required.

The out-of-memory issue was caused by an extremely large amount of JS modules the web pack was trying to parse/bundle. By removing dependencies from the shared modules we force Webpack to ignore a major portion of the dependencies modules.

I don't have a comparison of the build because I was unable to collect metrics without the changes dues to memory issues. This is what it looks like with the fix:

screenshot-127 0 0 1_8888-2023 08 16-10_53_26

The final build includes around 3.8K individual JS modules. When we removed moves of the application during testing and managed to get output like this, we got to almost 10K modules without the fixes. I can only image we would be getting close to 20k for the entire application. Such large amounts of individual modules that webpack has to cover, not including the tree-shaken modules can cause major memory issues.


dmzoneill:
merged in !977

PR #984, opened 2023-08-23
[AA-1815] Fix redirect to job explorer

daphnemaeve:
Jira issue: https://issues.redhat.com/browse/AA-1815

Should fix bug on stage and prod that is breaking redirect to job explorer from automation calculator. 
Error occurs when user clicks on a template from table, and is navigated to a page with 404 error.
Added a replace statement to remove extra '/' in slug, which works locally.


skochay330 (approved):
Tested locally in stage env, LGTM

PR #986, opened 2023-09-01
Bumping node to v18 (PR check fix)

fullsushidev:
Attempt to upgrade node versions for PR check.

fixes: https://issues.redhat.com/browse/AA-1835


fullsushidev:
node 18 is compatible with Linux distributions based on glibc 2.28, meaning that we need ubuntu >20.04

PR #990, opened 2023-09-22
[AA-1791] fix api calls in report tiles

daphnemaeve:
Jira issue: https://issues.redhat.com/browse/AA-1791

This PR fixes the excessive API calls made when clicking on a card to preview a report in the reports view. This was tested by putting a console.log statement in the useEffect in src/Containers/Reports/List/List.tsx, on line 97, as well as by looking at the API calls made in the Network tab when inspecting the page.


src/Containers/Reports/List/List.tsx
```diff
@@ -8,6 +8,7 @@ import {
   ButtonVariant,
   Card,
   CardActions,
+  CardBody,
```
skochay330, line 11: We can remove this, it is never used. Otherwise, LGTM!

skochay330 (changes requested):

skochay330 (approved):

PR #1000, opened 2023-10-11
[AA-1798] Fix redirect to job explorer from clusters and savings planner pages

skochay330:
This PR fixes redirects to the job explorer page from the clusters page and the savings planner details tab. 

Related to https://github.com/RedHatInsights/tower-analytics-frontend/pull/984 where the same thing was done for the redirect from automation calculator. 

Jira issue: https://issues.redhat.com/browse/AA-1798


skochay330:
Ignore codecov failure, we are not able to use jest to check redirects as of right now.

daphnemaeve (approved):

PR #1012, opened 2023-11-08
[AA-1927] Fix broken redirects in stage/prod

skochay330:
This PR fixes redirects from charts and savings planner list view to the job explorer page.

Jira issue: https://issues.redhat.com/browse/AA-1927


skochay330:
Ignore codecov failure, we are not able to use jest to check redirects as of right now.

PR #1013, opened 2023-11-14
Github action sync to master branch on push

dmzoneill:
Sync devel to master.
the frontend operator pipeline is designed to run on main/master branch push/merge events.
Because patches are landing onto devel, the pipeline won't run, and the latest updates won't appear on stage.
This github action automatically syncs devel to master on merge.

### DIFF
--- a/.github/workflows/devel.yml
+++ b/.github/workflows/devel.yml
@@ -69,4 +69,6 @@ jobs:
           FROM_BRANCH: "devel"
           TO_BRANCH: "master"
           PULL_REQUEST_AUTO_MERGE_METHOD: rebase
+          PULL_REQUEST_TITLE: "Sync devel -> master"
+          PULL_REQUEST_BODY: "Automatically syncing devel with master"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/devel.yml
```diff
@@ -50,3 +50,25 @@ jobs:
...
+  sync-branches:
+    runs-on: ubuntu-latest
+    name: Syncing branches
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
```
fullsushidev, line 63: why node version 12?

fullsushidev (changes requested):

.github/workflows/devel.yml
fullsushidev, line 63: Shouldn't we use at least v18?

fullsushidev (dismissed):
is there a specific reason why are we not using the latest versions of the actions?

e.g.:
```
codecov/codecov-action@v3
actions/checkout@v4
actions/setup-node@v4
```

.github/workflows/devel.yml
dmzoneill: This has nothing to do with our code, it has to do with the code in the github action.
The author of the action indicates version 12.  So we use version 12.

h-kataria (approved):

PR #1015, opened 2023-11-27
Resource match routes. Remove paths.

adamrdrew:
This patch makes the frontend resource name match the desired open shift route names, and then removes the redundant paths.


himdel:
Looks like this was fixed in db2a16be975b8b3669614399d4c310e2b6090bd2, closing

PR #1046, opened 2024-09-18
PF5 - src/framework fixups, unused export cleanup

himdel:
Remove dead files

DeleteButton - imported directly, not via index
ROITopTemplates - unused since 2022, remove
getBaseName - was used above AppEntry, not needed
framework/* - context never initialized

and fix typescript/linter issues under src/framework, use pf-v5* classnames, event in click handlers, updated imports


ZitaNemeckova (approved):
Tested in UI. LGTM 👍

PR #1049, opened 2024-09-20
Fix chart toolbar in preview

MilanPospisil:


MilanPospisil:
image

The tags were shifted to right and switch between report types was not displayed at all.

himdel (approved):

PR #1058, opened 2024-09-24
Fix layout of dropdown items

prat98:
Fixes the layout of the dropdown filters in notifications and also fixes the issue with the click target being only on the label


himdel (approved):

himdel:
Fixed :tada:

PR #1149, opened 2026-02-23
PF6 Migration

justinorringer:
- Branched from work done by Pratyush
- Removes shims in place (pf6 and notifications shims), fixed to either use frontend components.
- Updates the frontend yaml with FEO migration (needs to be enabled once initially pushed to stage).
- Partially fixed cypress config, the tests are in rough shape with or without these changes.
- Updates fec-config to match our modern process of running locally, same as all frontends 'npm run start'.


cypress/e2e/AutomationCalculator.spec.js
```diff
@@ -7,116 +7,134 @@ const waitToLoad = () => {
...
+          expect(newSavingsValue).not.to.eq(originalSavingsValues[index]);
+        });
+      }
 
-    cy.get('#manual-cost').clear();
-    waitToLoad();
-    cy.get('#manual-cost').should('have.value', '0');
+      cy.get('#manual-cost').type('5');
+      waitToLoad();
+      // TODO explain trailing 0
```
catastrophe-brandon, line 68: Bug or feature? The world may never know!

cypress/e2e/AutomationCalculator.spec.js
```diff
@@ -7,116 +7,134 @@ const waitToLoad = () => {
...
+        });
+    });
+  });
+
+  it('can change automated cost', () => {
+    // Skip test if no data available
+    cy.get('body').then($body => {
+      if ($body.find('.pf-v6-c-empty-state__content').length > 0) {
+        cy.log('Empty state found - skipping automated cost test');
+        expect(true).to.be.true;
```
catastrophe-brandon: Is this assertion acting as a placeholder for some future assertion check? I think `return` statement alone would work for skipping the test.

cypress/e2e/PatternFly6Components.spec.js
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  it('Verify switch components use PF6 classes', () => {
+    cy.visit('/organization-statistics');
+    cy.wait(2000);
+    
+    // Look for switch toggles
+    cy.get('body').then($body => {
+      const hasV6Switch = $body.find('.pf-v6-c-switch__toggle').length > 0;
+      const hasV5Switch = $body.find('.pf-v5-c-switch__toggle').length > 0;
```
catastrophe-brandon: Is this because there are some lingering V5 elements on the page?

src/Components/Toolbar/Groups/CategoryDropdown.tsx
```diff
@@ -1,8 +1,18 @@
...
   Select,
   SelectOption,
-  SelectVariant,
-} from '@patternfly/react-core/deprecated';
+  SelectList,
+  MenuToggle,
+} from '@patternfly/react-core';
+
+// SelectVariant enum for backward compatibility
+const SelectVariant = {
```
catastrophe-brandon: It seems like we have an enum like this in a few different places. Would it make sense to try to reuse it, or are the differences subtle enough to warrant different enums?

src/Components/Toolbar/Groups/FiltersCategoriesGroup.tsx
```diff
@@ -1,4 +1,11 @@
-import { SelectOptionProps } from '@patternfly/react-core/deprecated';
+// SelectOptionProps type for compatibility
+export type SelectOptionProps = {
```
catastrophe-brandon: Similar comment here. I see `SelectOptionProps` in a few places.  Since we're not using the `@patternfly/react-core/deprecated` anymore and creating our own version for backward compat, maybe we could have our own "deprecated" locally to import it from?

src/framework/PageTable/PageToolbar.tsx
```diff
@@ -550,28 +567,37 @@ function ToolbarSelectFilter(props: {
...
+            ) : (
+              <span style={{ opacity: 0.7 }}>{props.placeholder}</span>
+            )}
+          </MenuToggle>
+        )}
+      >
+        <SelectList>
+          {options.map((option) => (
+            <SelectOption
+              id={option.value}
```
catastrophe-brandon, line 591: Is there any possibility that options might contain a duplicate value?

catastrophe-brandon (changes requested):
One small V5 mention in one of my comments could use a look. Once that's addressed, lgtm!

prat98 (approved):

PR #1151, opened 2026-02-26
FEO config enabled flag; build-tools submodule; sample konflux pipelines

justinorringer:


.tekton/automation-analytics-frontend-pull-request.yaml
karelhala, line 1: We shouldn't include these in here, because it messes up Konflux onboardment. If you want to adjust how the app is built we can change build_deploy.sh script.

karelhala (commented):
Looking good! Just a small problem wrt Konflux configs.

.tekton/automation-analytics-frontend-pull-request.yaml
prat98: Removed the commit adding the pipeline files

justinorringer: Thanks Pratyush, should be good now

PR #1184, opened 2026-04-29
Overrides transitive axios dependency to resolve CVE

nicklemmon:
Adds `"axios"` to `"overrides"` to force dependency resolution to a more secure version.


daphnemaeve:
Verified successful build, LGTM!

daphnemaeve (approved):

PR #1238, opened 2026-08-11
feat(calculator): add default manual time input to Automation Calculator

dmzoneill:
[AAP-84046] Add "Default manual time per template (minutes)" input alongside existing Manual cost / Automation cost inputs. 

Saves via existing roi_cost_effort_data endpoint with new default_manual_effort_minutes field. Positive integer validation, RBAC-controlled editability.

[AAP-84046]: https://redhat.atlassian.net/browse/AAP-84046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/e2e/AutomationCalculator.spec.js
+++ b/cypress/e2e/AutomationCalculator.spec.js
@@ -136,6 +136,28 @@ describe('Automation Calculator page', () => {
     });
   });
 
+  it('can change default manual effort', () => {
+    // Skip test if no data available
+    cy.get('body').then(($body) => {
+      if ($body.find('.pf-v6-c-empty-state__content').length > 0) {
+        cy.log('Empty state found - skipping default manual effort test');
+        return;
+      }
+
+      cy.get('#default-manual-effort').clear();
+      waitToLoad();
+      cy.get('#default-manual-effort').should('have.value', '0');
+
+      cy.get('#default-manual-effort').type('60');
+      waitToLoad();
+      cy.get('#default-manual-effort')
+        .invoke('val')
+        .then((val) => {
+          expect(Number(val)).to.be.greaterThan(0);
+        });
+    });
+  });
+
   it('can change visibility', () => {
     let originalTotalSavingsValue = cy
       .getByCy('total_savings')
--- a/src/Api/types.ts
+++ b/src/Api/types.ts
@@ -70,6 +70,7 @@ export interface saveROIParams {
   currency: string;
   hourly_manual_labor_cost: number;
   hourly_automation_cost: number;
+  default_manual_effort_minutes?: number;
   templates_manual_equivalent: {
     template_id: number;
     effort_minutes: number;
--- a/src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
+++ b/src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
@@ -181,7 +181,9 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
       currency: 'USD',
       hourly_manual_labor_cost: manualCost,
       hourly_automation_cost: automationCost,
-      default_manual_effort_minutes: manualEffort,
+      ...(typeof manualEffort === 'number' && !isNaN(manualEffort)
+        ? { default_manual_effort_minutes: manualEffort }
+        : {}),
       templates_manual_equivalent: updatedDataApi,
     };
   };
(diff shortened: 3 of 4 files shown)


drodowic:
## Code Review (AAP-84046)

### Summary 
Reviewed with multi-agent adversarial analysis. Found a few things worth fixing before merge — grouped by priority.

### Blocking

1. **Legacy `''` value corrupts the field on unrelated saves** — `AutomationCalculator.tsx:335`
   When `api.result.cost.default_manual_effort_minutes` is missing (older report / backend hasn't deployed the field yet), state hydrates to `''` via `?? ''`. That `''` then gets sent as `default_manual_effort_minutes` on *every* subsequent save — including edits to Manual cost or Automation cost that have nothing to do with this field. That risks stomping a backend field typed `number` with an empty string.
   → Hydrate to `undefined` instead of `''`, and have `getROISaveData` omit the key entirely when there's no valid number, rather than always including it.

2. **Invalid input saves silently instead of being rejected** — `CalculationCost.tsx:102`
   `validPositiveInt` returns `NaN` for negative/decimal input, but `onChange` calls `setFromCalculation` unconditionally — nothing stops the save. `JSON.stringify(NaN)` → `null`, so an invalid value is persisted as `null` rather than rejected, and the field just goes blank with no error shown. This also means the AC's "Input validates for positive integer values" and "Visual feedback on save success/failure" aren't actually met for this path.
   → Short-circuit `onChange` when the value is `NaN`: skip the save call and show an inline error (`validated='error'` + helper text) instead of letting it flow through.

3. **No cypress/unit tests added** — AC explicitly calls for "Unit / integration (cypress)," but no test file changed in this PR. The existing spec (`cypress/e2e/AutomationCalculator.spec.js`) has parity coverage for `#manual-cost` / `#automation-cost` (clear → save → assert, type → save → assert) but nothing for `#default-manual-effort`.
   → Mirror the existing cost-field tests for the new input, including the invalid-input and save-failure-notification paths.

### Should fix

4. **No upper bound on the value** — `CalculationCost.tsx:18`. AC asks for "positive integer, *reasonable range*," but `validPositiveInt` has no max, so `999999999` saves fine.

5. **Four prettier violations will fail lint/CI as-is** — run `npx prettier --write` on both changed files:
   - `CalculationCost.tsx:87` (label wrapped across lines)
   - `CalculationCost.tsx:105` (missing trailing comma)
   - `AutomationCalculator.tsx:207-210` (ternary not collapsed)
   - `AutomationCalculator.tsx:336` (missing trailing comma)

### Nits

6. Label text is "Default manual time per template" — Jira spec says "Default manual time per template (minutes)." Unit currently only shows as the small "min" suffix in the input group, not in the label itself (weaker for screen readers).
7. `validPositiveInt` accepts `0`, so it's really validating non-negative integers, not strictly positive ones — name is a bit misleading.
8. `default_manual_effort_minutes?: number` in `src/Api/types.ts:73` doesn't match the actual runtime type (`number | string | undefined`, can be `''`/`NaN`) — only hidden by the `as any` cast on the save payload.
9. `humanVarNames[varName] || varName` fallback (`AutomationCalculator.tsx:216`) is unreachable — every call site already passes a known key. Fine to drop.

Once #1–#3 are addressed I think this is good to go.

dmzoneill:
Thanks for the thorough review! Addressed all items in 465da2f7:

### Blocking — fixed
1. **`''` hydration** — `getROISaveData` now conditionally spreads `default_manual_effort_minutes` only when it's a valid number. Empty string / undefined / NaN are omitted from the payload entirely.
2. **NaN saves silently** — Added `isNaN(value)` guard at the top of `updateCalculationValues`. This protects **all three** cost inputs (manual cost, automation cost, and the new field), not just the new one — the pre-existing fields had the same gap.
3. **No tests** — Added cypress test for `#default-manual-effort` following the existing pattern (clear → verify 0 → type → verify saved).

### Should fix — fixed
4. **No upper bound** — Added `MAX_PG_INT` (2,147,483,646) cap to validation. Values exceeding PostgreSQL integer max are now rejected as NaN.
5. **Prettier** — Fixed all 4 violations, both files pass `--check`.

### Nits — fixed
6. **Label** — Added "(minutes)" to label text for screen reader clarity.
7. **Function name** — Renamed `validPositiveInt` → `validNonNegativeInt` since it accepts 0.
8. **Type in types.ts** — Skipped. The `saveROIParams` type represents the API contract (`number`), which is correct. The internal state type (`number | string | undefined`) is already different. The `as any` cast at the call site bridges the gap — same pattern as the existing cost fields.
9. **Unreachable fallback** — Removed `|| varName` from `humanVarNames` lookup.

drodowic:
## Round 2 review (AAP-84046)

Two non-blocking follow-ups from the new code, neither should hold up merging:

1. **New cypress test only covers the happy path** — `cypress/e2e/AutomationCalculator.spec.js:139-159`
   The added "can change default manual effort" test checks clear→`0` and type→`60`, but nothing exercises rejected input (negative/decimal), the upper bound, or the save-failure notification — so a regression in any of those wouldn't be caught.
   → Add cases mirroring the gap: one for an invalid value being rejected, one for the upper bound, and one forcing a save failure to check the notification fires.

2. **Nice to have: the upper bound is a storage guard, not a real range** — `CalculationCost.tsx:18`
   `MAX_PG_INT` stops int4 overflow, but it doesn't stop nonsense: ~2 billion minutes (~4083 years) is still accepted as a per-template default. Jira asked for "reasonable range," which this doesn't fully deliver.
   → Optional: add a business-meaningful cap (e.g. something in the hours/days scale) instead of/in addition to the overflow guard.

drodowic (approved):

PR #1240, opened 2026-08-14
AAP-84048 - fix(calculator): replace hardcoded 60-minute fallback with configurab…

dmzoneill:
- mapApi now accepts defaultManualMinutes param; uses API value from response.cost.default_manual_effort_minutes, falls back to 60 for backward compat with older backends
- Row input reset on invalid value uses defaultManualTime prop instead of hardcoded '60'
- Pass defaultManualTime through AutomationCalculator -> TemplatesTable -> Row

Closes AAP-84048

### DIFF
--- a/src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
+++ b/src/Containers/Reports/Layouts/AutomationCalculator/AutomationCalculator.tsx
@@ -103,11 +103,11 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
   const { queryParams, setFromToolbar, setFromPagination } =
     useQueryParams(defaultParams);
 
-  const mapApi = ({ legend = [] }) => {
+  const mapApi = ({ legend = [] }, defaultManualMinutes = 60) => {
     return legend.map((el: { [key: string]: number }) => ({
       ...el,
       delta: 0,
-      avgRunTime: el.manual_effort_minutes * 60 || 3600,
+      avgRunTime: el.manual_effort_minutes * 60 || defaultManualMinutes * 60,
       manualCost: 0,
       automatedCost: 0,
       enabled: el.template_weigh_in,
@@ -142,7 +142,10 @@ const AutomationCalculator: FC<AutmationCalculatorProps> = ({
       return {
         ...response,
         items: updateDeltaCost(
-          mapApi(response.meta),
+          mapApi(
+            response.meta,
+            response.cost?.default_manual_effort_minutes ?? 60,
+          ),
           response.cost.hourly_manual_labor_cost,
           response.cost.hourly_automation_cost,
         ),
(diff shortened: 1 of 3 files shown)


drodowic:
## Code Review (AAP-84048)

### Summary
Reviewed with multi-agent adversarial analysis. Found one thing worth fixing before merge, one edge case worth a decision, plus a couple of nits.

### Blocking

1. **`defaultManualTime` goes stale after any save** — `AutomationCalculator.tsx` (new JSX line passing `defaultManualTime` to `TemplatesTable`)
   The prop is derived directly from `api.result?.cost?.default_manual_effort_minutes`, but `api.result.cost` is only ever populated on the *initial* fetch. Every subsequent save path (`updateCalculationValues`, `setDataRunTime`, `setEnabled`, `applyDefaultToAll`) goes through `update()`, which does `setValue(mapApi(res.meta, res.cost?.default_manual_effort_minutes ?? 60))` → `setApiData({ ...api.result, items })`. The fresh `res.cost` value is used correctly *inside* that one `mapApi()` call, but it's never merged back into `api.result` — the spread just carries forward the old `cost` object. So once a user edits "Default manual time" and saves, `api.result.cost.default_manual_effort_minutes` is permanently pinned to whatever it was on page load, and `Row`'s reset-on-invalid-input keeps resetting to that stale value instead of the newly saved one, until a full page reload. This reintroduces the exact class of bug the ticket is meant to fix, just moved from a hardcoded literal to a stale API snapshot.
   → There's already a `defaultManualEffort` React state value kept correctly in sync on every save (`setDefaultManualEffort(value)` in `updateCalculationValues`). Feed `defaultManualTime` from that instead of reading `api.result.cost` directly.

### Should fix

2. **A saved default of `0` breaks the invariant it's meant to enforce** — `Row.tsx:131` / `CalculationCost.tsx` (`validNonNegativeInt`)
   `validNonNegativeInt` explicitly allows `0` as a valid saved org default. Every fallback here uses `?? 60`, which only catches `null`/`undefined`, not `0`. So: admin sets the org default to `0` → a row gets an invalid input → `onBlur` resets it to `String(defaultManualTime)` = `'0'` → that value itself fails the `minutes <= 0` check the handler exists to enforce → the row silently saves a 0-minute manual time, zeroing that row's cost/savings.
   → Decide intentionally: either disallow `0` as a valid org default in `CalculationCost`, or guard the `Row` reset so it never lands on a value that fails its own validation.

### Nits

3. `mapApi`'s own `defaultManualMinutes = 60` default parameter (`AutomationCalculator.tsx`) is unreachable — both call sites always pass an explicit second argument (each with its own `?? 60`). Between this and the other two `?? 60` occurrences, the fallback value is now duplicated in three places that have to be kept in sync by hand — arguably how # 1 went unnoticed. Consider a single `DEFAULT_MANUAL_EFFORT_MINUTES` constant, or better, deriving all three from the same synced state value.
4. `response.cost?.default_manual_effort_minutes ?? 60` is optional-chained, but `response.cost.hourly_manual_labor_cost` / `hourly_automation_cost` right below it aren't. Pre-existing pattern, not introduced here, but worth tightening for consistency while touching this block.

Once item # 1 is addressed I think this is good to go.

PR #1247, opened 2026-09-01
AAP-87504 - fix(deps): override form-data to >=4.0.6 for CVE-2026-12143

daphnemaeve:
CRLF injection in the form-data npm package (through v4.0.5) allows arbitrary header injection via unescaped CR/LF/double-quote characters in field names and filenames passed to FormData#append.

form-data is a transitive dependency via axios. Adding an npm override to force resolution to the fixed version 4.0.6.

Ref: https://access.redhat.com/security/cve/CVE-2026-12143

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11367,16 +11367,16 @@
       }
     },
     "node_modules/form-data": {
-      "version": "4.0.5",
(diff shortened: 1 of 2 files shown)


h-kataria:
Closing in favor of https://github.com/RedHatInsights/tower-analytics-frontend/pull/1252

